PR #458, opened 2015-01-06
Fixed @Optional on the RF generator tile to check for the proper API

Vexatos:


Vexatos:
Also, uhm, shouldn't it be IEnergySender?

PR #459, opened 2015-01-06
Update en_US.lang

MrKunji:
Fixed a really small typo at line 168, where it was 'quatzr' instead of 'quartz'


Vexatos:
But I like my Quatrz :cry:

MrKunji:
It's okay to deny the PR, it's a part of the config file, almost no one looks at it ;)

PR #492, opened 2015-01-21
API update and sanity checks

soundlogic2236:
Added a function to the API for modtweaker and added sanity checks to the lexicon.


Vazkii:
What's with the 2 first commits?

PR #576, opened 2015-02-18
Add DarkOre to OreChid weights

rubensworks:
It seemed easier this way than to include the Botania API to only call one of code.


Vazkii:
Update the package-info too.

rubensworks:
API version bump is now added in first commit.

PR #622, opened 2015-03-04
Make Mana Bursts use the correct source coordinates when fired from a Blaster. Fixes #621

### LINKED ISSUE #621: Bore Warp lens does not drop items when fired from a Mana Blaster
When a mana burst is fired from a mana blaster with a bore warp composite lens, any blocks it breaks appear to disappear into oblivion, while when fired from a redstone mana spreader, any blocks broken drops behind the spreader as an item.

ruifung:
Make the constructor set the burst source as the players coordinates, when its initialized with a EntityPlayer object.

### DIFF
--- a/src/main/java/vazkii/botania/common/entity/EntityManaBurst.java
+++ b/src/main/java/vazkii/botania/common/entity/EntityManaBurst.java
@@ -110,7 +110,7 @@ public class EntityManaBurst extends EntityThrowable implements IManaBurst {
 	public EntityManaBurst(EntityPlayer player) {
 		this(player.worldObj);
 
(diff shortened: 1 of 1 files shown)


Vazkii:
That will break everything else.

ruifung:
Well, you know the codebase better then me. But out of curiosity, how would it break everything else?

Vazkii:
I don't mean literally everything else. But the -1 on Y tells the burst that it wasn't fired from a spreader which is used in a few places. Next off, you can't convert entity positions to block positions by casting to int, you need to run them through MathHelper.floor_double().

ruifung:
Thanks for the clarification.

PR #624, opened 2015-03-05
Fix Bore Warp Lens not dropping blocks if fired from Mana Blaster.

ruifung:
- Add getter to EntityManaBurst to get source player if burst was initialized with a EntityPlayer object.
- Make LensMine add mined items to inventory of EntityPlayer who fired burst, drop on player coords if inventory is full.

I'm casting to EntityManaBurst because at this point, I am reluctant to change anything in the API, although IManaBurst appears to be internal?

### DIFF
--- a/src/main/java/vazkii/botania/common/item/lens/LensMine.java
+++ b/src/main/java/vazkii/botania/common/item/lens/LensMine.java
@@ -15,6 +15,7 @@ import java.util.List;
 
 import net.minecraft.block.Block;
 import net.minecraft.entity.item.EntityItem;
+import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.projectile.EntityThrowable;
 import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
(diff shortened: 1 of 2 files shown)


src/main/java/vazkii/botania/common/item/lens/LensMine.java
```diff
@@ -62,13 +64,24 @@ public boolean collideBurst(IManaBurst burst, EntityThrowable entity, MovingObje
...
-					int dropX = warp ? coords.posX : x;
-					int dropY = warp ? coords.posY : y;
-					int dropZ = warp ? coords.posZ : z;
-
-					for(ItemStack stack_ : items)
-						world.spawnEntityInWorld(new EntityItem(world, dropX + 0.5, dropY + 0.5, dropZ + 0.5, stack_));
-
+					if(warp && coords.posY == -1) {
+						if(burst instanceof EntityManaBurst) {
+							EntityPlayer sourcePlayer = ((EntityManaBurst) burst).getSourcePlayer();
```
ruifung: Ideally this function would be part of the IManaBurst interface, but I'm refraining from modifying anything in the API package.

ruifung:
Well so far I've only had one problem with this, if a burst is fired, and for some reason a player logs off (or does something to invalidate the EntityPlayer object, items disappear - is there a method to check if the EnittyPlayer object is still valid?)

Vazkii:
I'll "fix" this by just making it drop the items in the place. This a bit hacky for my tastes.

PR #631, opened 2015-03-06
Made Helmets of Revealing give vis discounts.

eduardog3000:
All 3 Helmets of Revealing now give a 5% vis discount, the same discount given by the Goggles of Revealing.

I copied IVisDiscountGear from the thaumcraft api, in the same way that IGoggles and IRevealer are copied. IVisDiscountGear imports thaumcraft.api.aspects.Aspect, so I copied that too. I don't know if I have to do anything about Aspect importing org.apache.commons.lang3.text.WordUtils.

Sorry if I did anything wrong, I'm pretty new to Minecraft modding.


Vazkii:
It's intended that they don't.

PR #656, opened 2015-03-15
Flower withering will not be affected by breaking and replacing

pixlepix:
As discussed on IRC, this pull request, when withering is enabled, creates a timer called 'decay ticks'. It will save this timer to NBT when the block is broken, and fetch the value from NBT when a new block is placed.


Vazkii:
Not a big fan of the implementation. I'll redo it later in another way.

PR #660, opened 2015-03-15
Create pt_BR.lang

MrKunji:
Only items and blocks localised.
Only the Pure Daisy got a 'new name', we don't want Termolírios ¬¬
The Lexica Botania entries will take time :|


MrKunji:
Re-doing somethings
/facepalm

Issue #683, opened 2015-03-24
Idea: Use IMC or API  to blacklist items from being attracted the magnet ring


LordIllyohs:
Currently the way to blacklist items from the magnet ring is to open a PR and them to the private list. Using IMC or this API for adding items to the blacklist would remove the need to make a PR for this.

nabijaczleweli:
The simplest way for this would be to modify `ItemMagnetRing.BLACKLIST` (which you can reflect into and change yourself anyway).
The more complex and versatile way would be to have some sort of a `Filter`/`Matcher` list and checking against that (which would require Vazkii to support it).

Relevant lines: [ItemMagnetRing.java#L48](https://github.com/Vazkii/Botania/tree/master/src/main/java/vazkii/botania/common/item/equipment/bauble/ItemMagnetRing.java#L48), [ItemMagnetRing.java#L114](https://github.com/Vazkii/Botania/tree/master/src/main/java/vazkii/botania/common/item/equipment/bauble/ItemMagnetRing.java#L114).

Kubuxu:
@nabijaczleweli To modify this list one would have to include BotaniaAPI. IMC was created to replace plain API in cases like that (ItemStack sending).

PR #684, opened 2015-03-25
Allow black listing of items via IMC Messages.

Lomeli12:
Very basic implementation of the idea mentioned in #683 using IMC.


Vexatos:
Could you allow for disabling items with specific metadata as well, maybe? I'd need to have that to disable a certain sub-item.

Lomeli12:
I'm sure @Vazkii can take care of that since it's a simple fix and the PR has already been pulled.

PR #806, opened 2015-04-23
Fixed Petite Rannuncarpus's range not being affected by mana.

xaeroverse:
We were discussing ranges on IRC and I found this oversight in the code.


Vazkii:
Cheers.

PR #874, opened 2015-05-02
fixed mis-named file, added ores from Simple Ores & akkamaddi's Additions

Sinhika:
I am the current maintainer of akkamaddi's Additions to Simple Ores, and a user of Alexndr's Simple Ores2. This pull request adds almost all of Simple Ores2 & akkamaddi's Additions ores to the orechid and orechid ignem lists, with a few exceptions:

I excluded Simple Cobalt's (from akkamaddi's Additions) "oreCobalt", to avoid conflicting with and unbalancing anything relying on TiC's cobalt; Simple Cobalt's cobalt is an uncommon overworld ore, vs TiC's cobalt which is a rare nether ore. I excluded Simple Ores Netherrocks' "oreIllumenite" because it spawns in glowstone, not netherrack.  Finally, I excluded Simple Cthon's (another akkamaddi Addition) cthon ore, because I have deliberately not ore-dictionaried it to avoid grinder/pulveriser/etc machines from automatically adding ore->dust => smelt to ingot recipes--that would bypass how cthon ore is supposed to be refined in its original mod.  Addresses issue #875.

Finally, your SubTileEntity class file was mis-spelled and would not build. I fixed that.


Vazkii:
Need to update the API version number as well if you're messing with API files.

Sinhika:
The one in package-info.java? okay, done.

PR #894, opened 2015-05-03
Resize category images to 64x64

### LINKED ISSUE #627: Resize categories icons in lexica from 48x48 to 64x64 to support old graphics card
Hi, Vazkii!
I have an issue with botania on my pc:
In game in lexica botania on first page I can't see icons of categories, because that icons resolution unsupported by my graphics card, but when I resized the icons (in mod assets) to 64x64 it became to works perfectly. Can you make it 64x64 by default please? Thanks!

Lazersmoke:
I couldn't find anywhere in the rendering code that relied on the images being 48x48. I also think that this will look the same as before in-game quality-wise. This definitely deserves a double check, though, @Vazkii . hopefully closes #627


Vazkii:
This probably won't work without redoing the render code. And if it would I'd make new images rather than resample old ones.

Lazersmoke:
@Vazkii just tested this, and it works! The rendering code copes fine, so don't worry about changing it when you add new images :)

PR #902, opened 2015-05-05
Added null checks when accessing item frames...

### LINKED ISSUE #897: [Updated OP] Crash when using Corporea Intercepor or Funnel with an Empty Item Frame
[Edited to consolidate issues].

Original Corporea Interceptor Crash:
--
Original Crashlog:
http://pastebin.com/xgngrKRB
Original Setup:

Upon Experimenting, I discovered the crash only occurs when there is an empty item frame on the Corporea Interceptor. Removing the item frame stopped the crash. 

Corporea Funnel Crash:
--
After more testing, the crash also occurs for Corporea Funnels when they are triggered wth an empty Item Frame.

New Crashlogs:
http://pastebin.com/91aKEv7A
Crash occured when I triggered a Corporea Funnel that had an empty Item Frame with a Project Red State Cell. After a relog, the State Cell appears to be triggered until updated.

Upon further testing, a second crash occurs when I attempt to place a button:
http://pastebin.com/1eU2HDyP
Picture is after a relog.

Crash also occurs when removing button:
http://pastebin.com/K2qPYA36
Picture is after a relog.

Removing the empty Item Frame removes all crashes

Lazersmoke:
from coporea things
1 commit for interceptor, 1 commit for funnel

theory is that frame.getDisplayedItem() returns null or empty or something weird when frame is empty, so if(frame.getDisplayedItem()){CODE} replacing CODE should work 
closes #897

### DIFF
--- a/src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaFunnel.java
+++ b/src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaFunnel.java
@@ -70,9 +70,12 @@ public class TileCorporeaFunnel extends TileCorporeaBase {
 			for(EntityItemFrame frame : frames) {
 				int orientation = frame.hangingDirection;
 				if(orientationToDir[orientation] == dir.ordinal()) {
-					ItemStack copy = frame.getDisplayedItem().copy();
-					copy.stackSize = rotationToStackSize[frame.getRotation()];
-					filter.add(copy);
+					ItemStack displayItem = frame.getDisplayedItem()
+					if(displayItem) {
+						ItemStack copy = displayItem.copy();
+						copy.stackSize = rotationToStackSize[frame.getRotation()];
+						filter.add(copy);
+					}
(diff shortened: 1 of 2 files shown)


src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaFunnel.java
```diff
@@ -70,9 +70,12 @@ public void doRequest() {
 			for(EntityItemFrame frame : frames) {
 				int orientation = frame.hangingDirection;
 				if(orientationToDir[orientation] == dir.ordinal()) {
-					ItemStack copy = frame.getDisplayedItem().copy();
-					copy.stackSize = rotationToStackSize[frame.getRotation()];
-					filter.add(copy);
+					ItemStack displayItem = frame.getDisplayedItem()
+					if(displayItem) {
```
Vazkii: This isn't PHP-

PR #913, opened 2015-05-06
Fix Thor ring give SS achievement when using S pick

### LINKED ISSUE #911: Talent Shredder achievement taken with rank S Terra Shatterer
Botania v187

The Talent Shredder achievement is taken with a rank S Terra Shatterer instead of rank SS if the Ring of Thor or Ring of Aesir is equipped. I do not know if this is intended.

Lazersmoke:
Fixes #911


Lazersmoke:
@Vazkii I pinky promise that I tested this one and it works. Like I compilied it in eclipse and put it in my mods folder.

Issue #926, opened 2015-05-08
Drum of Gathering converts mooshrooms to 1 normal cow and 1 fake normal cow


cdfa:
Hi, I was playing around with the drum of gathering because i wanted to make a milk farm, but with mooshrooms (because i can get them easier in my world). When the drum receives a mana burst it converts each mooshroom to a normal cow and another fake cow (it stands still and after relog it disappears).

Not the worst issue, but a fix would be nice :)
Regards cdfa

MrKunji:
Well, Mooshrooms can only be milked with bowls (giving Mushroom Stew), but it seems that the drum is trying to shear them, wich should turn them into normal cows, but there's something weird about the fake ones ¯\_(ツ)_/¯

cdfa:
If you right click a mooshroom with an empty bucket it gives milk, but the shearing makes sense. I accidently discovered a mushroom farm haha.

Regards cdfa

CatDany:
I could reproduce this, it's probably because "onSheared" happens when world.isRemote==true

PR #932, opened 2015-05-09
Mooshroom Cow + onSheared on clientside spawns a fake client-only Cow

CatDany:
#926


CatDany:
I haven't tested this yet, but at least you could try and get the idea

Issue #766, opened 2015-04-18
Terrasteel helm of revealing loses ancient wills on crafting


akvgergo:
If I have a terrasteel helmet with ancient wills applied, crafting a terrasteel helmet of revealing out of it will remove all already applied wills.

Vazkii:
._.

Gencode:
Also loses any enchantments aswell.

Lazersmoke:
@Vazkii pull request submittedt

PR #935, opened 2015-05-09
Finally Helm Revealing crafting closing #766

Lazersmoke:
Please excuse the mangled commit. I had to squash 8-ish of them from when I was still using the web interface. I tested this pretty well, but some of it is a bit hacky. I also wasn't entirely sure what semantics to use (i.e. yes or no newline for 1-line if's) so they may be a bit non-standard. This also includes the y -> posY patch because I couldn't compile otherwise. The only things being copied are the ench tag and the ancient wills. NOTE: the wills use straight NBTHelper calls because I couldn't get addAncientWill() to work :{ #766


Vazkii:
Not pulling because of the remap.

Lazersmoke:
The remap from using a straight up shapeless recipe to a recipe/crafting handler is unavoidable I think (without doing other significantly more insane things) To which remap are you refering?

PR #984, opened 2015-05-16
Added check for water when setting last recipe

### LINKED ISSUE #979: Can add items to dry petal apothecary using last recipe.
If you complete a recipe of the petal apothecary, and then right click it while it is dry, it will make the splash sound and particles will appear. If you then add water you'll see the petals were added. It probably shouldn't allow you to add in the last recipe if it's dry?

Lazersmoke:
closes #979


Vazkii:
Yeah, this doesn't compile. Next time you make a PR without testing it I'm banning you from the repo.

Lazersmoke:
:(

PR #986, opened 2015-05-17
Allow custom scan range for SparkHelper.getSparksAround()

Lazersmoke:
It helps for addons like mine ;) It also keeps backward compatibility with the old, no range function. Untested, but %99 confident.


Vazkii:
This breaks backwards compatibility since the old getEntitiesAround function won't exist.

Lazersmoke:
oh oops, I thought it was private. My bad

PR #994, opened 2015-05-19
Added a config option for configuring the mana cost of Flugel Tiara

lepakko683:
Made the mana cost of Flügel Tiara configurable for balancing purposes.

### DIFF
--- a/src/main/java/vazkii/botania/common/core/handler/ConfigHandler.java
+++ b/src/main/java/vazkii/botania/common/core/handler/ConfigHandler.java
@@ -60,6 +60,7 @@ public final class ConfigHandler {
 	public static boolean relicsEnabled = true;
 	public static boolean stones18Enabled = true;
 	public static boolean ringOfOdinFireResist = true;
+	public static int flightTiaraManaCost = 35;
 
 	public static int flowerQuantity = 2;
 	public static int flowerDensity = 16;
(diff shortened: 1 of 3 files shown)


RX14:
This broke the build https://github.com/Vazkii/Botania/blob/007a365e39518c1ebb9a27db2bfba4f24bd6746f/src/main/java/vazkii/botania/common/item/equipment/bauble/ItemFlightTiara.java#L246 here because `COST` no longer exists.

lepakko683:
Would [this](https://github.com/lepakko683/Botania/commit/064c55fd1357091c9b0d099ad8c3a4bce7732b0c) work? 
I would test it myself if I could.

RX14:
@lepakko683 why can't you test it?

lepakko683:
@RX14 the buildscript doesn't work for me

RX14:
Can you paste the gradle error?

Lazersmoke:
@lepakko683 Try changing https://github.com/Vazkii/Botania/blob/master/build.properties#L4 to `dir_forgecraft=C\:/`. For some reason Vazkii put in a required directory that no one actually has ಠ_ಠ
It won't actually put any thing in your c drive. Make sure to `git rm build.properties` before commiting, though.

lepakko683:
@Lazersmoke @RX14 I get this error: http://pastebin.com/K0MbWtvy

RX14:
@lepakko683 ah I see you are using Linux. Simply touch private.properties, and set the dir_forgecraft to be /dev/null. This works on my arch machine.

Its a shame that Vazkii hates gradle and won't let anyone touch the build script, because it's really lacking.

Issue #998, opened 2015-05-21
Lexica entry change for Exoflame


Shadewing666:
The current entry says "given an efficiency boost"; I'm suggesting changing it to speed boost.

PR #999, opened 2015-05-21
Update en_US.lang

MrKunji:
In favor to https://github.com/Vazkii/Botania/issues/998


brainiac744:
Shouldn't this say "a speed boost" rather than "an speed boost"?

MrKunji:
ಠ_ಠ I will fix soon.

PR #1053, opened 2015-06-05
Update zh_CN.lang

Mrkwtkr:
Removed "汉化:绿冰" (means "Localized by Greenice") because this localization was made by serval people, and there is no sense to add anyone's name in the lang file.


Mrkwtkr:
Make another PR, so close this one.

PR #1055, opened 2015-06-05
Update fr_FR.lang

Mazdallier:


Mazdallier:
Items, entities and blocks translation over.

PR #1079, opened 2015-06-10
Switch particle coloring to static final variable

Lazersmoke:
Hopefully fixes a memory leak; see https://github.com/CatDany/Botania/commit/a629ce25f8946e633757495a65afe68f108b5286#commitcomment-11601022


Vazkii:
There's no memory leak here.

PR #1097, opened 2015-06-16
Possible unbindig  fix and a couple optimizations

Nagapito:
First the optimizations: Doing the same checks over and over is unhealthy. Its preferred to have a uglier but more efficient code like in writeToPacketNBT.

The possible fix, in the linkCollector you could end up with the linkCollector a null and the needsNew at false. All it was needed was for the cached coords to be invalid and not returning a block. Also, when reading theNBT packet, if the coords are 0,-1,0 just set the cache to null. Its an invalid cache coord, no need to force a call to the world that will fail returning a block.

### DIFF
--- a/src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
+++ b/src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
@@ -84,23 +84,18 @@ public class SubTileGenerating extends SubTileEntity {
 		if(linkedCollector == null) {
 			needsNew = true;
 
-			if(cachedCollectorCoordinates != null) {
-				needsNew = false;
-				if(supertile.getWorldObj().blockExists(cachedCollectorCoordinates.posX, cachedCollectorCoordinates.posY, cachedCollectorCoordinates.posZ)) {
-					needsNew = true;
-					TileEntity tileAt = supertile.getWorldObj().getTileEntity(cachedCollectorCoordinates.posX, cachedCollectorCoordinates.posY, cachedCollectorCoordinates.posZ);
-					if(tileAt != null && tileAt instanceof IManaCollector && !tileAt.isInvalid()) {
-						linkedCollector = tileAt;
-						needsNew = false;
-					}
-					cachedCollectorCoordinates = null;
-				}
+			if(cachedCollectorCoordinates != null) {				
+        TileEntity tileAt = supertile.getWorldObj().getTileEntity(cachedCollectorCoordinates.posX, cachedCollectorCoordinates.posY, cachedCollectorCoordinates.posZ);
+        if(tileAt != null && tileAt instanceof IManaCollector && !tileAt.isInvalid()) {
+          linkedCollector = tileAt;
+          needsNew = false;
+        }
+        cachedCollectorCoordinates = null;				
 			}
 		}
-
-		if(!needsNew && linkedCollector != null) {
+    else {
 			TileEntity tileAt = supertile.getWorldObj().getTileEntity(linkedCollector.xCoord, linkedCollector.yCoord, linkedCollector.zCoord);
-			if(!(tileAt instanceof IManaCollector) || tileAt.isInvalid()) {
+			if(tileAt == null || !(tileAt instanceof IManaCollector) || tileAt.isInvalid()) {
 				linkedCollector = null;
 				needsNew = true;
 			} else linkedCollector = tileAt;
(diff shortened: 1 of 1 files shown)


src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
```diff
@@ -84,23 +84,18 @@ public void linkCollector() {
...
-					needsNew = true;
-					TileEntity tileAt = supertile.getWorldObj().getTileEntity(cachedCollectorCoordinates.posX, cachedCollectorCoordinates.posY, cachedCollectorCoordinates.posZ);
-					if(tileAt != null && tileAt instanceof IManaCollector && !tileAt.isInvalid()) {
-						linkedCollector = tileAt;
-						needsNew = false;
-					}
-					cachedCollectorCoordinates = null;
-				}
+			if(cachedCollectorCoordinates != null) {				
+        TileEntity tileAt = supertile.getWorldObj().getTileEntity(cachedCollectorCoordinates.posX, cachedCollectorCoordinates.posY, cachedCollectorCoordinates.posZ);
```
Vazkii: Fix your indentation.

src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
```diff
@@ -184,7 +179,10 @@ public void readFromPacketNBT(NBTTagCompound cmp) {
 		int y = cmp.getInteger(TAG_COLLECTOR_Y);
 		int z = cmp.getInteger(TAG_COLLECTOR_Z);
 
-		cachedCollectorCoordinates = new ChunkCoordinates(x, y, z);
+    
+		cachedCollectorCoordinates = (x!= 0 && y != -1 && z != 0) ? new ChunkCoordinates(x, y, z) : null;
```
Vazkii: x and z can be 0.

src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
```diff
@@ -192,13 +190,16 @@ public void writeToPacketNBT(NBTTagCompound cmp) {
...
 		cmp.setInteger(TAG_TICKS_EXISTED, ticksExisted);
 
-		int x = linkedCollector == null ? 0 : linkedCollector.xCoord;
-		int y = linkedCollector == null ? -1 : linkedCollector.yCoord;
-		int z = linkedCollector == null ? 0 : linkedCollector.zCoord;
-
-		cmp.setInteger(TAG_COLLECTOR_X, x);
-		cmp.setInteger(TAG_COLLECTOR_Y, y);
-		cmp.setInteger(TAG_COLLECTOR_Z, z);
+    if (linkedCollector == null){
```
Vazkii: Keep this consistent, no space after if, space before the bracket.

src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
```diff
@@ -192,13 +190,16 @@ public void writeToPacketNBT(NBTTagCompound cmp) {
...
+      cmp.setInteger(TAG_COLLECTOR_X, 0);
+      cmp.setInteger(TAG_COLLECTOR_Y, -1);
+      cmp.setInteger(TAG_COLLECTOR_Z, 0);
+    }
+    else{
```
Vazkii: else on the same line as the last bracket, space after it.

Vazkii:
You need to duplicate these changes to TilePool and increase the API version in package-info. Have you tested this?

src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
Nagapito: yes, they can but not the combination of the 3 x/y/z.

Its just a simple check to prevent loading to the cachedCollector the coordinates saved in the NBT when the linkedCollector is null.

Nagapito:
Not tested, all of this was analysed and checked by eye. I dont even develop in java, so I used notepad++ :)

Also, I just realized this does not fix the issue. The issue is that readFromPacketNBT is returning a bad set of coordinates or writeToPacketNBT is not saving them correctly. So when the chunk loads, with the conditions with had in linkCollector(), it would set needsNew to false and keep linkedCollector set to null.

My changed will just force the flower to rebind to the nearest collector, not to the correct one, since cachedCollectorCoordinates will be null. 

But I could be totally wrong here. I wish I could properly test it but, since I am a .NET developer, I like to stay far away from Java :)

src/main/java/vazkii/botania/api/subtile/SubTileGenerating.java
Vazkii: Why not just check for y != .1?

Nagapito:
I also just noted, at your line 180, mine 175, is the condition supposed to be negated?
It is not in SubTileEntity...

Vazkii:
Yeah, that's a minor oversight, but it won't break anything. Packet problems is a possibility I could test, but I'm not too sure about that being the problem, because the yCoord will only be -1 if the linkedCollector was null. Either way, I'm closing this, if you want to make a PR, test it first.

PR #1148, opened 2015-07-06
Add mobGriefing check to block breaking section

### LINKED ISSUE #1147: Gaia Guardian on gamerule mobGriefing false
If you have set the gamerule mobGriefing false and spawn the gaia guardian it looks on your client as if it would break blocks and it drops the items but the blocks just disappear (if you relog they're there again) and you can't pick up the dropped items and they don't despawn.They only disappear if you relog.

Lazersmoke:
previously, ghost block destruction was created client side, as in #1147 
closes #1147 
NOTE: check pulled directly from minecraft source code for endermen picking up blocks


Vazkii:
This should happen even with mobGriefing disabled because it's used to fix exploits.

PR #1171, opened 2015-07-15
Fix API and other spelling errors

Lazersmoke:


Vazkii:
So breaking backwards compatibility because of a typo is a good idea? How about no.

Kubuxu:
How about deprecating it now and providing correct alternative.
Next time you will be releasing breaking update  you will remove deprecated methods.

PR #1233, opened 2015-08-03
Spelling, dying → dyeing

trejkaz:
Same as last time minus the two problematic spots.


ruifung:
You should squash the commits.

trejkaz:
I guess doing that is easy enough, but I figured it would cause problems because my existing commits are already public and the git docs say never to do that for stuff that's already visible to others.

ruifung:
If nobody has pulled from your repository, you could just force push - or alternately make a new branch with the squashed commits for PR.

trejkaz:
Aside from spooky automated pullbots, there wouldn't be anyone I hope.

Time to test the pullbots I guess. Well, time tonight anyway, right now I can't get to my computer.

trejkaz:
Stuff caught fire, going to have to re-clone, manually apply, re-submit.

PR #1252, opened 2015-08-14
Added config option to decrease guardian drop chances.

Saereth:
Wasn't able to get in touch before submitting the PR but I did read through the contributing guidelines. I'm hoping everything is in order and we can use this privately even if you decide against.


Saereth:
Yeaah woops, almost pulled an officespace there, good catch!

PR #1253, opened 2015-08-14
Added config option to decrease guardian drop chances.

Saereth:
Wasn't able to get in touch before submitting the PR but I did read through the contributing guidelines. I'm hoping everything is in order and we can use this privately even if you decide against.

Also fixed 1 off error, (thankx Kubuxu) and added a check for min/max values from the original PR


Kubuxu:
Easier would  be to make it `<` instead of `<=` and keep `nextInt(100)` then 0 would mean no drop chance.

Saereth:
makes sense, and that would look better as well, I'll change it this afternoon after work

Titanium237458:
"int lChance = Math.max(100, ConfigHandler.bosslootchance);"
Don't you want a Math.min there? Or no method call at all and just set lChance to bosslootchance? You are checking later if the given value is in the correct interval.

Saereth:
yeah I meant to remove the math.max when I added in the value checks, will remove that as well

src/main/java/vazkii/botania/common/core/handler/ConfigHandler.java
```diff
@@ -162,7 +163,10 @@ public static void load() {
 
 		desc = "Set to anything other than -1 for passive generation flowers (dayblooms, nightshades, hydroangeas) to die after a specific amount of ticks. 24000 is 2 Minecraft days, that's a recomended value.";
 		hardcorePassiveGeneration = loadPropInt("passiveWither.time", desc, hardcorePassiveGeneration);
-
+		
+		desc = "Loot ratio modifier for Gaia Guardians, enter a value from 0-100 100 = standard drop rate 0=no drops";
```
Vazkii: This should be 0-1.

src/main/java/vazkii/botania/common/core/handler/ConfigHandler.java
```diff
@@ -162,7 +163,10 @@ public static void load() {
...
+		bosslootchance = loadPropInt("boss.lootchance",desc, bosslootchance);
```
Vazkii: Fix your commas.

src/main/java/vazkii/botania/common/entity/EntityDoppleganger.java
```diff
@@ -336,32 +336,39 @@ protected boolean canDespawn() {
 	@Override
 	protected void dropFewItems(boolean par1, int par2) {
 		if(par1) {
+			int lChance = ConfigHandler.bosslootchance;
 			boolean hard = isHardMode();
 			entityDropItem(new ItemStack(ModItems.manaResource, hard ? 16 : 8, 5), 1F);
 			boolean droppedRecord = false;
 
+			if (lChance > 100)
```
Vazkii: No space after the if.

src/main/java/vazkii/botania/common/entity/EntityDoppleganger.java
```diff
@@ -336,32 +336,39 @@ protected boolean canDespawn() {
...
+				lChance = 100;
+			if (lChance < -1)
+				lChance = -1;
+				
 			if(hard) {
+				if (rand.nextInt(100) < lChance)
 				entityDropItem(new ItemStack(ModItems.ancientWill, 1, rand.nextInt(6)), 1F);
-				if(ConfigHandler.relicsEnabled)
+				if(ConfigHandler.relicsEnabled && (rand.nextInt(100) < lChance))
```
Vazkii: No.

Saereth:
I'll get all this fixed up today

Issue #1245, opened 2015-08-09
Loki Ring not firing a place block event


AfterWind:
This is for protection, please take this issue into consideration. With a simple event implementation we can prevent all the griefing that can be done with this item.

Vazkii:
Delegating this to @TheWhiteWolves because I'm lazy.

Vazkii:
Oh, and #858

Vazkii:
And #1217. All the same thing.

spannerman79:
Expressing an interest in this, block breaks need to fire the correct Forge BreakEvent standards  - hence a fakeplayer is needed.

If it helps Mekanism already introduced this change in https://github.com/aidancbrady/Mekanism/commit/77db81bcf0c0eaaa7f3757e7b1a610b804491253 for the readout/log file for people wishing to place the fake player with the correct permissions.

Eufranio:
Is there a fix?

PR #1257, opened 2015-08-15
Adds an event when Ring of Loki places a block

Yopu:
Fixes https://github.com/Vazkii/Botania/issues/1245


Vazkii:
This fixes the Ring of Loki, but not the other items that also cause problems. I'll fix this later.

Yopu:
Ah, gotcha. I wasn't aware it was a widespread issue.

PR #1278, opened 2015-08-19
Implement a couple of Thermal Foundation ores in the Orechid

### LINKED ISSUE #1265: Platinum/Shiny Ore
Any chance you can add the Thermal Expansion shiny ore to the Orechid? Based on the default, and from what I can understand looking at the orechid's current list of ores, it should be weighted similarly Emerald ore, or possibly even a bit less. (It was added to normal spawning in a worldgen, but is meant to be rarer than diamonds.)

phantamanta44:
Closes #1265.
Weights are based on generation notes here: http://teamcofh.com/docs/thermal-foundation/world-generation/ores/

### DIFF
--- a/src/main/java/vazkii/botania/api/BotaniaAPI.java
+++ b/src/main/java/vazkii/botania/api/BotaniaAPI.java
@@ -149,8 +149,10 @@ public final class BotaniaAPI {
 		addOreWeight("oreLapis", 1285); // Vanilla
 		addOreWeight("oreLead", 7985); // IC2, Thermal Expansion, Factorization, etc.
 		addOreWeight("oreMCropsEssence", 3085); // Magical Crops
+		addOreWeight("oreMithril", 8); // Thermal Expansionm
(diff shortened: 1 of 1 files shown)


ruifung:
Why not add osmium while at it =P

src/main/java/vazkii/botania/api/BotaniaAPI.java
```diff
@@ -149,8 +149,10 @@
 		addOreWeight("oreLapis", 1285); // Vanilla
 		addOreWeight("oreLead", 7985); // IC2, Thermal Expansion, Factorization, etc.
 		addOreWeight("oreMCropsEssence", 3085); // Magical Crops
+		addOreWeight("oreMithril", 8); // Thermal Expansionm
```
Nincodedo: Expansion, not Expansionm

phantamanta44: Whoops, wasn't paying attention at midnight

Issue #1524, opened 2015-11-07
Worldshaper's Sextant Crash


ColeBeall:
My game crashes when I try to use the Worldshaper's Sextant. I point at a block, right click, and the game crashes after freezing up for a second or two. This happens consistently whenever I attempt to use the item. So far, nothing else causes a crash like this.

FML Log: http://pastebin.com/MGryARw4

Titanium237458:
You don't have CodeChickenCore installed.
Seems like Vazkii imported the wrong Vector class :3

ColeBeall:
Thank you. Didn't realize I needed CodeChicken.

TheWhiteWolves:
You've not supposed to need it, it's happened before it's just a mis-import when using the Vector class.

yrsegal:
shouldn't be that hard a PR

PR #1525, opened 2015-11-08
Change sextant import to Botania vector3

yrsegal:
fix issue #1524


yrsegal:
after reading the contributing.md, i realized that this is for Vazkii to do herself ¯\\\_(ツ)\_/¯

PR #1574, opened 2015-11-22
Make ItemLens combination checking work with any ILens

yrsegal:
This was giving me trouble trying to make my own lens combine in an addon.


yrsegal:
Ignore the first three commit, that's the original reason I made this repo and had to bring it back in to line with master :3

yrsegal:
This has been tested on local, and works (even with addon lenses).

Vazkii:
Rebase this so it's not 4 commits please.

PR #1575, opened 2015-11-22
disable the Fallen Kanade when a redstone signal is applied

MatthiasMann:


Nincodedo:
Does the current recipe for this flower require a redstone root? If not you should include that now.

Vazkii:
Why?

PR #1580, opened 2015-11-23
make ItemLens combination checking more generic

yrsegal:
this will allow ILens instances that aren’t ItemLens to combine with other ILens instances


Vazkii:
Please don't modify an existing API class. Make an ICompositableLens (extends ILens) or something of the sort if you want to extend the behaviour.

PR #1581, opened 2015-11-23
make ItemLens combination checking more generic (ICompositableLens)

yrsegal:
this will allow ILens instances that aren’t ItemLens to combine with other ICompositableLens instances


Vazkii:
Squash the commits >_>

PR #1590, opened 2015-11-27
Fix for botania-share command for non Botania lexicon entries 

alexgeek:
Modified the botania share command logic to use the method getUnlocalizedName instead of the public property unlocalizedName. Now works for both Botania and other mods that use the lexicon.


Vazkii:
Have you tested this?

alexgeek:
I tested to make sure that Botania entries and my entries work.
Two of the edits only affect the share command so will not have a knock on affect, and the alteration to getForcedPage in ItemLexicon is only called in the same class by onItemRightClick so should not affect anything else either.

Vazkii:
This broke the alfheim portal's message.

alexgeek:
Fixed in [this commit](https://github.com/Vazkii/Botania/commit/1b93c9c806e509b7eebfe0d2299d45c88faae95c) I See.
No other problems?

PR #1664, opened 2015-12-16
Update dependencies.

howtonotwin:
Update NEI and CodeChicken dependencies. Also update Multipart and
update Forge because NEI crashes with the old version.

P.S. That Forge was **really** old. Like *woah*.


blacksmithgu:
This unfortunately has merge conflicts right now, but updating depdencencies would be good. I get a lot of random crashes using the current defaults in `build.gradle`.

howtonotwin:
It's actually really strange how `mc_version`, `forge_version`, `build_number` and `dir_repo` seem to be scrambled every time there's a new version, but oh well.

PR #1674, opened 2015-12-20
Fixes a crash when an addon developer adds an empty ArrayList

Eladkay:


JRoffel:
That commit name really cracked my up for some reason... I have to thank you for that :P

Eladkay:
Uh why was this closed?

JRoffel:
Probably because it is faster to just write this himself, than merge it... Or because Vazkii found a problem in the code you wrote...

Eladkay:
Neither of these are possible.

JRoffel:
So writing 4 lines of code can't be faster than a merge? Do you really think that?

Eladkay:
Merging is literally pressing a button.

JRoffel:
And will cost time to pull changes to local which, depending on the internet speed, will take a while (I think Vazkii has already stated that his internet isn't the fastest, though I can't be sure)

Eladkay:
... that's all I have to say

Lazersmoke:
@Eladkay probably the newline at line 68

Eladkay:
In any case the issue isn't fixed. I'll try to push a new commit to fix formatting sometime.

PR #1690, opened 2015-12-25
Allow Hopperhock to split stacks

### LINKED ISSUE #1683: Hopperhocks don't split stacks
When a stack of items is on the floor and the adjacent inventories do not have enough room (say 64 stone on ground but only room for 16 in the chest) for the entire stack, the hopperhock will attempt to pick up the stack and will spam particle effects, but will not actually pick it up and will leave it on the ground.

Steps:
1. Have an almost full chest with the last slot partially filled.
2. Place [floating] hopperhock [petite] next to it.
3. Drop a full stack of the same item in the last slot of the chest near the hh.
4. Observe.

Pics:

howtonotwin:
Change hopperhock logic so that it will split a stack on the ground if the adjacent inventories only have space for some of it. Fixes #1683.

This time, less STOOPID and fewer deletes.


Vazkii:
Have you tested this one properly?

howtonotwin:
Yes, this time it should be fine. Sorry for that fiasco. ;-;

PR #1691, opened 2015-12-26
Update zh_CN.lang

sfmiyuki:


sfmiyuki:
changed the Solegnolia based on 246

PR #1693, opened 2015-12-27
UUID for relics

Katrix-:
Adds UUID support for relics. 

Everything should work fine, both with items that have UUID and for those that doesn't. If an item has both UUID and username information, it will prefer the UUID, if not it will fall back to using the username. Username will be converted to UUID when the owner has the item in his inventory.

I've tested that this works fine with previous builds of Botania.

I was a bit unsure about what to call the static methods in the ItemRelics file though. UUIDS sounds wierd, but I didn't want to break away from what you already had.

### DIFF
--- a/src/main/java/vazkii/botania/api/item/IRelic.java
+++ b/src/main/java/vazkii/botania/api/item/IRelic.java
@@ -10,6 +10,8 @@
  */
 package vazkii.botania.api.item;
 
+import java.util.UUID;
+
 import net.minecraft.item.ItemStack;
 import net.minecraft.stats.Achievement;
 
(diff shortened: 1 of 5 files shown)


williewillus:
If this ends up not being taken I might steal it for the 1.8 port because my uuid conversion is pretty temporary (no migration mechanisms) and this one seems more thought out :p

Katrix-:
Sounds good. 

Is it even reasonable to migrate between 1.7 and 1.8 though? Normally migrating between versions like that is strongly discouraged.

williewillus:
If possible, I'd like at least people who play with just botania alone or with a couple other mods to be able to keep their botania stuff into 1.8

Vazkii:
Will pull for 1.8.

williewillus:
I just merged this into my branch - this can be closed for now, I guess

PR #1733, opened 2016-01-10
Baubles can now IBlockProvide when worn

yrsegal:
This allows for things like the Botanical Addons Toolbelt to provide from
the tools in them while worn.


yrsegal:
There is a better solution. I'll work on it soon.

PR #1734, opened 2016-01-10
Make baubles able to IBlockProvide

yrsegal:
This allows for things like Botanical Addons’ Toolbelt to provide from
the tools in them while worn.


yrsegal:
oops, another bug >.<

PR #1735, opened 2016-01-10
Make baubles able to IBlockProvide

yrsegal:
This allows for things like the Botanical Addons toolbelt to provide from
the tools in them while worn.

HAS BEEN ACTUALLY TESTED THIS TIME :V

### DIFF
--- a/src/main/java/vazkii/botania/common/item/rod/ItemExchangeRod.java
+++ b/src/main/java/vazkii/botania/common/item/rod/ItemExchangeRod.java
@@ -31,6 +31,7 @@ import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.util.ForgeDirection;
 import net.minecraftforge.event.entity.player.PlayerInteractEvent;
 import net.minecraftforge.event.entity.player.PlayerInteractEvent.Action;
+import vazkii.botania.api.BotaniaAPI;
 import vazkii.botania.api.item.IBlockProvider;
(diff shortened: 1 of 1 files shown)


TheWhiteWolves:
Question: Wouldn't it be more useful from a player perspective to to try pulling from the bauble inventory first rather than your personal inventory, i.e. say your pulling cobble, it would be more useful to leave a supply of cobble in your main inventory and take from the bauble inventory before depleting your on hand supply.

It's a simple thing but personally I would think it might be more beneficial this way.

yrsegal:
True, true. That's a simple fix.

yrsegal:
there you are.

TheWhiteWolves:
Not to be picky but just want to point out you switched the count ones, so now the variable names are opposite to what they are counting :stuck_out_tongue_closed_eyes:

yrsegal:
oops. also, easy to fix >.<

yrsegal:
there :3

TheWhiteWolves:
nice, also won't that also make it so the count doesn't include your inventory unless you have some in a bauble inventory.

Would this bit mean it will stop the count if your bauble has none, even when your inventory does? or would it return 0 for none found, I can't remember what the count returns.

```Java 
if (baubleCount == -1) return -1;
```

yrsegal:
-1 means infinity in this case.

TheWhiteWolves:
Yep, ignore me, forgot that's a method in that file, so once I expanded it and had a look it basically means your in creative mode

PR #1738, opened 2016-01-11
Fix #1728 - Grass seeds will not spread to dirt/grass blocks under other blocks

### LINKED ISSUE #1728: Pasture seeds will spread to dirt blocks underneath other blocks
When using a seed, the initial spread will travel to dirt blocks covered by other blocks.

blacksmithgu:
Updates BlockSpawner in common/item/ItemGrassSeeds to not spread to dirt and grass blocks which have a non-air-like block above them. Added a good amount of comments and fixed all of the warnings and strange variable names (eg, "par1ItemStack" -> "itemStack") as well.

If neccessary, the non-functionality-critical parts can be removed.

Are there other block spawners where this change is neccessary?

### DIFF
--- a/src/main/java/vazkii/botania/common/item/ItemGrassSeeds.java
+++ b/src/main/java/vazkii/botania/common/item/ItemGrassSeeds.java
@@ -12,6 +12,7 @@ package vazkii.botania.common.item;
 
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Random;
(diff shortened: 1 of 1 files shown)


yrsegal:
you messed with V's item params. she won't accept this PR request, probably. Just edit the code you're working on.

blacksmithgu:
Thought as much. This should get rid of the bulk of the extraneous changes.

TheWhiteWolves:
I'll let Vaz decide if this should be pulled but first have you tested that it all functions without breaking anything else and works as intended?

blacksmithgu:
As in the other PR, there isn't a big unit testing base to work on, but here are some screenshots:

Example Original Setup, with some blocks sitting on top of the dirt:


Addition of Pasture Seeds allows grass to grow under snow (as it has a light opacity of 1), but prevents it from growing under cobblestone, slabs, or stairs. Grass grows normally in the opposite direction.





The other variants also work (golden scorched, etc).

PR #1739, opened 2016-01-12
Fix #1724 - Slabs and snow don't prevent Rod of Shifting Crust replacement

### LINKED ISSUE #1724: Possible issue or quirk of the Rod of Shifting Crust with snow layers
When trying to replace blocks that have snow around and on top of them, the rod doesn't function. That is, with the 1 snow layers. Its likely not an issue at all, just a quirk of the behavior. A few pictures for clarification: http://imgur.com/a/peNmx

Botania version 1.8-243

blacksmithgu:
Corrects the issue where the Rod of Shifting Crust would not replace blocks surrounded by snow or other specific non-opaque cubes (eg, slabs). Blocks are checked for visibility by checking if there are any adjacent blocks which do not have an adjacent solid side.

### DIFF
--- a/src/main/java/vazkii/botania/common/item/rod/ItemExchangeRod.java
+++ b/src/main/java/vazkii/botania/common/item/rod/ItemExchangeRod.java
@@ -154,56 +154,57 @@ public class ItemExchangeRod extends ItemMod implements IManaUsingItem, IWirefra
 		}
 	}
 
-	public Boolean[][][] createMatrix(World world, ItemStack stack, Block blockToSwap, int metaToSwap, int xc, int yc, int zc, Block targetBlock, int targetMeta) {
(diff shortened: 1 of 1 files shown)


TheWhiteWolves:
As with the other PR have you regression tested it

blacksmithgu:
There isn't much of a unit testing base here, of course, but it functions equivalently when there are no snow-like or slab-like blocks around. When there are, it takes whether or not the block is actually visible into account. I'll give you a few screenshots.

blacksmithgu:
Testing setup for sanity, again. The only code I changed was the logic for finding which blocks to swap, so we'll focus on that.

Using cobblestone as the material we're swapping to. Used on a standard homogenous surface (with default range and so forth):



Replacement works, and blocks which are already the target block aren't selected again:



Also only selects visible blocks:



And a small recreation of the actual screenshot in #1724, with snow surrounding dirt:



As you can see, all of the dirt is properly selected instead of only the dirt not surrounded by snow. Replacement also works, as do mana costs and using other staffs as block providers.

PR #1752, opened 2016-01-15
Fix #1749 - Mana Tablets no longer lose durability in storage drawer

### LINKED ISSUE #1749: [Storage Drawers] Mana Tablet Damage bar vanishing
Full mana tablets lose their damage bars when placed into storage drawers.  I placed a bug report with them and ended up back here to report the same issue.

Here is a link to the issue report: https://github.com/jaquadro/StorageDrawers/issues/193

"So, the drawer indeed writes out a 0 damage value on NBT save. But the prototype stack stored within the drawer itself has the correct damage value of 1.

Vakii has an inconsistency in the mana tablet implementation. The item overrides the getDamage() accessor, and scales the internal mana value (range 0 - 500,000) to a damage value in the range (0 - 1000). If the mana is completely full, it calculates an inconsistent damage value of 0. This needs to be brought up there."

blacksmithgu:
Makes Mana Tablet durability bar actually use `showDurabilityBar` and `getDurabilityForDisplay` instead of the deprecated `getDisplayDamage`; as `getDisplayDamage` is no longer used, damage differences can be removed entirely (so no more item stack damage 1 to 1000).

In order to ensure compatibility with previous releases, `getDamage` is overridden to ensure that all mana tablets have a damage value of 0.

Some examples of it working with storage drawers:

Before placing in drawer:

In drawer:

After:

(From left to right: empty, semi-full, full, and creative)

### DIFF
--- a/src/main/java/vazkii/botania/common/item/ItemManaTablet.java
+++ b/src/main/java/vazkii/botania/common/item/ItemManaTablet.java
@@ -49,13 +49,16 @@ public class ItemManaTablet extends ItemMod implements IManaItem, ICreativeManaP
 
 	@Override
 	public void getSubItems(Item par1, CreativeTabs par2CreativeTabs, List par3List) {
-		par3List.add(new ItemStack(par1, 1, 10000));
+		// Empty tablet
+		par3List.add(new ItemStack(par1, 1));
 
-		ItemStack fullPower = new ItemStack(par1, 1, 1);
+		// Full tablet
+		ItemStack fullPower = new ItemStack(par1, 1);
 		setMana(fullPower, MAX_MANA);
 		par3List.add(fullPower);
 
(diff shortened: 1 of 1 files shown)


TheWhiteWolves:
Do those exist in the current minimum forge version (which I think is currently set to 10.13.2.1291), if not the minimum will need bringing up to when they were added.

blacksmithgu:
I'm currently building against `10.13.2.1291`, so both `showDurabilityBar` and `getDurabilityForDisplay` are both available in the current minimum.

Any other compatibility concerns? The code automatically corrects the damage to 0 for all items so they stack in barrels and so forth properly now, so world upgrading should work correctly.

TheWhiteWolves:
Looks good to me, didn't even know those things existed before

PR #1754, opened 2016-01-16
Bugfix followup - Block swappers use sidedness to control concurrency

blacksmithgu:
Followup on PR #1750, which thanks to @williewillus 's input, changes the means of concurrency to explicit sidedness instead of just concurrent data structures (as lovely as they may be).

This actually makes for a notable difference in single player, as the block swapper is no longer duplicated on both the local client and server (saving a respectable amount of duplicate computation).

As usual, proof of working for both the Terra Axe and Pasture Seed - tested both on single player and multiplayer:

Note that the solution simply employs remote checks, which works fine in principle. If I revisit this class for a clean up, however, a proper server and client proxy (where the client only handles visual effects and the server does most of the actual work) would definitely first in my priority list. Comments welcome.

### DIFF
--- a/src/main/java/vazkii/botania/common/item/ItemGrassSeeds.java
+++ b/src/main/java/vazkii/botania/common/item/ItemGrassSeeds.java
@@ -11,13 +11,13 @@
 package vazkii.botania.common.item;
 
 import java.util.ArrayList;
-import java.util.Collections;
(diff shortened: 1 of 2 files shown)


williewillus:
Proxies would be unnecessary, remote checks would be the standard way to do this kind of stuff.

PR #1774, opened 2016-01-24
Added support for StorageDrawers and Deep Storage API

the-vindex:
Added wrapper around IInventory for integration with different storage systems. Used this for Storage Drawers and JABBA (general Deep Storage API).

I had to include Deep Storage API in Botania, which is quite ugly :-( But it will add mass support for all inventories using this API, which is apparently plenty.

TODO:
I don't have correct formatter for Botania - formatting may be off.

### DIFF
--- a/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
+++ b/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
@@ -12,10 +12,14 @@ package vazkii.botania.api.corporea;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.WeakHashMap;
 import java.util.regex.Pattern;
+
+import com.jaquadro.minecraft.storagedrawers.api.storage.IDrawer;
+import com.jaquadro.minecraft.storagedrawers.api.storage.IDrawerGroup;
 
 import net.minecraft.inventory.IInventory;
 import net.minecraft.inventory.ISidedInventory;
@@ -25,6 +29,7 @@ import net.minecraft.util.AxisAlignedBB;
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.util.ForgeDirection;
+import vazkii.botania.common.Botania;
 
 public final class CorporeaHelper {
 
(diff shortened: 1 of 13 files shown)


src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
```diff
@@ -167,69 +165,71 @@ public static int getCountInNetwork(ItemStack stack, Map<IInventory, Integer> in
...
+	 * purposes of counting huge amounts.
 	 */
-	public static List<ItemStack> requestItem(Object matcher, int itemCount, ICorporeaSpark spark, boolean checkNBT, boolean doit) {
-		List<ItemStack> stacks = new ArrayList();
+	public static List<ItemStack> requestItem(Object matcher, int itemCount, ICorporeaSpark spark, boolean checkNBT,
+			boolean doit) {
+		List<ItemStack> stacks = new ArrayList<ItemStack>();
 		CorporeaRequestEvent event = new CorporeaRequestEvent(matcher, itemCount, spark, checkNBT, doit);
-		if(MinecraftForge.EVENT_BUS.post(event))
+		if (MinecraftForge.EVENT_BUS.post(event))
```
TheWhiteWolves: remove all the spaces between your if and for statements and the brackets

src/main/java/vazkii/botania/api/package-info.java
```diff
@@ -1,4 +1,4 @@
-@API(owner = "Botania", apiVersion = "76", provides = "BotaniaAPI")
+@API(owner = "Botania", apiVersion = "77", provides = "BotaniaAPI")
```
TheWhiteWolves: Don't need to bump the API version as a release has not been made since it was increased to 76

TheWhiteWolves:
As a general comment, do you need all of those files in the StorageDrawers API?, only include the ones that are required, we don't need to ship all the extra stuff

src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
the-vindex: I'll wait for proper Eclipse formatter in the repo. After that I'll reformat all my code with it so it matches what is correct. If you have it, please commit it to /eclipse folder and/or ping me on IRC.

the-vindex:
We don't have to include all StorageDrawers API, I could try to cherrypick files from that API. On the other hand API is not included in JAR file, so it doesn't pollute runtime environment. So should I try to cherrypick?

TheWhiteWolves:
you should, it may not be shipped with the jar but its included in the repo and pulled down to us, the less there is the better, means less stuff likely to need updating and less for us to manage.

the-vindex:
I tried and it went very well. So done.

Vazkii:
Proxy any API requests. The Botania API should not require others.

src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
TheWhiteWolves: you should be able to use the cleanup profile that is in the eclipse folder

the-vindex: I have that imported, but clean up profile won't format the code (like those whitespaces between if and (). This is defined in Code Style = formatter.

the-vindex:
Could you give me some example of what you mean by proxying?

Vazkii:
https://github.com/Vazkii/Botania/blob/master/src/main/java/vazkii/botania/api/internal/IInternalMethodHandler.java#L118

the-vindex:
Ok, done, new PR opened. Hope this is what you intended.

PR #1775, opened 2016-01-24
Added support for StorageDrawers and Deep Storage API

the-vindex:
Added wrapper around IInventory for integration with different storage systems. Used this for Storage Drawers and JABBA (general Deep Storage API).

I had to include Deep Storage API in Botania, which is quite ugly :-( But it will add mass support for all inventories using this API, which is apparently plenty.

TODO:
I don't have correct formatter for Botania - formatting may be off.

### DIFF
--- a/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
+++ b/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
@@ -25,6 +25,7 @@ import net.minecraft.util.AxisAlignedBB;
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.util.ForgeDirection;
+import vazkii.botania.api.BotaniaAPI;
 
 public final class CorporeaHelper {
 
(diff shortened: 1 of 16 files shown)


yrsegal:
"WrappedInvenotryBase"

the-vindex:
As far as I've deduced, Botania uses BlahBlahBase instead of AbstractBlaBlah.
If you'd like to see better naming for individual interfaces/classes, I'd welcome it.

yrsegal:
"Invenotry"

the-vindex:
ah, lol, fixing it now

src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
```diff
@@ -25,6 +29,11 @@
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.common.util.ForgeDirection;
+import vazkii.botania.api.BotaniaAPI;
+import vazkii.botania.common.Botania;
+import vazkii.botania.common.integration.corporea.WrappedDeepStorage;
+import vazkii.botania.common.integration.corporea.WrappedIInventory;
+import vazkii.botania.common.integration.corporea.WrappedStorageDrawers;
```
TheWhiteWolves: Need to get @Vazkii to confirm this but I think there will be a problem here with calling into the main Botania code from the API, afaik this will cause problems for other people who are using the API

yrsegal:
yeah, the api may only call java code, minecraft code, and api code. Nothing from main botania.

the-vindex:
Actually those were old imports which should have been removed. Now it uses IInternalHandler.

src/main/java/vazkii/botania/api/internal/IInternalMethodHandler.java
```diff
@@ -125,4 +126,9 @@
 
 	public void sendBaubleUpdatePacket(EntityPlayer player, int slot);
 
+	/**
+	 * Wrap inventories in the network into wrappers providing compatibility for storage mods.
+	 */
+	List<IWrappedInventory> wrapInventory(List<IInventory> inventories);
```
the-vindex: This is were API calls Botania internals

TheWhiteWolves:
I can't see anything else that's a problem code wise, as long as you have fully tested it, I'll pass over to @Vazkii for if she's ok with it being merged.

the-vindex:
I did test it:
https://docs.google.com/spreadsheets/d/10ezf483BcXQIFnXL0wkF15Ku7JRjDlPMYbu7SAXaGIg/edit?usp=sharing

the-vindex:
I've improved Deep Storage API detection in order to exclude it from JAR. Otherwise this is final from me. 
As I can't seem to get official code style formatter, I've followed instruction to remove spaces between if and ( from previous PR.

Please pull :)

williewillus:
Might want to add the copyleft header to everything (you can find on top of basically any class, and I left the date/time off of it when I did it in 1.8)

the-vindex:
Done

TheWhiteWolves:
@Vazkii have you had a chance to take a look and make a decision on this one yet?

PR #1782, opened 2016-02-01
Fixed Marimorphosises having incorrect downward range

PikadudeNo1:
oh noes I broke the "don't use the GitHub editor" rule, truly I am horrible and this PR is completely unacceptable


TheWhiteWolves:
I have tested this for you this time, but in future do not use the github editor as this leaves the code untested and you have no clue if your change has caused a problem elsewhere.

PR #1800, opened 2016-02-10
Multiparts now work on both the client and the server.

yrsegal:
Has been tested.


Vazkii:
This crashes if you don't have FMP installed.

PR #1801, opened 2016-02-10
Revert "Multiparts now work on both the client and the server."

Vazkii:
Reverts Vazkii/Botania#1800


howtonotwin:
Can you please delete these extra branches? They are rather annoying to look at.

PR #1802, opened 2016-02-10
Multiparts now work on both the client and the server.

yrsegal:
Ok, not removing the reflection should allow this to work without FMP.


TheWhiteWolves:
I guess you've checked that it doesnt crash this time?

yrsegal:
It doesn't, but my machine tends to run code which crashes when vazkii
tests it.

TheWhiteWolves:
Hopefully nothing goes wrong this time lol

PR #1809, opened 2016-02-16
Added ja_JP.lang

elbo2525:
Update ja_JP.lang


TheWhiteWolves:
Squash your commits down to one please

williewillus:
Also since this is so big it'd be nice if we could have someone else knowledgeable look over it as well

williewillus:
Oh I just looked at the lang file..

If you have not translated something, do not include it in the ja-JP lang file, it will be inherited from en_US automatically. If you copy it, then it will desync - if en_US gets updated then the copy in ja-JP will be old and potentially translated incorrectly.

Only include the lines that are actually in Japanese, leave anything out that you haven't translated.

williewillus:
I don't think the other lang files follow this rule exactly, but no better time to start :P

elbo2525:
OK, I got it. I'll correct the file.

PR #1811, opened 2016-02-21
Add ja_JP.lang

elbo2525:
Add ja_JP.lang


TheWhiteWolves:
I'll leave this here for a few days, so if anyone knowledgeable can review it please do, otherwise I'l take it as correct for now

PR #1814, opened 2016-02-22
Update zh_CN.lang to latest.

3TUSK:
I have tested these translation in game so that they *are* suitable.

Also the language file itself has been cleaned up, so that it keeps the same line number as that in `en_US.lang`.


TheWhiteWolves:
please squash your commit down to 1 please

Issue #1825, opened 2016-03-11
Botania shows up in stack trace of crash with Ichun's PiP mod


Robijnvogel:
See issue https://github.com/iChun/IssueTracker-Mods/issues/31
Take a look at my [crash report](http://pastebin.com/Mc56MX0C) please.

Can Botania do anything to solve this? Any idea what is going on? IChun's Github doesn't look very active.

PR #1826, opened 2016-03-11
Update MultiblockRenderHandler.java

Robijnvogel:
Fixed some code that was definitely buggy. Possibly solves #1825

Why it was buggy? Checking for mc.thePlayer != null and then before analyzing the result calling !mc.thePlayer.isSneaking() in the same expression is not a good idea. But you have more experience with Java than I do, so you know... you probably just missed this.

### DIFF
--- a/src/main/java/vazkii/botania/client/core/handler/MultiblockRenderHandler.java
+++ b/src/main/java/vazkii/botania/client/core/handler/MultiblockRenderHandler.java
@@ -62,11 +62,15 @@ public final class MultiblockRenderHandler {
 	@SubscribeEvent
 	public void onWorldRenderLast(RenderWorldLastEvent event) {
 		Minecraft mc = Minecraft.getMinecraft();
-		if(mc.thePlayer != null && mc.objectMouseOver != null) {
-			if( (!mc.thePlayer.isSneaking() || anchor != null)) {
-				mc.thePlayer.getCurrentEquippedItem();
-				renderPlayerLook(mc.thePlayer, mc.objectMouseOver);
-			}
+		boolean playerIsNotSneaking = false; //whether the player is not null and not sneaking
+		
+		if(mc.thePlayer != null) {
+			playerIsNotSneaking = !mc.thePlayer.isSneaking();
+		}
+		
+		if(mc.objectMouseOver != null && playerIsNotSneaking || anchor != null) {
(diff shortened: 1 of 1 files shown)


yrsegal:
No, her code is fine. Yours is a mess, though.

Robijnvogel:
@yrsegal , so please tell me what would happen if you run her code and mc.thePlayer returns null?

For as far as I know, Java always evaluates every single part of a boolean expression. (Except if an exception or error is thrown of course.)

So evaluating (a && b && c) would evaluate if b holds **and** evaluate if c holds, even if (a == false). (This is possibly something that Java could improve in their compiler though. They probably have their reasons not to though.)

In this case, a is defined as (mc.thePlayer != null) and c is defined as (!mc.thePlayer.isSneaking())
The way it WAS formulated, Java would evaluate the (!mc.thePlayer.isSneaking()) statement even if (mc.thePlayer == null), see my previous statement.

I agree that my code looks like a messy workaround, but I tried to not duplicate code and keeping that in mind, this sounded like the simplest solution to:
1. Make sure that the isSneaking() method is only called if thePlayer is not null.
2. Keep the logic expression intact.

If you can solve the bug, because there definitely is one, and at the mean time keep the total logic expression intact **and** the code clean. Then by all means, show me how.

NOTE:
You should look at the second commit. The first commit was wrong and I know.

yrsegal:
a && b will only evaluate B should A be true. a || b will only evaluate B
should A be false.

Also, two messy commits, with a bad comment? Rebase it into one for ANY
chance of merge.

Robijnvogel:
I just tested it with a simple program and it seems you are correct. Well that's going to change the way I write code quite a bit.
I must have been confusing Java with some older version or another language (Python? Lua maybe?) entirely.

So that means that it is iChun's "FakePlayer," or whatever class he uses that extends EntityPlayer, that crashes on the isSneaking() method, hmm?

If I had been right about that one little thing, then my whole commit would have been justified, but it turns out that you were right and I was terribly underestimating Java's capacities. I'm sorry for wasting your time.

I'll close this pull request now.

yrsegal:
Python works the same way as Java for this (which is great for code golf),
so probably lua.

mindforger:
O_o this reminds me strongly of that Baubles VS Ender IO Capacitor incident(while wearing Botania Stuff)! Where the player instance was lost for some unknown reason!

PR #1910, opened 2016-10-05
Added ManaProficiencyEvent and reworked IManaProficiencyArmor

Flaxbeard:
Allows mods to interface with rod proficiency from outside just armor items. Also allows ManaProficiencyArmor more discretion on what rods to make proficient.


Flaxbeard:
Just realized I didn't package my changes to the rods, they still call the deprecated methods. Let me fix.

PR #1911, opened 2016-10-05
Reworked getFullDiscountForTools and ManaDiscountEvent

Flaxbeard:
This pull allows armor sets that grant a mana discount (or mods that add a discount through ManaDiscountEvent) to change the discount based on the item in use. This would enable armors that only grant a discount to certain rods, or certain tools (or give an increased discount for certain rods or tools).

For example, one could implement a "fire" armor set that would grant a 50% discount to the fire rod but 10% to other rods.


Flaxbeard:
Ignore the commits from the other PR that were reversed. I derped with git.

Issue #1961, opened 2016-10-20
Update Bauble Case Recipe -> Ore Dictionary chestWood


VT-14:
Please update the Bauble Case recipe to use the ore dictionary (chestWood) rather than minecraft:chest.  At the moment, I am unable to use a Spruce Chest from Quark, which I find mildly humorous.

I took a quick look (maybe 10 seconds) through JEI [Oak Chest -> uses].  The only other recipe that I saw using a chest in Botania was the Red Stringed Container, and that already uses the ore dictionary.

Botania r1.9-333

PR #1962, opened 2016-10-20
Use the ore dictionary for chests in the Bauble Box recipe.

InsomniaKitten:
https://github.com/Vazkii/Botania/issues/1961 - I have nothing better to do today so I'm just helping people out.


VT-14:
I took another look through JEI and found that the Corporea Retainer also has this issue.  Line 1960.

Ctrl + F through that entire file, and that looks like the last use of blocks.CHEST using vanilla crafting.

PR #2074, opened 2016-12-16
Modified the hopperhock to check for ticksExisted instead of age.

### LINKED ISSUE #2018: Hopperhock doesn't pickup witherstar
The hopperhock doesn't pick up the wither star, after its dropped from the wither
it does pick it up after i drop it

so it seemed a bit weird, so i report it 😕 

Forge 2121
Botania r1.9-337
java 1.8
Linux server, Windows client

MightyCupcakes:
They both represent the amount of ticks the EntityItem existed except that the EntityItem class has some magic numbers for age to represent certain item states.

Fixes #2018


williewillus:
this won't work because ticksExisted is not saved. thus this will mess up the timing of everything if you reload the world

PR #2118, opened 2017-01-09
Change hopperhock to check for negative age values

### LINKED ISSUE #2018: Hopperhock doesn't pickup witherstar

MightyCupcakes:
This is mainly for items dropped by the game that are set to no despawn (age = -6000). Proposed solution is to add the item's lifespan to the age if the age is found to be negative.

Reading through the EntityItem code, the only time age is set to negative is when setNoDespawn() function is called or age is set to -32768 which in this case age will never be incremented and the default lifespan value is not sufficient to bring the age to positive values.

Tested and it works. Fixes #2018.


williewillus:
i don't think this will interact well with for example the open crate, which sets the age to -200 if it's given a redstone signal, precisely so the hopperhock won't pick it up early

williewillus:
I say we should just hard code the wither star for now

MightyCupcakes:
@williewillus 

I implemented another check to check if the negative age value is large enough before adding it.

This assumes that item.lifespan is 6000, so any age value larger than -5900 will fail the check and the value will not be added. Therefore this will only work if the EntityItem is specifically set to no despawn.

codewarrior0:
May be related to #2676

PR #2162, opened 2017-02-13
Removed @Nonnull descriptors as needed to not crash when viewing the …

Penrif:
…Botania creative tab and using the Lexicon

### DIFF
--- a/src/main/java/vazkii/botania/client/model/FloatingFlowerModel.java
+++ b/src/main/java/vazkii/botania/client/model/FloatingFlowerModel.java
@@ -218,7 +218,7 @@ public class FloatingFlowerModel implements IBakedModel {
 	private final ItemOverrideList itemHandler = new ItemOverrideList(ImmutableList.of()) {
 		@Nonnull
 		@Override
-		public IBakedModel handleItemState(@Nonnull IBakedModel model, ItemStack stack, @Nonnull World world, @Nonnull EntityLivingBase entity) {
+		public IBakedModel handleItemState(@Nonnull IBakedModel model, ItemStack stack, World world, @Nonnull EntityLivingBase entity) {
 			// Items always have GRASS island
 			IFloatingFlower.IslandType islandType = IFloatingFlower.IslandType.GRASS;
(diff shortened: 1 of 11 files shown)


williewillus:
what is this for? 

nullability annotations are not retained at runtime (unless your ide injects runtime checks, but that would cause crashes in forge itself before botania even runs)

Penrif:
Just trying to get a functional dev environment.  Pulled down source, got it building, then would crash with stacks like the following:

java.lang.IllegalArgumentException: Argument for @Nonnull parameter 'bellows' of vazkii/botania/client/render/tile/RenderTileBellows.renderTileEntityAt must not be null
	at vazkii.botania.client.render.tile.RenderTileBellows.$$$reportNull$$$0(RenderTileBellows.java)
	at vazkii.botania.client.render.tile.RenderTileBellows.renderTileEntityAt(RenderTileBellows.java)
	at vazkii.botania.client.render.tile.RenderTileBellows.renderTileEntityAt(RenderTileBellows.java:24)
	at net.minecraftforge.client.ForgeHooksClient.renderTileItem(ForgeHooksClient.java:603)
...

So this is me fixing all the ones I hit by scanning through the Botania creative tab and using the Lexicon.  Spot checked a few deeper into the stack and in all the cases I found the null was an appropriate param for the situation.

williewillus:
Disable runtime nonnull assertions in IntelliJ

williewillus:
I'll pull it to fix the crashes in dev but generally those assertions are just annoying in modded :P

Penrif:
Can do.  Curious tho, why mark up with 'em if they aren't enforced?

Bluexin:
I find them useful as I write my mods in Kotlin (which is null-safe), so these annotation checks are useful as bridge ^-^

williewillus:
Half of these aren't problems in 1.11 anymore, and I fixed the rest + other stuff in 1.11 as well, so closing this since it's not *super* critical

Penrif:
Allright, I'll rebase my other branch that was off this...

PR #2175, opened 2017-02-21
Entity drop item null checks

### LINKED ISSUE #2126: The Drum of the Gathering causes server to crash (Direwolf20 1.10.2)
I'm currently using the Direwolf20 1.10.2 modpack version 1.2.2, and the Drum of the Gathering crashes the server.

Are you in the latest version of the mod?
Yes

Are you in the latest version of Forge?
Yes

Are you using vanilla Forge (not MCPC+)?
Sponge, so no

Are you using Optifine?
Server, N/A

The crash log seems to indicate that on upon collision with the Drum of the Gathering, a mana burst had a java.lang.NullPointerException.

This was likely an automated system, and it doesn't happen every time. I am unsure of what exactly causes the crash, since it was an automated system. This is obviously probably not very easy to reproduce.

Crash Report: http://pastebin.com/10d37y5j

Penrif:
Should fix #2126

### DIFF
--- a/src/main/java/vazkii/botania/common/block/mana/BlockForestDrum.java
+++ b/src/main/java/vazkii/botania/common/block/mana/BlockForestDrum.java
@@ -149,9 +149,11 @@ public class BlockForestDrum extends BlockMod implements IManaTrigger, ILexicona
 						if(itemstack != null && itemstack.getItem() == Items.BUCKET && !world.isRemote) {
 							while(itemstack.stackSize > 0) {
 								EntityItem ent = entity.entityDropItem(new ItemStack(Items.MILK_BUCKET), 1.0F);
-								ent.motionY += world.rand.nextFloat() * 0.05F;
-								ent.motionX += (world.rand.nextFloat() - world.rand.nextFloat()) * 0.1F;
-								ent.motionZ += (world.rand.nextFloat() - world.rand.nextFloat()) * 0.1F;
+								if(ent != null) { // Shouldn't be given that we specified the item to drop, but let's be safe anyway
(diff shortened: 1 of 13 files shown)


williewillus:
you included some unrelated commits?

src/main/java/vazkii/botania/common/block/mana/BlockForestDrum.java
```diff
@@ -149,9 +149,11 @@ else if(variant == DrumVariant.CANOPY)
 						if(itemstack != null && itemstack.getItem() == Items.BUCKET && !world.isRemote) {
 							while(itemstack.stackSize > 0) {
 								EntityItem ent = entity.entityDropItem(new ItemStack(Items.MILK_BUCKET), 1.0F);
-								ent.motionY += world.rand.nextFloat() * 0.05F;
-								ent.motionX += (world.rand.nextFloat() - world.rand.nextFloat()) * 0.1F;
-								ent.motionZ += (world.rand.nextFloat() - world.rand.nextFloat()) * 0.1F;
+								if(ent != null) { // Shouldn't be given that we specified the item to drop, but let's be safe anyway
```
williewillus: remove the comment since it's clear we check bc entityDropItem is nullable

looks good besides this + extra commits

Penrif: The reason I called this one out and not the others is because this is, in practice, not needed since it only returns null based on the ItemStack being malformed.  The other cases do not control that parameter as this one does.

All the same, if you prefer it gone the PR is set up for contributors to edit vOv

williewillus:
could you remove the unrelated commits from the branch?

Issue #2185, opened 2017-03-06
Orechid crashes server when non-block ores are in the ore dictionary


smbarbour:
The latest version of Ex Nihilo Adscensio registers its chunks in the ore dictionary as ore<Name> (e.g. oreGold), but the chunks are not blocks, causing the following exception:

java.lang.NullPointerException: Ticking block entity
        at vazkii.botania.common.block.subtile.functional.SubTileOrechid.onUpdate(SubTileOrechid.java:60)
        at vazkii.botania.common.block.tile.TileSpecialFlower.func_73660_a(TileSpecialFlower.java:100)
        at net.minecraft.world.World.func_72939_s(World.java:1804)
        at net.minecraft.world.WorldServer.func_72939_s(WorldServer.java:620)
        at net.minecraft.server.MinecraftServer.func_71190_q(MinecraftServer.java:709)
        at net.minecraft.server.dedicated.DedicatedServer.func_71190_q(DedicatedServer.java:387)
        at net.minecraft.server.MinecraftServer.func_71217_p(MinecraftServer.java:613)
        at net.minecraft.server.MinecraftServer.run(MinecraftServer.java:471)
        at java.lang.Thread.run(Thread.java:745)

I am working on a fix to check if the ore returned by OreDictionary.getOres() is a block.

PR #2186, opened 2017-03-06
Restrict ores from the ore dictionary to only those that are blocks.

smbarbour:
Simple check to ensure that the ore item returned by the ore dictionary is a subclass of ItemBlock


smbarbour:
Fixes #2185

PR #2212, opened 2017-03-23
Fixes #2207

### LINKED ISSUE #2207: Overload potion nerfed by mc 1.9 potion changes
in minecraft 1.9 15w33a the strength and weakness effects were changed
copied from gamepedia
Potion of Strength changed from 130% increase to 3-point increase.
Potion of Weakness changed to 4-point decrease, was 0.5-point decrease.

an overload potion gives strength 4 and weakness 3
at current this comes to a total of +0 damage
I would suggest reducing the weakness to level 1 and possibly bumping the strength up to 5

codycoolwaffle:
MC 1.9 changed the potion effect numbers, this is to accommodate with those changes.


williewillus:
this'll do for now

PR #2273, opened 2017-06-04
Fixes #2267. Storage Drawers causing crashes and exceptions.

### LINKED ISSUE #2267: Requesting items from StorageDrawers controller using Corporea network causes exceptions or crash depending on setup
Log: https://gist.github.com/Zarathul/b48fbe9d4bcc14e7d1d1c8b5d0738e86

Left button produces the crash, right button causes same exception but no crash. With more complex setups a world can become unloadable because it crashes everytime (happened with mine).

Briefly looking over the log and code, it seems like the controller only supports capabilities (no IInventory) while the drawers themselves support both. Botanias StorageDrawers wrapper expects IInventory so shit goes boom.

https://github.com/jaquadro/StorageDrawers/blob/025c0e2501e842f3ec3de06b54013fe1301b265e/src/com/jaquadro/minecraft/storagedrawers/block/tile/TileEntityController.java

https://github.com/Vazkii/Botania/blob/69e48d116e1ed8f77c19c37277aff775d4d63586/src/main/java/vazkii/botania/common/integration/corporea/WrappedStorageDrawers.java#L41

palaster:
The Storage Drawers mod has changed from using IInventory to fully using Capabilities, which works by default with the current corporeal sparks. So, I removed the integration, wrapper, and the API because it was causing issues with newer version of Storage Drawers.

Fixes #2267.


williewillus:
this is tested and it works fully with storage drawers?

palaster:
Yes, I have tested it both in and out of my dev. environment. And, yes both drawers and controllers appear to be working correctly.

williewillus:
excellent, thanks

PR #2340, opened 2017-08-29
Fixes the shaking mouse-over elements in the Lexicon.

gerhard2202:
This should fix the following weird shaking when using the Lexicon:

- The tooltip that indicates that you can hold Shift to get more information about an entry should no longer flash in and out of view.
- The tooltip that is displayed when you hold Shift should no longer pop into and out of existence, and it should stay at a fixed width instead of shaking.
- The shadow placed behind entries in the Lexicon when you mouse over them should no longer shake on the right edge of the Lexicon.

### DIFF
--- a/src/main/java/vazkii/botania/client/gui/lexicon/GuiLexicon.java
+++ b/src/main/java/vazkii/botania/client/gui/lexicon/GuiLexicon.java
@@ -259,7 +259,7 @@ public class GuiLexicon extends GuiScreen {
 	
 	public void drawScreenAfterScale(int par1, int par2, float par3) {
 		float time = ClientTickHandler.ticksInGame + par3;
-		timeDelta = time - lastTime;
+		timeDelta = time - lastTime + partialTicks;
 		lastTime = time;
 		partialTicks = par3;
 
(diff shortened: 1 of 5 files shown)


src/main/java/vazkii/botania/client/gui/lexicon/GuiLexicon.java
```diff
@@ -259,7 +259,7 @@ public final void drawScreen(int x, int y, float partialTicks) {
 	
 	public void drawScreenAfterScale(int par1, int par2, float par3) {
 		float time = ClientTickHandler.ticksInGame + par3;
-		timeDelta = time - lastTime;
+		timeDelta = time - lastTime + partialTicks;
```
williewillus: should this use the new partialTicks (`par3`) instead of the old one?

gerhard2202: Not sure, but when I tested it and used the result, there was no visible shaking. Will test both tomorrow.

gerhard2202: They both work fine. Should I update the pull request to use par3 for consistency?

williewillus: yeah, also rename those parameters to have actual names while you're at it :P

gerhard2202:
Parameters should be more descriptive now and the fix now uses the new partial ticks.

williewillus:
Thanks!

PR #2416, opened 2017-10-16
Add Beetroot to Full Crop Farm challenge

codewarrior0:
It's needed for Cellular Blocks, after all.


codycoolwaffle:
#EndBeetrootDiversity

PR #2446, opened 2017-11-22
Fix incorrect step assist height

InsomniaKitten:
1.25F is player jump height, which is what should ideally be used for step assist. This allows a player to step up to places where they would otherwise be able to jump to, such as a block with 2 layers of snow on top of it.

If there are other places in the code where this needs changing, feel free to close this PR and I'll open an issue for it to be dealt with later.


williewillus:
sure why not

PR #2505, opened 2018-01-09
Fix #2497 by removing unneeded special-casing for TileEntityChest

### LINKED ISSUE #2497: Hopperhocks not sorting correctly with quark's varient chests
Forge version: 14.23.1.2562
Botania version: r1.10-350
Quark version: r1.4-120

It was suggested on reddit i post here after noticing hopperhocks werent filtering correctly in the 1.12.2 version of Garden of glass by melanX. 
I have modified the pack, however the below can be reproduced on a fresh install, on the latest forge and botania.

Steps to reproduce:
1. Load into a new gog world.
2. Place 3 quark chests, 2 of one wood type 1 of another side by side. (think normal, trapped, normal chest pattern)
3. Add item frames to the face and add items. (i used each chest and the item frame as all 3 are already in inventory)
4. Add either floating hopperhocks behind the chests. 
5. Toss the items associated to the item frames onto the middle chest or hopperhocks. (the range indicated by the monocle covers both areas)

What I expected to happen:
Items sorted into correct chest, indicated by the item frames.

What happened instead:
If hopperhock is used, all items are placed into the chests at random. 2 items in 1 chest, 1 in another. 
If hopperhock petite is used, all items are placed into middle chest.

quat1024:
~~Admission: I didn't actually check to see if this fixed the specific problem with Quark chests, because that means downloading a new mod and putting it in my dev workspace. Which is EFFORT.~~

~~I can't imagine why it wouldn't work though *(famous last words)*~~

i have tested! scroll down

### DIFF
--- a/src/main/java/vazkii/botania/common/core/helper/InventoryHelper.java
+++ b/src/main/java/vazkii/botania/common/core/helper/InventoryHelper.java
@@ -40,12 +40,6 @@ public class InventoryHelper {
 		if(te == null)
 			return null;
 
-		if(te instanceof TileEntityChest) {
-			IItemHandler doubleChest = VanillaDoubleChestItemHandler.get((TileEntityChest) te);
-			if(doubleChest != VanillaDoubleChestItemHandler.NO_ADJACENT_CHESTS_INSTANCE)
-				return doubleChest;
-		}
-
 		IItemHandler ret = te.hasCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, side) ?
 				te.getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, side) : null;
 


williewillus:
I remember needing to have that there for some reason or another, so it'd be nice if you checked :P 

both with quark and vanilla chests, in a variety of orientations and with the hopperhock targeting both the upper or lower sides of the chest

williewillus:
Have you tested this?

quat1024:
Oh, not yet. I'll do it now.

quat1024:
hopperhocks never inserted into the outside chests:
image

hopperhocks never inserted into the middle chest:
image
(also that ^ with the hopperhocks on the top side)

here, the corporea system cannot see the cobblestone inside the middle chest:
image

the reverse is also true (it cannot see the items in the outer chest): 
image

tried really hard, but everything seems to work a-ok with quark chests
image

and finally, tried some "regular" double chest insertion; all of these got filled
image

it's looking good to me, any other configurations i missed?

williewillus:
Does a single corporea spark on a double chest still detect items in both sides of the chest?

quat1024:
[Yep](https://user-images.githubusercontent.com/2068021/35362476-7aec5ebc-0133-11e8-8c99-014eec84e6fa.png)

Issue #2485, opened 2017-12-23
Gaia Guardian music does not loop


Zsashas:
Forge version: 2565
Botania version: 352

Steps to reproduce:
1. Fight the Gaia Guardian for long enough to hear the full song

What I expected to happen: For the music to restart

What happened instead: Ambience's  assigned boss music started playing

Also, I assume it';s intentional that the Unstable Reservoir cannot target it?

PR #2600, opened 2018-04-10
Close Issue #2485

rco9er:
I have added a new variable, timestampMusicStart, to keep track of the time the music for the Gaia Guardian boss has been playing. When the song has reached its end, the boolean isPlayingMusic is set to false, and thus the music is replayed.


williewillus:
this isn't entirely correct, since sounds pause while the game is paused, but the system clock will keep ticking.

williewillus:
can you address the above?

williewillus:
c177cda

Issue #2612, opened 2018-04-24
Botania crash with Thaumcraft


DevTimur:
# Version Information
Forge version:  14.23.3.2673
Botania version:  Botania+r1.10-353
Compatibility issue mod: Thaumcraft-1.12.2-6.1.BETA10

# Further Information
Link to crash log: https://gist.github.com/DevTimur/a265eb28ce8ed6d70722c19ca7c67339
Steps to reproduce:
1. Add Thaumcraft mod
2. Lauch game
3. Crash
What I expected to happen:
I expected modpack to work :)
What happened instead:
Thaumcraft killed Botania, Azanor claim it's Botania's fault

NillerMedDild:
Duplicate of #2608

DevTimur:
Yeah, but it's 1.12.2 version

CurePindal:
Its still a duplicate, because its from an outdated API reference.

kasper-tfg:
Even with Botania 354 it s still crashing.  
https://paste.ee/p/CpLR0

CurePindal:
So the commit that should be a hotfix isnt a fix? Derp.

Issue #2608, opened 2018-04-20
Crash when TC6 is loaded


tyra314:
A Minecraft instance with only Botania and Thaumcraft beta9 is crashing with a class not found error. Without TC6, botania loads just fine.

# Version Information
Forge version: 14.23.3.2655
Botania version: r1.10-353
Thaumcraft: 6.1.BETA9

# Further Information
Link to crash log: https://gist.github.com/tyra314/6033a50d3be1fe17e5d3e198f695ed0e
Steps to reproduce:
- Happens during startup.

See also: Azanor/thaumcraft-beta/issues/468

CurePindal:
Also happens with TC6 Beta10. According to Azanor its probably an outdated TC integration. 
Can we haz a hotfix? Was so looking forward for Botania and TC on 1.12.2 =(

CurePindal:
Oh, just noticed this commit as hotfix?
https://github.com/Vazkii/Botania/commit/ed72ff3d5fcf92b1760a164c30797d6ba68da374
When does this get live?

Dalethium:
I just built and tested that commit, this error comes up: 
https://pastebin.com/aaCPMbvc

Dalethium:
Thanks for the speedy fix! I'll test it later today!

PR #2613, opened 2018-04-24
Fix crash with Thaumcraft 6 1.12.2

kasper-tfg:
This is a simple fix for issues #2612 and #2608


williewillus:
I didnt see this until now :thonk:, sorry! Fixed in e6e7568

PopDinosaur:
Dont know if you still look at this, but it seems as if i have ran into this problem as well, could someone talk me through how to stop thaum crashing botania

PR #2622, opened 2018-05-03
Fixed usages of deprecated getFullDiscountForTools method

TheIllusiveC4:
This commit changes the two calls to the deprecated getFullDiscountForTools method in requestManaForTool and requestManaForToolExact to the newer ItemStack-sensitive one. Otherwise, the stack is always null.


williewillus:
looks good, sorry for the delay

PR #2632, opened 2018-05-13
Add JEI handlers for Orechid and Orechid Ignem

codewarrior0:
Adds two JEI pages that show the expected number of ores when a stack of stone is processed by the Orechid. Entries are sorted by weight. Shows a single ore when the expected chance is less than 1/64. Output ores rotate through the oredict to allow the player to look up any variety of e.g. Tin Ore to see the Orechid entry. Skips any ores that don't exist as ItemBlocks, and skips any oredict keys that don't contain any ItemBlocks at all.

image

### DIFF
--- a/src/main/java/vazkii/botania/client/integration/jei/JEIBotaniaPlugin.java
+++ b/src/main/java/vazkii/botania/client/integration/jei/JEIBotaniaPlugin.java
@@ -15,7 +15,9 @@ import mezz.jei.api.JEIPlugin;
 import mezz.jei.api.recipe.IRecipeCategoryRegistration;
 import mezz.jei.api.recipe.VanillaRecipeCategoryUid;
 import net.minecraft.item.Item;
+import net.minecraft.item.ItemBlock;
 import net.minecraft.item.ItemStack;
+import net.minecraftforge.oredict.OreDictionary;
 import vazkii.botania.api.BotaniaAPI;
 import vazkii.botania.api.recipe.RecipeBrew;
 import vazkii.botania.api.recipe.RecipeElvenTrade;
@@ -33,6 +35,10 @@ import vazkii.botania.client.integration.jei.elventrade.ElvenTradeRecipeCategory
 import vazkii.botania.client.integration.jei.elventrade.ElvenTradeRecipeWrapper;
 import vazkii.botania.client.integration.jei.manapool.ManaPoolRecipeCategory;
 import vazkii.botania.client.integration.jei.manapool.ManaPoolRecipeWrapper;
+import vazkii.botania.client.integration.jei.orechid.OrechidIgnemRecipeCategory;
+import vazkii.botania.client.integration.jei.orechid.OrechidIgnemRecipeWrapper;
+import vazkii.botania.client.integration.jei.orechid.OrechidRecipeCategory;
+import vazkii.botania.client.integration.jei.orechid.OrechidRecipeWrapper;
 import vazkii.botania.client.integration.jei.petalapothecary.PetalApothecaryRecipeCategory;
 import vazkii.botania.client.integration.jei.petalapothecary.PetalApothecaryRecipeWrapper;
 import vazkii.botania.client.integration.jei.puredaisy.PureDaisyRecipeCategory;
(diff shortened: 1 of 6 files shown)


src/main/java/vazkii/botania/client/integration/jei/JEIBotaniaPlugin.java
```diff
@@ -63,8 +75,16 @@ public void registerCategories(IRecipeCategoryRegistration registry) {
...
 				new ElvenTradeRecipeCategory(registry.getJeiHelpers().getGuiHelper()),
-				new ManaPoolRecipeCategory(registry.getJeiHelpers().getGuiHelper())
-				);
+				new ManaPoolRecipeCategory(registry.getJeiHelpers().getGuiHelper()),
+				new OrechidRecipeCategory(registry.getJeiHelpers().getGuiHelper()),
+				new OrechidIgnemRecipeCategory(registry.getJeiHelpers().getGuiHelper())
+		);
+	}
+
+	public static boolean doesOreExist(Map.Entry<String, Integer> entry) {
```
williewillus: signature is a bit weird for the name, why not just pass the key only?

src/main/java/vazkii/botania/client/integration/jei/orechid/OrechidRecipeWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+public class OrechidRecipeWrapper implements IRecipeWrapper, Comparable<OrechidRecipeWrapper> {
+    public String oreDictKey;
+    public int weight;
+
+    static ItemStack inputStack;
```
williewillus: why is this static?

src/main/java/vazkii/botania/client/integration/jei/orechid/OrechidRecipeWrapper.java
```diff
@@ -0,0 +1,82 @@
...
+    List<List<ItemStack>> outputStacks;
+
+    public ItemStack getInputStack() {
+        if (inputStack == null) inputStack = new ItemStack(Blocks.STONE, 64);
+        return inputStack;
+    }
+
+    public List<List<ItemStack>> getOutputStacks() {
+        if (outputStacks == null) {
```
williewillus: why the weird lazy-init stuff (same for the input and in the other class)? I'd rather just compute it all up front and have it be final

src/main/java/vazkii/botania/client/integration/jei/orechid/OrechidIgnemRecipeCategory.java
```diff
@@ -0,0 +1,95 @@
...
+
+        itemStacks.set(0, ingredients.getInputs(ItemStack.class).get(0));
+
+        itemStacks.init(1, true, 70, 12);
+        itemStacks.set(1, ItemBlockSpecialFlower.ofType(LibBlockNames.SUBTILE_ORECHID_IGNEM));
+
+        itemStacks.init(2, true, 99, 12);
+
+        itemStacks.set(2, ingredients.getOutputs(ItemStack.class).get(0));
+
```
williewillus: make your whitespace consistent between init and set (here and the other categories too)

src/main/java/vazkii/botania/client/integration/jei/JEIBotaniaPlugin.java
codewarrior0: Mostly because I'm using it to filter `oreWeights.entrySet()` and chose to do `filter(JEIBotaniaPlugin::doesOreExist)`. Could just as easily do `filter(e -> doesOreExist(e.getKey()))`

src/main/java/vazkii/botania/client/integration/jei/orechid/OrechidRecipeWrapper.java
codewarrior0: Unless I'm mistaken, the orechid only ever uses Stone as an input block, so I shared the input stack of 64 stone between all recipes.

src/main/java/vazkii/botania/client/integration/jei/orechid/OrechidRecipeWrapper.java
codewarrior0: I wasn't sure if everything below that would be available at static init time. I've had problems before with doing things in static initializers that should have been done later.

codewarrior0: But I should have just put it all in the normal constructor. Right.

src/main/java/vazkii/botania/client/integration/jei/orechid/OrechidIgnemRecipeCategory.java
codewarrior0: Got it.

codewarrior0:
Done.

Issue #2646, opened 2018-06-07
Elementium equipment is repairable with Manasteel ingots, and not Elementium ingots


Reassembly:
# Version Information
Forge version:  1.12.2-14.23.4.2703
Botania version:  r1.10-354

# Further Information
Link to crash log: n/a
Steps to reproduce:
1. Remove Mana tablets/mirrors from inventory, to prevent Mana-fueled auto-repair.
2. Use a piece of Elementium equipment (axe/shovel/etc.) a few times, allowing it to become slightly damaged.
3. Attempt to repair the damaged Elementium tool at an anvil with an Elementium ingot. Observe that this is rejected by the anvil.
4. Attempt to repair the damaged Elementium tool at an anvil with a Manasteel ingot. Observe that this is permitted by the anvil.

**What I expected to happen:** Damaged Elementium tools, should they not be auto-repaired with Mana, should require (costly) Elementium ingots to repair.

**What happened instead:** Damaged Elementium tools can be, and can only be, repaired with much cheaper Manasteel ingots. (Or Mana.)

Typically, this would be an unnoticeable non-issue, as it is reasonable to assume that by the time one has the resources and infrastructure to make Elementium equipment, one would also have a ready supply of Mana for repair, and would never _need_ to use actual ingots for that purpose. However, this can cause inconsistent and/or exploitable behavior with other mods which use the registered repair material of a given tool for _other_ purposes.

For example, I am using the "QualityTools" mod, which assigns tools/armor/baubles a "quality" (a set of random bonuses/maluses), which can be re-rolled (for better or worse) by "reforging" a tool with an ingot of its repair material - which keeps the reforging resource cost parallel with the tool's base resource cost, regardless of whether the tool is vanilla or mod-added. For game-balance purposes, if I want to reforge, say, an Elementium axe to try to get the best possible quality, the process _should_ be an expensive gamble of Elementium ingots, and not merely burn through (relatively) cheap Manasteel ingots.

Alternatively, if this is _intended_ behavior (that Elementium equipment is repairable with Manasteel), then this should be mentioned in the Lexica Botania. (At first, I thought that Elementium equipment was not repairable with ingots _at all,_ until I did some further experimenting.)

TheWhiteWolves:
[...]

PR #2656, opened 2018-06-21
Elementium Tools now repair with Elementium

ALongStringOfNumbers:
This fixes https://github.com/Vazkii/Botania/issues/2646 
I assumed that Elementium was the correct material to repair the tools instead of manasteel because the tools were made from Elementium, also Elementium is used to repair the armor instead of manasteel.


williewillus:
Can you give the parameters proper names?

ALongStringOfNumbers:
I can do so, but I was just using the parameter names found in the Elementium armor repair as well as the manasteel repair

williewillus:
looks good

PR #2667, opened 2018-07-03
Update Thaumcraft integration

### LINKED ISSUE #2629: Suggestion: Thaumcraft Aspects
i just wanted to say that it would be cool if you could give the items and blocks and so on thaumcraft aspects .

Hubry:
Resolves #2629. I hope Willie is still interested in adding aspects!

This PR changes the following:
* Thaumcraft 6 is now pulled directly from Curseforge. The alternative was dumping the whole TC6 API into the api folder in the repo, because adding aspects requires referencing the ThaumcraftApi class, which references a whole lof of stuff, so good luck with compiling. (Removed the marker interfaces from the api folder while I were at it.)
* Made upgrade recipes move attached Thaumcraft item data properly. While infusion enchantments (and, by extension, warping) are not yet available on armor, that's probably coming at some point, as you can see some starts of armor enchantments in Thaumcraft's internal classes.
* Fixed full mana inkwells pulled from creative menu not working. They were just missing the mana tag and Thaumcraft thought they were empty for that reason.
* Blacklisted flux rifts instead of (removed) aura nodes from Rod of the Shaded Mesa.
* Cleaned up a few TODOs around Thaumcraft integration.
* Finally, assigned Thaumcraft aspects to Botania items!

### Aspects

This was probably the most time consuming task and I am still not fully satisfied with it. I tried to assign aspects to the base items and things that don't get them automatically generated, but in the end I at least added aspects to the default assignment for a lot of stuff.

I did reference Wire's aspects from Botanical Addons (see the issue linked on top), however since Thaumcraft had some pretty big changes done to the aspect list and other things since TC4 (a lot of aspects no longer exist and items contain a lot more essentia nowadays), I had to do my own thing for a big majority of items. I also use autoassigned aspects a lot more.

I waited with this PR for Beta 17, as that one fixes items containing NBT ignoring aspects assigned to items with no NBT. Turns out a lot of items, including every single Botania bauble, get at least an empty tag the moment you press Shift while looking at it.

I only noticed one real issue - white petal powder doesn't have Herba I assigned to it and white petal blocks don't have any aspects. Haven't noticed anything else weird other than that.

I'm open to feedback about the assigned aspects.
Oh, and before someone asks, the twigs have only 8 Herba because I wasn't a fan of how every item made with them had like 30 to 60 Herba. That aspect is way too common anyways.


williewillus:
looks good, we can improve the aspects in a separate pr

Hubry:
And as it got merged I noticed spaces sneaked in in some spots for indentation. Oops. If I make a followup I'll fix those... 😓

PR #2668, opened 2018-07-05
Reintroduce corporea index requesting from inventory/JEI item panel.

### LINKED ISSUE #2556: JEI click-to-request got lost in the misty depths of 1.7
Here's the issue you said on discord someone should post :P

1.7 had a feature where you could press a special "Corporea Request" hotkey while hovering in NEI and issue corporea requests. It would use the item count specified in the little number entry box in NEI, and I think you could also hold combinations of Shift and Ctrl to change the amount of items per request. Here's an old vasgoo video of the feature https://www.youtube.com/watch?v=uWFwAIhcOe0

This doesn't exist for JEI :(

One consideration here is that JEI doesn't have the weird number entry box that NEI has. Could possibly be replicated with a little popup dialog or something.

Hubry:
Resolves #2556.

This was removed in 08550cc442e7b7f67146d02752037ba12b78c4b8, during the move from NEI to JEI. Rewritten to use JEI instead of NEI. Well, I say rewritten, but basically only the part handling the request was still usable.

The keybind only shows up if JEI is present and is in a separate category.

Should function exactly like it did in 1.7 (more or less, I didn't actually test the old one). Sorry @quat1024 but the amount bar was ditched before that build's release (and the replacement method is still usable) :P 

I am considering actually decoupling this from JEI because most of the logic is actually JEI-independent (unlike the old version, which heavily depended on NEI) and the part that handles requesting from the JEI panel could be separated out. Still submitting for review though.

One problem I noticed is that when the JEI panel is closed when hovering over an item, it gets stuck and requests it whenever the keybind is triggered in an inventory. This should not be happening on latest JEI as checks to see if the panel is hidden were added.


williewillus:
update the other lang files' lexicon page8 from NEI->JEI while you're at it

Hubry:
I checked all the other ones when making the PR - only eu_us and zh_cn have the page 8, and I already covered those.

Hubry:
So, I did the thing I wanted and decoupled the feature from JEI. It is now using a supplier that gets set to an actual one pulling from the JEI panel if JEI is present. I did test this in an instance without JEI and it works as expected.

I did edit the en_us lexicon page to note the change (though not the zh_cn one, I'm leaving that to someone who knows the language).

williewillus:
Are you good with your changes now?

Hubry:
Yeah, should be all good now.

PR #2678, opened 2018-07-20
Add "Thinking Hand" cosmetic bauble

quat1024:
image

When you wear this bauble, you appear to be deep in thought.

:thinking: :thinking: :thinking:

image

Recipe is 8 tiny potatoes around a mana string. (All 16 petal and 16 floral-powder recipes were already used up; I thought it would be weird to use a 16-colorable item but add just 1 more bauble without adding baubles for the other colors in the set; and it's funny.)

***ALSO:*** threw in a tiny bugfix for the "Googly Eyes" bauble rendering inside the player's head. It's just a tiny bug I noticed while testing the baubles, didn't think it was worth a 2nd PR, and it's kinda related

image


quat1024:
:thinking:

PR #2689, opened 2018-08-03
Corporea retainer signal strength now depends on the size of the request

quat1024:
The comparator signal strength of the corporea retainer is now larger for larger requests. The math is exactly the same as the corporea crystal cube - it starts at one, and each step up needs double the item count.

This will probably break a few redstone contraptions, but it's easy to fix by just putting a repeater in front of the comparator.

Originally this was a patch in my addon Incorporeal but it doesn't really belong there.


williewillus:
if the intention's to keep the math the same between the two, extract it to a common method and have both call it

quat1024:
This is true, I blame still being in "patch as little as possible mode"

Will close in favor of a new pr

PR #2691, opened 2018-08-03
Add some events and make things a bit more extensible

quat1024:
This is a little PR that helps me get rid of a *lot* of hackery in Incorporeal.

- adds `InventoryWrapEvent`, fired when the corporea system wraps an inventory, and allows for replacing the `IWrappedInventory` instance
- adds `CorporeaIndexRequestEvent`, fired just before a player requests an item from a corporea index; cancellable, for a feature I have that allows for denying specific players access to a corporea index
- changes some `private` things in `WrappedIInventory` to `protected`, just so I can extend it more easily (beforehand I would just extend `WrappedInventoryBase` and duplicate a lot of the logic)

I do feel kinda weird about adding events and stuff that, realistically, isn't going to get much use outside of my mod. Especially making addons to the corporea system which just piles on more technical debt. But hey.


williewillus:
is it possible to instead add an interface (either directly used by TE's or exposed as a cap) that TE's can implement/provide to allow corporea access to them?

botania doesn't usually do this kind of thing with events, using new interfaces is the convention here.

quat1024:
This is true - I *kinda* want to move away from interfaces all over the place since something something Forge 1.13 removing `@Optional`, but this isn't really the place to be doing it...

One trick with `wrapInventory` is that `InvWithLocation` is just an item handler with a position, no tile entity. I'd rather have people apply their `ISpecialWrappedInventory` interface to their tile entity (or even to their block), but that would involve a lot of changes, since by the time `wrapInventory` is called the specific tile entity that the inventory belongs to is long gone, and just doing `getTileEntity` *again* is kinda silly.

I've reverted those inventory wrap changes for now, maybe it needs more excavating than I thought.

PR #2693, opened 2018-08-04
Update TC integration to beta 21, small tweaks to aspects

Hubry:
Beta 20 includes a rather big change, an event for aspect registration, and deprecates the old ways to register aspects. This PR updates TC integration to use the event. I also tweaked some aspects and added them to lenses, and took the opportunity to fix formatting things from my last 2 PRs.

As TC6 beta 20 requires Forge 14.23.4.2705, I bumped the Forge version to that one. (I've been tempted to bump JEI as well to use Thaumic JEI during testing, to be honest).

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,8 +82,8 @@ jar {
 
 dependencies {
     deobfCompile "com.azanor.baubles:Baubles:1.12-1.5.2"
-    deobfCompile "mezz.jei:jei_1.12:4.7.0.67"
-    deobfCompile "thaumcraft:Thaumcraft:1.12.2:6.1.BETA17"
+    deobfCompile "mezz.jei:jei_1.12.2:4.10.0.198"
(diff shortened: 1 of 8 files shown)


williewillus:
you can go ahead with the JEI/thaumic jei bumps, just keep the jei dep on a stable ver and update the dependency strings as appropriate to prevent crashes with older jei versions

Hubry:
Um, maybe don't pull this yet until Azanor/thaumcraft-api#38 gets fixed. Ugh.

williewillus:
forget about the second part, i'm dumb

Hubry:
It is ready, yes. I just redid the commit history there, the contents are the same as 5 days ago.

I don't really have many further tweaks in mind (maybe aspects on brews, but that's gonna be either quite complicated or a lot of hardcoding) and it is compatible with beta 21 and above. 

Might require one followup if the entity registration gets moved into the event as well, in which case I'll probably try to do the brews.

PR #2702, opened 2018-08-16
Corporea retainers use richer signal strength depending on request size (but better)

quat1024:
This PR:
* adds a cute helper function in `CorporeaHelper` to get a signal strength 0-15 from a corporea request size 0-16384, using the usual "each step up needs double the size" method,
* modifies the corporea crystal cube to use that helper function to determine its comparator strength
* modifies the corporea retainer to use that helper function to determine its comparator strength, as opposed to the previous "15 if there is any pending request at all, ever" behavior

### DIFF
--- a/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
+++ b/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
@@ -17,6 +17,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.math.AxisAlignedBB;
 import net.minecraft.util.math.BlockPos;
+import net.minecraft.util.math.MathHelper;
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
 import vazkii.botania.api.BotaniaAPI;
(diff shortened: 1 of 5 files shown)


src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
```diff
@@ -297,4 +297,12 @@ public static boolean shouldAutoComplete() {
 	public static String stripControlCodes(String str) {
 		return patternControlCode.matcher(str).replaceAll("");
 	}
+	
+	/** 
+	 * Returns the comparator strength for a corporea request that corporea crystal cubes and retainers use, following the usual "each step up requires double the items" formula.
+	 */
+	private static final double LOG_2 = Math.log(2);
+	public static int signalStrengthForRequestSize(int requestSize) {
+		return requestSize == 0 ? 0 : Math.min(15, (int) Math.floor(Math.log(requestSize) / LOG_2) + 1);
```
williewillus: I think Math or mojangs MathHelper might have utils for log2

quat1024: oh snap didn't know!

PR #2722, opened 2018-08-27
Resize and reorient Tiny Potatoes held as equipment by other Tiny Potatoes

codewarrior0:
Also adjusts the placement of any blocks on the top side of the potato
such that they rest flatly on the potato's head.

image


Vazkii:
I can't believe you've done this.

PR #2723, opened 2018-08-28
Allows avatars to use the wand of the forest

folding-code:
Upon use, the wand will go to bind mode and turn off.
It resets when the avatar is deactivated, so you still need to trigger it with redstone.
Also, this is basically just a nicer looking dispenser, so it doesn't use mana.
Uses the dirt rod overlay because it isn't too bright.

I haven't tested it since minecraft modding environments don't compile on my computer.


Vazkii:
>I haven't tested it since minecraft modding environments don't compile on my computer.

Definitely not pulling something that's untested.

PR #2739, opened 2018-09-14
Prevent Fallen Kanade from healing during GG fights

codewarrior0:


williewillus:
I might be dumb but how does this prevent that?

codewarrior0:
Regen III applies healing whenever the effect duration is modulo 6. If the duration starts at 60, it heals on the first tick. Then GG removes the effect, the Kanade reapplies it, and it ends up healing on *every* tick.

Issue #2684, opened 2018-07-26
 Author forgot to determine if the state is null


zjxx:
https://github.com/Vazkii/Botania/blob/master/src/main/java/vazkii/botania/client/model/PlatformModel.java#L43

state is nullable,but author forgot to determine.
This lead to a crash.
[crash-2018-07-26_10.20.48-client.txt](https://github.com/Vazkii/Botania/files/2230437/crash-2018-07-26_10.20.48-client.txt)

PR #2751, opened 2018-10-01
Added null-check to PlatformModel for #2684

SaurabhTotey:


williewillus:
No idea how this could ever happen (since the PlatformModel is only replaced for the block model path), but whatever, this'll work

PR #2766, opened 2018-10-15
Attempt to fix ATs in jar manifest, and add sources jar

### LINKED ISSUE #2536: Manifest doesn't provide AT to dependant mods
https://github.com/yrsegal/NaturalPledge/tree/1.12

When I setup using Botania (as of [latest maven build](https://github.com/yrsegal/NaturalPledge/blob/1.12/build.gradle#L87)), despite [using dep ATs](https://github.com/yrsegal/NaturalPledge/blob/1.12/build.gradle#L39), the `botania_at.cfg` isn't copied into my `build/depAts` folder, causing all builds and runs to fail with a security exception.

This can be solved by copying `botania_at.cfg` into the `build/depAts` manually, so it's not a problem with the AT, but in how it's set up in the manifest.

Hubry:
Probably fixes #2536.
This PR should make it easier to dep on Botania with Maven:
* The main jar file it builds now no longer contains the stuff it should exclude, and has a proper manifest.
* Removed the possibility of building a deobf jar, it interferes with deobfCompile which is ugly with mismatching mappings
* Added a source jar so that IDEs can pull the sources.

I say "probably" and "attempt" because I am not 100% sure it will work despite all the testing I did and the few dozens of local maven deploys I've done.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -78,10 +78,16 @@ configurations {
 //End of the SCP config
 
 jar {
-    classifier = 'universal'
     manifest {
         attributes 'FMLAT': 'botania_at.cfg'
     }
+    
+    //rename the default output, for some better... sanity with scipts
+    archiveName = "${baseName} ${version}.${extension}"
+
+    exclude "**/*.bat"
+    exclude "**/*.psd"
+    exclude "**/*.exe"
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


williewillus:
we'll give it a shot :man_shrugging:

PR #2768, opened 2018-10-23
Tweaks to Thaumcraft integration

Hubry:
* Add brew aspects, generated from the potion's ingredients similiarly to vanilla potions.
* Add Spectrolus aspects, as I missed that one...
* Tweak a few aspect lists.
* Move TC brew recipes to init - registering them in postinit was too late to assign aspects to the brew, as aspect registration is done in Thaumcraft's postinit which runs earlier. Should not have any side effects.
* Replace wood and stone stacks with their aspects, avoids livingwood having just Victus in some modpacks.
* Reduce the amount of wildcard metadata, which can rarely cause some minor issues with items not having expected aspects and is just unnecessary for many things.
* Remove aspects from entities that are just entity form of items. I was getting annoyed around Corporea sparks when using my Thaumometer.

Beta 26 added some new stabilizer mechanics and I am wondering if it is worth messing with (for example, tweaking the effects pylons have on infusion), but I am not sure what are his plans with that part of the API, so leaving that alone for now.


Hubry:
Done, I kinda forgot that my C&B fix touched the same area.

Issue #2787, opened 2018-11-27
Extreme tps issues/hangs when firing a spreader into a whole lot of resistance lenses


quat1024:
_I could have SWORN there was an issue on this, went to go discuss it with someone, and turns out it never made its way to the issue tracker! Aaa_

Aiming a mana spreader at a collection of Mana Prisms with Resistance Lenses causes the game to lagspike or even hang.

![image](https://user-images.githubusercontent.com/2068021/49063145-6ee2e080-f1e4-11e8-9cb1-35fbb9aa30df.png)

This is because the mana spreader attempts to precalculate the path of the mana burst (for "should I fire" purposes), but the huge amount of resistance lenses makes the mana burst travel slowly and almost never despawn.

Selecting a Wand of the Forest while a nearby spreader is in the above state causes unplayable client lag (around 1/8 FPS for me, with the pictured setup)

quat1024:
A simple solution is simply arbitrarily capping the amount of mana burst simulation ticks that can happen. Basically, just capping this loop

https://github.com/Vazkii/Botania/blob/968a3b4742df5c5397740c070a05777c6f97d8e5/src/main/java/vazkii/botania/common/entity/EntityManaBurst.java#L367-L368

to a fixed number of ticks. Of course this technically changes behavior, since if you set the value to 20 seconds or whatever, and aim a spreader at something where the burst would take more than 20 seconds to reach its target, it wouldn't automatically fire anymore.

quat1024:
Fixed by merging that Pr which fixes that and a few more mana burst related performance issues

PR #2788, opened 2018-11-27
Tentative mana spreader path tracing optimizations

quat1024:
Mana spreader code is lovely as we all know so I'd like to get a few more eyes on this, if at all possible? (And someone who knows how to use a profiler would be really nice, ha!)

This PR:
* caps the maximum mana burst beam-scan distance to 400 ticks into the future, preventing big lags from https://github.com/Vazkii/Botania/issues/2787
  * This changes gameplay behavior but nobody will actually care
* stops beam-scan mana bursts from searching for entities (this information was simply thrown away, resulting in lots and lots and lots of wasted calls to `getEntitiesWithinAABBExcludingEntity`)
  * The Tripwire lens still works btw
* stops beam-scan mana bursts from sending "pings" to the mana spreader - this information was also simply thrown away, resulting in a lot of useless calls to `getTileEntity`
  * This is something I am unsure about, maybe something breaks here.
* comments out "inWater" related code from all mana bursts - I was never able to make a mana burst actually appear `inWater`, no matter how I aimed it at a pool of water.

Potential future areas of spreader/burst optimization (read: maybe don't merge yet?):

* is PositionProperties a hot spot? lots of little allocations there I'm sure
* what about `getLensInstance`?
* `onEntityUpdate` seems to be doing a *huge* amount of useless things. Running particles? Riding entities? Fire damage? DEFINITELY check this out. Lots of stuff that's not really that bad on its own but when you have a bunch of mana bursts doing these pointless checks every tick + spreaders occasionally doing 400 or so in a single tick... it can add up

Feel free to look in to any of those areas - I'll probably do it myself later when I get the time, but it's like 2:30am

### DIFF
--- a/src/main/java/vazkii/botania/common/core/handler/ConfigHandler.java
+++ b/src/main/java/vazkii/botania/common/core/handler/ConfigHandler.java
@@ -72,6 +72,7 @@ public final class ConfigHandler {
 
 //	public static boolean versionCheckEnabled = true;
 	public static int spreaderPositionShift = 1;
+	public static int spreaderTraceTime = 400;
 	public static boolean flowerForceCheck = true;
 	public static boolean enderPickpocketEnabled = true;
 
@@ -199,6 +200,9 @@ public final class ConfigHandler {
 
 		desc = "Do not ever touch this value if not asked to. Possible symptoms of doing so include your head turning backwards, the appearance of Titans near the walls or you being trapped in a game of Sword Art Online.";
 		spreaderPositionShift = loadPropInt("spreader.posShift", desc, spreaderPositionShift);
+
+		desc = "How many ticks into the future will mana spreaders attempt to predict where mana bursts go? Setting this lower will improve spreader performance, but will cause them to not fire at targets that are too far away.";
+		spreaderTraceTime = loadPropInt("spreader.traceTime", desc, spreaderTraceTime);
 
 		desc = "Turn this off ONLY IF you're on an extremely large world with an exaggerated count of Mana Spreaders/Mana Pools and are experiencing TPS lag. This toggles whether flowers are strict with their checking for connecting to pools/spreaders or just check whenever possible.";
 		flowerForceCheck = loadPropBool("flower.forceCheck", desc, flowerForceCheck);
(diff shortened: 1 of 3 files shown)


williewillus:
1. I would recommend against overriding things randomly to noop them unless it produces a real, measurable impact. It's more to maintain and keep track of when porting
2. If I recall correctly, the data manager already stores things in deserialized form internally, so this wouldn't help much
3. Yes, the simulation time should be configurable I think.

Hubry:
Profiling definitely is a good idea and Quat actually wanted to do that before doing more in this area (and I had to remind him of the profiler again).

Anyways, I did a little bit of profiling with Spark with the latest Botania release. Small testing setup of: 3 mana spreaders with resistance lenses pointing directly into a chain of 7 resistance prisms, with 7 blocks of space before a wall. 
For client thread profiling, held a wand of the forest and did `/sparkclient start --include-line-numbers --timeout 60 --interval 1` while looking at the setup.
For server thread profiling, set up a few pistons moving blocks just before the wall on a 3 second timer and did `/spark start --include-line-numbers --timeout 360 --interval 1`.

[Client results](https://sparkprofiler.github.io/#Zn9AiHZL5z), [server results](https://sparkprofiler.github.io/#f9sklj71Yt).
Seems like checking if the burst is in lava, entity search as well as all data manager calls combined are somewhat heavy (seems like data manager might be the heaviest thing when all the calls are added), while PositionProperties has a negligible impact. I'd say no-oping some of these things I mentioned might be a viable option.

quat1024:
Vazkii likes it so we're going to merge it pretty much as-is, hope that's okey:eyes:

I added a configuration option for the spreader trace time (there's no min or max value on it, I can trust people to be sensible here :stuck_out_tongue: ).

I also nooped `isInLava` since it's a pretty free fix for a relatively expensive function (according to Hubry's profiling about as bad as the pointless `getEntitiesWithinAABBExcludingEntity` checks, ouch)

> I would recommend against overriding things randomly to noop them unless it produces a real, measurable impact. It's more to maintain and keep track of when porting

Good point - I added a comment to the `isInLava` override to hopefully explain why it's there and I'll be more careful about this type of thing ifwhen I revisit optimizing.

PR #2818, opened 2019-01-03
Cleanup buildscript, add api jar, update gradle

Hubry:
* Gradle updated to 4.9, let's not stay with 2.7 forever.
* Added an api jar containing api classes + source, done similarly to Forestry and JEI. Should be automatically deployed to Jared's maven when built there. I can find at least 3 mods that would enjoy having the api on maven instead of having it in their sources.
* Replaced the weird gradle stuff with a thing taken from Vazkii/Patchouli#2. Removed the forgecraft task while doing that as it wasn't used anyways.
* Replaced deprecated << uses with doLast (it's the same thing, but it doesn't make Gradle complain about deprecated stuff and highlights properly in my IDE), and minor styling tweaks.

Output should be the same as before this PR.


williewillus:
yay someone who knows gradle

Hubry:
I wouldn't say I know gradle but I feel like I'm getting the hang of it :P

PR #2826, opened 2019-01-16
fix #2825 (manasteel and elementium shears drop apples and saplings)

### LINKED ISSUE #2825: Shears and Leaf Drops
# Version Information
Forge version:  14.23.5.2807
Botania version:  r1.10-358

# Further Information
Link to crash log: it's not a crash
Steps to reproduce:
1. Take a manasteel or elementium shear.
2. Break leaves until you get a sapling or apple.

What I expected to happen:
You only get leaves. No apples or saplings should drop.

What happened instead:
I got saplings and apples from breaking leaves (along with the leaf).

tcheinen:
The onBlockStartBreak function for manasteel and elementium shears didn't return true so the game treated it like it was being broken with your fists.  Since the shears did drop leaves this would allow someone to break the same leaf block repeatedly for materials.


williewillus:
💢

tcheinen:
oh, oops.  Sorry about that.  I definitely tested it so i'm not sure how it got past me.  I suppose it's possible I just missed it, but it seems pretty obvious.  I'll make sure to test more extensively in the future anyway.

PR #2845, opened 2019-02-12
Change how NBT matching works, now it's faster and works better with nested tags

quat1024:
This is an upstream fix of this issue in Botania Tweaks: https://github.com/quat1024/BotaniaTweaks/issues/42

The idea is that when an item has an NBT tag in a recipe, items must have *at least* those NBT tags. In other words, the NBT tag on the recipe item has to be a subset of the NBT tag on whatever item the player throws in. This makes it ok to not specify *all* the nbt tags *ever* on an item when writing recipes, which is what players would expect anyways.

However, previously this did not work properly with nested tags. An item with a tag like `{out:{in:1,extra:2}}` would match a recipe with a tag like `{}`, but would not match `{out:{in:1}}`. This creates problems when trying to create recipes for items with lots of weirdly structured NBT data (the use case in the original issue is bees lol `oof`)

I broke this code out into a separate function, too, so I can call it from btweaks next time we both update, and stop duplicating it.

### DIFF
--- a/src/main/java/vazkii/botania/api/recipe/RecipePetals.java
+++ b/src/main/java/vazkii/botania/api/recipe/RecipePetals.java
@@ -15,6 +15,7 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraftforge.items.IItemHandler;
 import net.minecraftforge.oredict.OreDictionary;
+import vazkii.botania.common.core.helper.ItemNBTHelper;
 
 import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


williewillus:
i wonder if vanilla/forge's nbt ingredients are susceptible to this? I'm looking into moving to those in 1.13.

quat1024:
tbh when writing this, I get the impression that this isn't exactly a rare problem but wasn't really able to find any methods that matched nbt tags in this way

codewarrior0:
Bees are an odd case. The nested tags include a list tag, and neither the old approach nor the new one will do a subset comparison on two list tags. To rewrite your example: an item with `{out: ["in", "extra"]` should match a recipe with `{out: ["in"]}`. There's a reason why list tags *shouldn't* be allowed to do a subset (or rather, startswith) match, but I can't quite put it into words right now.

PR #2859, opened 2019-02-21
Fix corporea interceptor causing wrong count on index requests

Hubry:
If the interceptor triggers a funnel making a request on the same network, it will overwrite the request counts with last funnel's request counts.

Despite getting 0 items from the request, the message will say that I have gotten the same amount as the last triggered funnel and had a bunch accessible.
For example, where I have a bunch of funnels requesting 1 snow and a double chest half filled with snow:
obraz

This PR swaps the order of interception and assigning this count to fix the message to correct it.


williewillus:
these global statics need to just die, but can't do it in 1.12 due to backward compat -.-

PR #2867, opened 2019-03-06
Add a hook for hidden lexicon categories

Hubry:
Adds a simple method that lets a lexicon category say if it should be displayed in the category index. The use case is quat's addon Incorporeal adding enough Corporea stuff to justify moving it to a separate category, but that means displaying an entirely empty category if the lexicon isn't upgraded yet. Doing this externally involved some pretty ugly removal and rearrangement of buttons in the index.

@quat1024 Does this appear alright for you? 🤠


quat1024:
Oh wow, totally missed this message. I blame github.

This looks great!

PR #2880, opened 2019-03-30
ItemEntity Limit for ManaPool Tile

ben-mkiv:
this commit adds a stack limit to craftingrecipes which have a output >1 (e.g. coal, redstone, glowstone dust duping) to pretend spawning of insane amounts of item entities in the world.

only applies if a recipe would output more than 1 item
only checks 2 block radius around the mana pool for items
can be disabled if the config value is set to 0 (defaults to max. 64 entities/stacks around the pool)
only checks on the first craft done and if the total amount of additional spawned items is larger than the limit


ToMe25:
You might want to increase the max item number default to 128 and resolve the merge conflicts :)

ben-mkiv:
maybe... if someone from the Botania team signals that they would merge it...

PR #2881, opened 2019-03-31
When matching recipe tags, also match the first items of nested lists

codewarrior0:
Mostly useful when adding CraftTweaker recipes which have Forestry bees
as inputs, since the bee's chromosomes are in a list and the recipe
only needs to match some of the item's chromosomes.

As requested here: https://github.com/quat1024/BotaniaTweaks/pull/48


Yoghurt4C:
waiting for this to be pulled like
maneko_small

williewillus:
seems a bit weird to just check the first element, but sure

PR #2894, opened 2019-04-22
Fix sticky comparators on corporea retainers

### LINKED ISSUE #2893: Corporea retainers don't clear their comparator value after fulfilling the request
I think this one's my fault when I prd this way back :|

quat1024:
Completing a request in a corporea retainer now actually clears the comparator value of it... Fixes #2893

> who was the loser who added this bug

> `git blame`

> oh


williewillus:
nice

Issue #2882, opened 2019-04-04
Armors are ignored in some other mods' calculations (includes fix)


MoriyaShiine:
# Version Information
Forge version:  Any
Botania version:  Any

# Further Information
Steps to reproduce:
1. Put on any of the mod's armors
2. See your armor increase
3. Find out that mods such as Astral Sorcery and Monk ignore the armor value because of ISpecialArmor and are incredibly overpowered

How to fix:  
  In ItemManaSteelArmor, return 0 in ISpecialArmor#getArmorDisplay()
  In ItemManaSteelArmor, delete the override of ItemArmor#getAttributeModifiers()
  On line 63 of ItemTerrasteelArmor, instead of using getArmorDisplay(), use damageReduceAmount

williewillus:
> because of ISpecialArmor and are incredibly overpowered

not sure what this means.

MoriyaShiine:
In Astral sorcery, there's a player infusion that removes all armor points and converts it into health. With armor like Botania's current one, it gives you the extra health but doesn't remove the armor, I'm assuming because it removes the attribute which doesn't exist anyway and it gets player total armor.

The Monk mod has things that only activate if you have no armor, same effect, Botania armor doesn't trigger that.

PR #2895, opened 2019-04-24
fix armor

MoriyaShiine:
you can ignore the launch profiles I accidentally pushed them lol


williewillus:
remove the launch profiles

williewillus:
also please describe the issue you're fixing

MoriyaShiine:
it fixes this issue: https://github.com/Vazkii/Botania/issues/2882

Certain mods don't check for ISpecialArmor so when doing effects that check for armor, Botania armor is ignored. Ex in the monk mod, some effects only work if you're not wearing armor, but they work with Botania armor due to how it's implemented.

williewillus:
you changed the whitespace of the entire file making it hard to see the real diff, revert those whitespace changes

Issue #2897, opened 2019-04-27
Gaia Guardian ritual not workng


mrdeadlocked:
# Version Information
Forge version:  14.23.5.2836
Botania version:  r1.10-361
Modpack: Dragons Dungeons and Space Shuttles (https://minecraft.curseforge.com/projects/dungeons-dragons-and-space-shuttles 4.11a)

![2019-04-27_12 59 49_2](https://user-images.githubusercontent.com/461352/56852809-b2558e00-68ed-11e9-8831-811ac24aa88b.png)

# Further Information
Steps to reproduce:
1. Setup Gaia Ritual
2. Try to summon

What I expected to happen:
Gaia Guardian spawns

What happened instead:

Ritual says cannot make

mrdeadlocked:
It appears that my previous arena layout does not meet the new stricter criteria. I had to lower the beacon/pylons down one. Sorry for this bad report. :-)

williewillus:
@quat1024 should this have happened?

quat1024:
I've seen a bunch of people on reddit with the same problem. I'll change it. I didn't account for the floor being 1 lower than how I and everyone else I play with builds an arena tbh :sweat_smile:

PR #2899, opened 2019-04-28
Gaia guardian arena stuff again, orz

quat1024:
* Slightly less strict arena floor check.
  * In every column, you need a block one *or two* blocks below the height of the beacon.
  * (Or a block in both positions. But you need at least one.)
  * This lets you build an arena with the beacon base placed *on* the floor, instead of inset into the floor like the multiblock preview shows, which seems to be a popular method of building the arena after all (rel: #2897)
* Minor changes to accommodate for low-floor fights.
  * Block-breaking now happens only *above* the height of the beacon (== changed to >=, basically)
  * GG will move one block down if the chosen TP destination is floating in the air.
* Quality of life improvements to the arena check.
  * Improper pylons now show particle effects for where the pylons are supposed to go.

### DIFF
--- a/src/main/java/vazkii/botania/common/entity/EntityDoppleganger.java
+++ b/src/main/java/vazkii/botania/common/entity/EntityDoppleganger.java
@@ -10,6 +10,7 @@
  */
 package vazkii.botania.common.entity;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


mrdeadlocked:
Arena changes would be very nice :-)

PR #2913, opened 2019-05-30
Move equippable handling back to item classes, add a handler for installs without Curios

Hubry:
* All the nested curio classes are merged back into their items
* Most checks for Curios being present are removed
* Added handling for installs without Curios - in that case, items in the hotbar work as if they were equipped.
* One forgotten belt is now a curio

Draft PR because there are a few things to fix up, namely Ring of Odin probably gives health if in hotbar of wrong player, etc. and possibly I might include render fixes if I figure out how to fix the bauble renders. At least the console spammy ones with GL stack overflows.


williewillus:
big thank, i was gonna split them back out myself but didn't have motivation to

Hubry:
I did some todos, didn't end up fixing the renders yet so if you happen to merge this I might just make a second PR. I did some testing of them and need to crosscheck with 1.12 anyways.

williewillus:
in that case, I'd handle all our rendering ourselves instead of going through curios

Hubry:
I think I might've screwed up the retarget a bit, it's really messy

PR #2916, opened 2019-06-02
1.13 crafting recipe fixes

Hubry:
Some tedious recipe stuff, because I was annoyed by the giant amount of log spam related to recipe parsing and decided to silence most of it. In short:
* Item renames, mostly `word1word2` -> `word1_word2` or `silver` -> `light_gray`
* Data tag removed, uses migrated to flattened items/tags (floating flowers mostly) or NBT tagged items (flugel tiara)
* Register shapeless mana upgrade recipes and fluxfield condition, because those were missing. I hope they were not omitted on purpose.
* Lastly, JEI no longer gets raw block instances as recipe catalysts. (I threw in support for JEI bookmarks for the corporea index keybind too as it was just 2 lines.)

The only remainder now is the brass tag being missing on the cacophonium, but that's one stack trace per log compared to the ~80 from before.


williewillus:
big thank

PR #2927, opened 2019-06-15
fix exoflame compatible blocks consuming fuel even when there is mana…

MoriyaShiine:
… available


williewillus:
explain your fixes please

MoriyaShiine:
It changes instances of IExoflameHeatable to run their code 1 tick earlier, just like the furnace does in the code above it. This fixes the issue where mods that add compatibility to the exoflame consume their fuel before the exoflame has a chance to check and refill the burn timer.

Hubry:
Just noticed I did this when moving exoflames to caps :V

Issue #2945, opened 2019-07-24
Cacophonium acts like a shield in 1.12


TangoCheetahs:
# Version Information
Forge version:  14.23.5.2838
Botania version:  r1.10-361

# Further Information
Link to proof: https://youtu.be/a_49Nytg1QQ
Steps to reproduce:
1. Hold Cacophonium
2. Hold Right Mouse/Use
3. Cacophonium is a shield

What I expected to happen: to take damage

What happened instead: I wasn't taking damage

TangoCheetahs:
Bump

Issue #2804, opened 2018-12-12
Leaving the game with the Ring of Chordata's night vision active grants effectively infinite night vision


Hubry:
## Version Information
Forge version: 14.23.5.2772
Botania version: r1.10-358

## Further Information
Link to crash log: n/a
Steps to reproduce:
1. Equip the Ring of Chordata, make sure you have mana
2. Jump into water, make sure you got night vision from the ring
3. Log off/leave the game/etc
4. Rejoin the world/etc

What I expected to happen: Ring continues to grant night vision only when underwater

What happened instead: The night vision effect persists with effectively infinite duration even if the ring is taken off, and stays until you die/cure with milk/etc.

Suspected cause is that potion effects with negative strength are saved as effects with positive strength instead when serialized. Some possible solutions would be to either stop giving a negative level of the effect (which would probably break the intended way it works, as it would interrupt night vision from potions, etc), or give the effect with relatively low duration and make it refresh when it gets too low.
I would try to fix this myself but it's getting late and I'm lazy :V

spylor:
Confirmed

PR #2947, opened 2019-07-30
Fixed issues #2945 and #2804

PaaaulZ:
For #2804 : Since the ring checks **every tick** for the working conditions we can give night vision for a small amount of time instead of giving an infinite effect. This way Night Vision can still expire when leaving the water and not get stuck by quitting the game.

For #2945 : Was that intended to also work as a shield?

---

Not big bugs but a little fix does not hurt anybody.

### DIFF
--- a/src/main/java/vazkii/botania/common/item/ItemCacophonium.java
+++ b/src/main/java/vazkii/botania/common/item/ItemCacophonium.java
@@ -121,12 +121,6 @@ public class ItemCacophonium extends ItemMod {
 			list.add(I18n.format(ItemNBTHelper.getString(stack, TAG_SOUND_NAME, "")));
 	}
 
-	@Nonnull
-	@Override
-	public EnumAction getItemUseAction(ItemStack par1ItemStack) {
-		return EnumAction.NONE;
-	}
-
 	@Override
 	public int getMaxItemUseDuration(ItemStack par1ItemStack) {
 		return 72000;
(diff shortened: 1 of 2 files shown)


TheWhiteWolves:
Have you tested to check that the small amount of night vision time you have set is long enough not to cause the flashing effect that it gives when its running out.

PaaaulZ:
I don't know how I did not notice that... The flashing effect comes at 10 seconds I think, increasing the duration to 600 (30 seconds) removes the flashing. Closing and reopening the game will only give 30 seconds of free nightvision instead of infinite, it's still a fix even if not beautiful. If you have other ideas let me know or else feel free to ignore the pull request, I just tried to help sorry I did not notice that.

williewillus:
for #2945 that was so the arm pose for the item shows up as "blocking", it's dumb the game ties "shieldness" to this method too

src/main/java/vazkii/botania/common/item/ItemCacophonium.java
```diff
@@ -124,7 +124,7 @@ else if(getSound(stack) != null)
 	@Nonnull
 	@Override
 	public EnumAction getItemUseAction(ItemStack par1ItemStack) {
-		return EnumAction.BLOCK;
+		return EnumAction.NONE;
```
williewillus: the default is none so just remove this override entirely

PR #2977, opened 2019-09-20
Port lexicon pages to Patchouli, improve generator code

Hubry:
* Created components for: 
  * a rotating list of items. Two variants are provided - one which takes a recipe, and one which takes a list of ingredient strings. The second one is basically only provided for modpack authors.
  * a mana bar. 
  * terra plate rendering. By default it renders the terra plate, but the blocks can be replaced with proper parameters. Required as blocks would weirdly overlap if stacked manually.
  * item list - created only because Patchouli will refuse to render air stacks as one of the conversion steps in item components is a conversion to an ingredient, which strips the air stacks from the rendered result. Marked deprecated.
* Created template processors for brews, elven trades, mana infusion, petal apothecaries, runic altars, and pages with multiple recipes on a single grid.
* Created templates for each of the above, as well as:
  * Lore pages - the background from old elven pages, as well as the elven garde seal page. The seal page is has a flipped background, and a flipped page without the small seal on the bottom is made for the left side.
  * Terrasteel - you can't just make a page with a custom arrangement of components without a template, so this is a template.
* Generator code now corrects camelCase names, accounts for multiblocks, and dumps lists of recipes as semicolon-separated lists, as component processors have a very limited access to the component json objects and cannot access lists.
* Very minor related tweaks - `IModRecipe#getId#; giving one petal recipe a group, minor tweaks to some lexicon entries, etc.

Things remaining:
* Assigning proper headers to all pages with multiple recipes
* Tying knowledge to advancements?
* Updating text in a few spots and tweaking pages, for example the head recipe would be switched to a link button page.
* Sorting recipes in places where they were ordered by something unstable like a tag
* Special reference features mentioned in README


williewillus:
lmk when this is ready

src/main/java/vazkii/botania/client/patchouli/component/ItemStackListComponent.java
```diff
@@ -0,0 +1,69 @@
...
+	public void render(IComponentRenderContext context, float pticks, int mouseX, int mouseY) {
+		int index = (PatchouliUtils.getBookTicksElapsed(context) / 20) % stacks.length;
+		context.renderItemStack(x, y, mouseX, mouseY, stacks[index]);
+	}
+
+	private static final Method splitStacksFromSerializedIngredient = ObfuscationReflectionHelper.findMethod(
+			vazkii.patchouli.common.util.ItemStackUtil.class, "splitStacksFromSerializedIngredient", String.class);
+
+	//TODO stop reflecting into Patchouli
+	private static String[] splitSerializedStacks(String ingredient) {
```
Hubry: I know this reflects into Patchouli, I was hoping to avoid it but I prefer this to copying that big chunk of logic and ideally neither will be required later (this entire class, really).

PR #2980, opened 2019-09-30
Ability to disable petal planting 

igormich:
I added ability to disable planting petal in the configuration file (it is still enabled by default). I think this option will be useful for hardcore mods-pack. Maybe this change looks small, but ability to copy instead of searching or growing random mystical flowers reduces difficulty.
Another option that I see is to add more similar changes to increase the complexity of the game and add one option in the configuration file "hardcore.enabled = false" to control all changes.


williewillus:
This can be done in a separate mod by using PlayerInteractEvent, so there's no need for this to be in botania itself.

PR #3019, opened 2019-11-16
Added "Wear It or No Balls", also fixed spectrolus being a trashcan for all items in range

WittySwat:


Hubry:
Kinda thinking, do we want to hide this recipe from JEI?

PR #3035, opened 2019-11-29
Adjust head bauble rendering so it doesn't look as broken

Hubry:
...until you sneak or trapdoor yourself or swim, so obviously still needs some adjustments, and I don't know rendering much. 

All head baubles were missing the head adjustment that was done before all head baubles in 1.12.


williewillus:
thank

PR #3059, opened 2019-12-20
Fix red string container destroying caps and popping off sparks sometimes

Hubry:
Two parts to this:
* If you moved the container away from its chest using something like Quark's piston pushing (looking at code, instead you could probably unload the chunk with the container but not the chest), the cap of the chest would be invalidated, destroying it and making the block incapable of doing anything not using IInventory. So we instead pack the cap in another optional. If it somehow gets destroyed without getting invalidated, the new optional gives an empty handler.
  * It's all cached so we don't create however many instances of lazy optionals every tick or anything, unless the thing you used it with doesn't cache its optionals. Please review this though.
* If you replaced the chest with another TE that doesnt have the item handler cap (for example, pushing the chest with a piston thanks to Quark), sparks would pop off, which is supposed to never happen nowadays, so we give out the empty handler.


williewillus:
christ this cap stuff didn't work out as beautifully as they said it would at all >.>

PR #3060, opened 2019-12-23
Overhaul Corporea to pass around CorporeaRequestMatchers instead of Objects

Alwinfy:
Notes:
 - This breaks some API stuff (sorry, getInventoriesWithItemInNetwork), but that's OK because nobody's written a 1.14 addon for Botania anyway
 - This also makes existing Retainers lose their currently-remembered requests (but AFAIK they're broken on HEAD rn anyway)
 - I put the current matchers in static subclasses CorporeaRequestDefaultMatchers for the sake of compactness and `import static` them; if this style isn't OK I can split them up too
 - Yes, I know the serialization mechanic for TileCorporeaRetainer is clunky. But it works, I'm pretty sure. (Couldn't test; see above note about them being broken)

### DIFF
--- a/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
+++ b/src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
@@ -78,7 +78,7 @@ public final class CorporeaHelper {
 	 * The higher level functions that use a List< IInventory > or a Map< IInventory, Integer > should be
 	 * called instead if the context for those exists to avoid having to get the values again.
 	 */
-	public static int getCountInNetwork(CorporeaRequestMatcher matcher, ICorporeaSpark spark) {
+	public static int getCountInNetwork(ICorporeaRequestMatcher matcher, ICorporeaSpark spark) {
 		List<InvWithLocation> inventories = getInventoriesOnNetwork(spark);
 		return getCountInNetwork(matcher, inventories);
 	}
@@ -88,7 +88,7 @@ public final class CorporeaHelper {
 	 * The higher level function that use a Map< IInventory, Integer > should be
 	 * called instead if the context for this exists to avoid having to get the value again.
 	 */
-	public static int getCountInNetwork(CorporeaRequestMatcher matcher, List<InvWithLocation> inventories) {
+	public static int getCountInNetwork(ICorporeaRequestMatcher matcher, List<InvWithLocation> inventories) {
 		Map<InvWithLocation, Integer> map = getInventoriesWithMatchInNetwork(matcher, inventories);
 		return getCountInNetwork(matcher, map);
 	}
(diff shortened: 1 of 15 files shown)


src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
```diff
@@ -20,24 +20,20 @@
 import net.minecraft.util.math.MathHelper;
 import net.minecraft.world.World;
 import net.minecraftforge.common.MinecraftForge;
+import static vazkii.botania.api.corporea.CorporeaRequestDefaultMatchers.*;
```
Hubry: Static imports go below all other imports

src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaRetainer.java
```diff
@@ -10,46 +10,56 @@
...
 
 	private boolean pendingRequest = false;
 	private BlockPos requestPos = BlockPos.ZERO;
-	private Object request;
+	private CorporeaRequestMatcher request;
 	private int requestCount;
 	private int compValue;
+	
+	static {
+		addCorporeaRequestMatcher("string", CorporeaStringMatcher.class,CorporeaStringMatcher::createFromNBT);
```
Hubry: formatting

src/main/java/vazkii/botania/api/corporea/CorporeaHelper.java
```diff
@@ -77,57 +73,57 @@
...
 	/**
 	 * Gets the amount of available items in the network of the type passed in, checking NBT or not.
 	 */
-	public static int getCountInNetwork(ItemStack stack, Map<InvWithLocation, Integer> inventories, boolean checkNBT) {
+	public static int getCountInNetwork(CorporeaRequestMatcher matcher, Map<InvWithLocation, Integer> inventories) {
 		int count = 0;
 
-		for(InvWithLocation inv : inventories.keySet())
-			count += inventories.get(inv);
+		for(int value: inventories.values())
```
Hubry: formatting

Hubry:
Also: 
* Would it be worth it to move all the `Map<InvWithLocation, Integer>` to FastUtil `Object2IntMap`s? 
* JEI requesting could probably be moved into a packet or something to request a stack instead of requesting the stack's name. A fairly annoying thing is that you can't request a specific item if two share a name. This could probably go into another PR, I might even do it at some point.

src/main/java/vazkii/botania/api/corporea/CorporeaRequestMatcher.java
```diff
@@ -0,0 +1,36 @@
...
+	 */
+	public boolean isStackValid(ItemStack stack) {
+		return false;
+	}
+
+	/**
+	 * Serialize to NBT data, for the Corporea Retainer's benefit.
+	 * No-op; should be overridden by subclasses.
+	 */
+	public void writeToNBT(CompoundNBT tag) {}
```
williewillus: if these methods "should all be overridden by subclasses", just make them abstract to force implementation. no point giving a noop impl that could potentially be missed. Also make this an interface maybe?

Alwinfy: Did that, thanks.

PR #3073, opened 2019-12-27
Fix ArrayIndexOutOfBoundsException in Flower Bag.

Vaelzan:
Can't have i set to -1 and then access an ArrayList using that as the index immediately afterwards.


williewillus:
patch is fine, but shouldn't that -1 branch never be hit? there's no way to get anything that's not BlockModFlower into the bag in the first place

Vaelzan:
Happened upon shift-clicking in a tall flower (wasn't looking at the type of flower, they look similar).

https://pastebin.com/DtNV8DkN

PR #3074, opened 2019-12-27
Make lexicon.php parse Patchouli codes

Alwinfy:


williewillus:
regex, truly a write-only language

PR #3080, opened 2019-12-28
Add the missing Livingwood variants to the botania creative tab

ToMe25:
Add Planks, Mossy Planks, Framed, FramedPattern and Glimmering Livingwood to the Botania CreativeTab


Hubry:
Thanks for the PR! I went and checked what was missing... and found a bunch of other things, so I decided to fix them all in one sweep.

PR #3087, opened 2019-12-30
Add Terra Harvester

ToMe25:
This Pull Request adds a Terra Harvester, a Terrasteel Shovel with the ability to harvest an 5x5 block area of Blocks that have the harvest tool shovel or crops.
It can also convert a 5x5 block area to Grass Path or Farmland and fertilize crops for mana.

**Other Changes:**
Moved the Terra Shatterer and the Terra Truncator in some Item Lists to get all Terrasteel Tools to the same place.

**Comments:**
The Textures aren't perfect but i think they are as good as i will get them.


ToMe25:
I don't know why it shows all the changes from Hubry's last commit as changes from this PR but i hope it doesn't causes problems.

TheWhiteWolves:
The Terra Shatterer already covers this functionality.

ToMe25:
I know it kind of covers that but i think a cheaper shovel stuff only version would be nice some times.
Feel free to close this if you think it doesn't fits this mod.

TheWhiteWolves:
Would be Vaz's call on whether to add it, personally I feel its not distinguishing enough from the already available options, currently its just an aoe manasteel shovel and has nothing that the shatterer cant do better.

When it comes the the shatterer and truncator they aren't just material upgrades, they have defining features that make that tier of tools have meaning, my thought would be to review what it is a shovel could do to provide something the others can't, it may be that you keep the aoe effect but it needs something to make it not just a crappier shatterer that only works on shovel things.

Also needs a different name, terrasteel grade tools have specific names (Terra Shatterer and Terra Truncator) not just [Material] [Tool Type] format, something like Terra Excavator or one that gives a good meaning to its abilities.

ToMe25:
I was thinking about some other functionalites before i ended on this one, but i will think about other functionalities for it later today, maybe i find a better one.

I know this name doesn't fits the Tool i just didn't knew a really good one so i used this one as a placeholder, i will probably change it if i find a good functionality to match that.

The Texture is the only thing i don't have any ideas for.

TheWhiteWolves:
For the texture I'd say maybe take the handle from the shatterer and stick a terrasteel head on it for now, maybe throw in the gem from the truncator, something like that.

As for abilities, one i can think of as a quick mockup would be to make it a super farming tool, does all a shovel can but also adds in ability to till land like a hoe and maybe even act as bonemeal for the cost of mana

ToMe25:
That sounds nice, i will probably do that for the functionality and at least try the texture, but do you think it should just convert dirt into farmland and grass int grass path or should it convert both in farmland when you sneak?
Also i would still give it the AOE.

TheWhiteWolves:
mmm, theres maybe 4 ways i can see it going, either:

1. have it treat grass and dirt separately (grass -> path, dirt -> farmland)
2. have it do one by default and the other on sneak
3. remove the path function and go all in on the farming tool and always goto farmland
4. have it cycle down from path to farmland (grass/dirt -> path -> farmland), kind of like you are shaving off the topsoil with 1 click then ploughing with the second.

also with the aoe function you will want to separate it out onto an active mode like the other 2 tools so that you can maintain single mine vs aoe mine functions. (maybe integrate the hoeing into the active/non-active mode so you can aoe hoe or single hoe aswell)

ToMe25:
4. is the way the Manasteel Shovel does it, but the way i would prefer is grass -> path and path/dirt -> farmland.

Adding an active mods is probably something i should do yes.

The AOE plant harvest is something i had planned to do.

ToMe25:
I will add the bonemealing and rework the crop harvesting later i just need to do something else rn.

TheWhiteWolves:
For consistency I'd recommend maintaining the same way that the manasteel shovel does the conversion, otherwise there would be a discrepancy in how the different tiers of tools do the same action.

But afaik vanilla only allows grass blocks to be made into path so it may be a case to switch to your preferred one of (grass -> path and path/dirt -> farmland) and alter the manasteel version to match, that would more closely match vanillas function and allow for the introduction of the hoes dirt -> farmland

ToMe25:
I will change the manasteel shovel to
grass -> grass path,
grass path/dirt -> farmland,
coarse dirt -> dirt
to be consistent with the Terrasteel one and vanilla hoes.

ToMe25:
Due to reasons not completely known to me the Terra Harvester can currently only fertilize one crop at a time, but i think thats a problem with the Event handling and not my method.

ToMe25:
or atleast it didn't show the changed...
seems like it worked the first time anyways.

ToMe25:
closing this as i will put this in an addon since botania is feature frozen as vazkii says.
the changes besides the Terra Harvester are in #3130.

ToMe25:
This is now moved to [Botania-Additions](https://github.com/ToMe25/Botania-Additions).

PR #3110, opened 2020-01-04
Improve lexicon.php to handle Lexica crosslinks

Alwinfy:
Makes the web-lexicon handle internal Patchouli links properly.

Unfortunately, there's no generated database mapping patchouli-entries to lang-file names, so it's… not the most performant. (Uses about eight file_get_contents.).

### DIFF
--- a/web/module/lexicon.php
+++ b/web/module/lexicon.php
@@ -21,14 +21,14 @@
 		'right' => 'w',
 		'sprint' => 'CTRL'
 	];
-	$replacements = array_merge(array_combine(array_map("quote_keys", array_keys($default_keys)), array_values($default_keys)), [
+	$entry_cache = array();
+	$replacements = [
 		'/\$\((br|p)\)/' => "<br />",
 		'/\$\(li\)/' => " &bull; ",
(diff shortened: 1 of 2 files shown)


williewillus:
what is going on here lol

web/module/lexicon.php
```diff
@@ -69,7 +73,21 @@ function match_entry($key) {
 	function match_page($key, $entry_name) {
 		return match("botania\.page\.$entry_name\d+", $key);
 	}
-	function quote_keys($key) {
-		return "/\\$\\(k:$key\\)/";
+	function key_resolve($match) {
+		global $default_keys; // i'm sorry, i don't know php
+		return (array_key_exists($match[1], $default_keys)
+			? $default_keys[$match[1]]
+			: "UNKNOWN KEY");
```
williewillus: where is this default_keys array being filled? can it just be passed in instead of a global?

Alwinfy: It's filled on line 17-- it's a lookup table for `$(k:) codes`; no, because `key_resolve` has to take one array and return a string where it's used in preg_replace_callback, and it's more performant to keep it at filescope.

Issue #3121, opened 2020-01-06
Phantom lenses trigger hourglasses to drop more mana dust than normal


tiggerbiggo:
# Version Information
Forge version:  28.1.107
Botania version:  1.11-373

# Further Information
No crash

What I expected to happen: A mana pulse with a phantom lens should pass through a floating hourglass with mana dust inside and trigger ONE dust to drop from the top to the bottom

What happened instead: A normal mana pulse triggers only a single dust to drop, however when a phantom lens is attached this suddenly increases to 4.

Demonstration of the issue in video form: https://imgur.com/DsZnFxY

tiggerbiggo:
I had a delve into the code and it is as I suspected. I believe that every tick the phantom burst is in contact with the hourglass it triggers it regardless of if it has been triggered before. Looking at the way mana bursts give mana to things they hit, they check with the lens whether or not to die, in the case of the phantom lens it just stops it from dying on collision.

`public class LensPhantom extends Lens {
	@Override
	public boolean collideBurst(...) {
		if(!isManaBlock) {
			dead = false;
			...`

My proposed solution would be to keep a list (hashtable maybe for speed) of hourglass items on the mana burst itself, and if we hit an hourglass check if it is in the list. If not, add it and trigger the pulse. If it is, don't trigger.

Thing is even though I know Java very well i'm not a minecraft modder, technically speaking I don't know much about forge apart from how to get the dev kit up and running. I'm currently trying to get this working myself before making a PR, just wondered if there is another easier way, maybe something that is already in place that I do not know about due to unfamiliarity with Botania's code base (or Forge). If by some coincidence you are already keeping track of this somewhere else then it may be more trivial than my idea 😛 

The main reason I want this fixed so badly is because it's gonna multiply the cost of all the mana powder in my contraption by 4, so I wanna try and fix it instead of having to spend 64 dust per hourglass instead of 16. Hopefully I can contribute something to this mod as well :)

tiggerbiggo:
Proposed PR holds the last hit block position and compares it to the currently hit block. If the position is the same, we are in the same block so we skip triggering it. If the position is different, update it and trigger the new block.

PR #3124, opened 2020-01-06
Fixed phantom mana bursts triggering the same block multiple times.

tiggerbiggo:
Fixes https://github.com/Vazkii/Botania/issues/3121


tiggerbiggo:
Ignore the force push, that was to fix a mistaken commit to this PR and move it to another branch.

williewillus:
remove the comment since the code is self-explanatory enough. otherwise looks good.

williewillus:
one more thing: does this need to be saved? I'm thinking it should be

Alwinfy:
At this point, why don't we just have a helper function for serializing/deserializing a position to NBT (heaven knows enough things use it)

PR #3126, opened 2020-01-07
Add snow layering and particles to Snowflake Pendant

ToMe25:
Make the Snowflake Pendant leave a trail of snow layers wherever you go while you aren't sneaking.
Add snowflake particles to the Snowflake Pendant.
Add something about the snow layering to the Lexica Botania.


williewillus:
should the particles only be emitted when not sneaking?

ToMe25:
I was thinking its always cold so no, but after reading the lexica entry again i think yes, so i will change that.

PR #3132, opened 2020-01-09
Bring the Timeless Ivy back

ToMe25:
This PR brings the Timeless Ivy from older versions of botania back as it is still very usefull, especially with other mods, by allowing you to repair your tools and armor with Mana rather than XP that you need for enough other stuff anyways.
If nothing went wrong thats the only change.
If this is considered not fitting for the mod anymore, or won't get added because botania is "feature frozen"(which i think shouldn't be a problem because its only "porting" an old feature) please tell me so i can add it to my addon :)
Also i am not sure whether the Lexica entry should be in the ender category as it currently is or in the alfhomancy one, feel free to leave feedback about that.


TheWhiteWolves:
I'd say may be better in your add-on, if I remember correctly vazkii retired it in favour of the mending enchant when that was added to vanilla

ToMe25:
I know why the Timeless Ivy got removed, but i fell it still has its place as its more usefull than mending imo.
Also i will move it to the addon i am making for the Terra Harvester if i get told it won't get merged, but i was thinking i should at least try it here.

ToMe25:
I will move this to my addon in if it doesn't get merged until the next update.

ToMe25:
This is now moved to [Botania-Additions](https://github.com/ToMe25/Botania-Additions).

PR #3136, opened 2020-01-10
Generating flowers give comparator output based on mana

tiggerbiggo:
0 redstone signal if flower is empty of mana, 15 redstone signal when
full

### DIFF
--- a/src/main/java/vazkii/botania/api/subtile/TileEntityGeneratingFlower.java
+++ b/src/main/java/vazkii/botania/api/subtile/TileEntityGeneratingFlower.java
@@ -305,4 +305,9 @@ public class TileEntityGeneratingFlower extends TileEntitySpecialFlower {
 	public int getMana() {
 		return mana;
 	}
+	
+	@Override
+	public int getComparatorInputOverride() {
+		return (int)(((double)getMana()/getMaxMana())*15);
+	}
 }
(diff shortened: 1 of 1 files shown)


williewillus:
use TilePool.calculateComparatorLevel instead of redoing the math, the method handles some edge cases (like when the fraction is so small it rounds towards zero but there's mana there so the output *shouldn't* be zero).

williewillus:
actually, since this is an api class, you might want to move that TilePool method somewhere in the API and update calls to it. Feel free to put it wherever for now, the API needs to be cleaned up anyways.

tiggerbiggo:
Ok, I changed the method used, but given my limited knowledge of Botania's API I do not feel I am qualified to move classes around, I can't really say I feel comfortable changing more than I need to, although I could potentially just do a refactor in intellij. Should that not be in a separate change though, or do you think it should be bundled in this one?

tiggerbiggo:
Don't approve this yet, there's a bug. It seems to update the comparator when it is generating mana, but doesn't update when it's sending that mana to a spreader. I got around this by adding a clock that pulses on the other side of the flower and gives it a block update, but i'm not sure how to solve that, since i'm new to mc modding. I'll do a bit of research but if anyone knows why this might be let me know and i'll make a change.

PR #3157, opened 2020-01-25
Add a recipe for the Haert of the Sea

ToMe25:
* Add a Alchemy recipe for the Heart of the Sea(mainly for GoG).


williewillus:
let's also make the recipe gog only: move the page to the gog page and move the registration code into that if statement below

ToMe25:
yes, it would probably be too op, and don't make much sense in a normal world, which is why i made a different recipe for it originally.

Issue #2435, opened 2017-11-04
The Agglomeration Plate should work with a Shimmerock base


ConnorPr:
Forge version: 1.12.2-forge1.12.2-14.23.0.2512 (outdated because Extra Utilities 2 breaks otherwise)
Botania version: r1.10-349

With the checkerboard pattern when making the "ritual" for Terrasteel, you should also be able to use Shimmerock since it matches the Gaia Spreader/Fabulous Mana Pool Theme. I know this isn't an issue and it's more of a suggestion, I just have no idea where to post suggestions since im relatively new to the issue tracker :)

codycoolwaffle:
Can someone label this as "idea"

PR #3158, opened 2020-01-25
Allow the Terrestrial Agglomeration Plane to use Shimmerrock

ToMe25:
Allow the Terrestrial Agglomeration Plane to use Shimmerrock in place of Livingrock.
This got suggested by #2435.
Should the Tag with Livingrock and Shimmerrock be the old Livingrock only one that isn't used for anything else as far is i know or should that get moved to a new Tag?
This depends on Vazkii/Patchouli#188 but i think i could rewrite it easily to not need that.
Also this makes the Multiblock use the Lapis Block Tag instead of the Vanilla Block only.

### DIFF
--- a/src/main/java/vazkii/botania/common/block/tile/TileTerraPlate.java
+++ b/src/main/java/vazkii/botania/common/block/tile/TileTerraPlate.java
@@ -58,8 +58,8 @@ public class TileTerraPlate extends TileMod implements ISparkAttachable, ITickab
 					}
 			},
 			'P', ModBlocks.terraPlate,
-			'R', "#botania:livingrock",
-			'0', "#botania:livingrock",
+			'R', "#botania:terra_plate_base",
+			'0', "#botania:terra_plate_base",
 			'L', "#forge:storage_blocks/lapis"
 	));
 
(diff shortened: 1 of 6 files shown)


src/main/java/vazkii/botania/common/block/tile/TileTerraPlate.java
```diff
@@ -61,9 +58,9 @@
 					}
 			},
 			'P', ModBlocks.terraPlate,
-			'R', ModBlocks.livingrock,
-			'0', ModBlocks.livingrock,
-			'L', Blocks.LAPIS_BLOCK
+			'R', "#botania:livingrock",
+			'0', "#botania:livingrock",
+			'L', "#forge:storage_blocks/lapis"
```
Hubry: What this should use is something like
`StateMatcher.fromPredicate(ModBlocks.livingrock, state -> state.getBlock().isIn(ModTags.Blocks.LIVINGROCK))`
like the mana enchanter is doing for flowers.

williewillus:
let's have a specific tag (`terra_plate_base` or something)

williewillus:
Almost good - no need to have an item tag for this, just the block tag will do

PR #3165, opened 2020-02-02
Dispenser Wand Functionality for Sneak-Right Clicking.

eutropius225:
Some blocks, specifically, **Mana Pools**, **Hopperhocks**, **Spreader Turntables** and **Force Relays**, have sneak-right click functionality for the Wand of the Forest. These functionalities could not be performed by dispensers, and I thought they should. This could potentially open up some new opportunities.

**Tests**

Switching Mana Pools from emptying tablets to filling them.
prev1
Cycling Hopperhock filter modes.
prev2
Changing Spreader Turntable rotation speed and direction.
prev3
Safely breaking a Force Relay.
prev4

**Lexicon Changes**

image
image
(There's a typo on "a Mana Spreaders" here, third line)
image
image


williewillus:
Bind mode to change speed and function mode to change direction seems a bit weird.
Maybe it should just cycle?

eutropius225:
I thought it would be worth having the option for both, given that both are useful features. Then, I thought using bind/function mode would be a good way to differentiate between which to use, and have them both be available.

williewillus:
it just seems odd to shoehorn that into the bind/functional mode split

eutropius225:
If you think so. I'd make it just switch direction, rather than cycle, if dropping one is necessary. Though I find the Bind/Function Mode split fitting:

You're *binding* it to a new speed, or changing its *function* in its rotation direction, perhaps.

eutropius225:
Another way to choose between switching direction or speed would be to base it on the direction the dispenser is clicking from. Would this be better?

eutropius225:
To make clear the change: a Dispenser with a Wand of the Forest from the side will change the speed, and one from the bottom will change the direction (as opposed to the bind/function distinction before).

Hubry:
@williewillus Vaz gave a go ahead on this, please review.

PR #3171, opened 2020-02-10
Add some recipe book support

ToMe25:
Add and run a Recipe Advancement Data Generator to allow for some recipe book support.

This isn't perfect, as some recipes don't work with the Data Generator and for some reason the Recipe Book sorts recipes by the Creative Tab the output is in instead of the folder the advancement is in, if you don't want to use this because of that feel free to close this PR.


src/main/java/vazkii/botania/data/RecipeAdvancementProvider.java
```diff
@@ -0,0 +1,260 @@
...
+			list1.add(p_200244_1_.getResourcePack());
+		});
+		CompletableFuture<Unit> completablefuture = resourceManager.reloadResourcesAndThen(Util.getServerExecutor(),
+				Util.getServerExecutor(), list1, CompletableFuture.completedFuture(Unit.INSTANCE));
+		BooleanSupplier isDone = completablefuture::isDone;
+		while (!isDone.getAsBoolean()) {
+			Thread.yield();
+			LockSupport.parkNanos("waiting for tasks", 100000L);
+		}
+		return recipeManager;
```
williewillus: is this actually required to have recipe advancements? 0.o
seems extremely hacky

ToMe25: the recipes need to be loaded for the Data Generator to load them and this is as close to the Vanilla Way as i could get.

ToMe25: it should work without the Thread.yield if that is what you mean,
but without anything to do inside the while it somehow just freezes forever.

ToMe25: It would also work with a
```java
try {
    Thread.sleep(10);
} catch(Exception e) {
}
`
inside of the loop, but i am not sure thats a better solution.

ToMe25: This loop could also be replaced with `completablefuture.join();`.

Issue #3180, opened 2020-02-16
Force relays don't save the bound block upon quitting single player world


sug0:
# Version Information
Forge version:  28.2.0
Botania version:  Botania+r1.11-377
Minecraft version: 1.14.4

# Further Information
No game crashes to report.

Steps to reproduce:
Place a force relay, bind it to a block with a wand of the forest, save and exit the single player world, enter the world again, block is unbound.

What I expected to happen:
...

What happened instead:
...

auxiliary-character:
Cannot reproduce on `162519d`
[Video](https://youtu.be/100cJ7Bng5UP)

sug0:
Correct, it seems the issue is with the mod [YAMDA](https://www.curseforge.com/minecraft/mc-mods/yamda), after further testing. The chunks mustn't be properly saved. :|

sug0:
Sorry for the confusion, the issue can be closed.

PR #3182, opened 2020-02-16
Allow force relays to store their bound blocks on other dimensions

sug0:
See https://github.com/Vazkii/Botania/issues/3180

Turns out the problem lied in the fact that only the OW state was being saved, and other dimensions', modded or vanilla, wasn't.

### DIFF
--- a/src/main/java/vazkii/botania/common/block/BlockPistonRelay.java
+++ b/src/main/java/vazkii/botania/common/block/BlockPistonRelay.java
@@ -139,13 +139,13 @@ public class BlockPistonRelay extends BlockMod implements IWandable {
 
 	@SubscribeEvent
 	public void onWorldLoad(WorldEvent.Load event) {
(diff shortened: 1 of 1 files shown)


auxiliary-character:
Hey, I was actually able to confirm this one this time.

[Video.](https://youtu.be/gL_tkQRFrGU)

williewillus:
this fix is confirmed to work?
As far as I could see when reworking the WSD in 1.13, it's all stored in the overworld anyways (all dims shared the overworld one)

williewillus:
Oh, looks like they got separated in 1.14. In that case, the proper fix for this is to actually separate all the pairs into their own dimensions, but that might be a bit sketchy for backwards compatibility.
This fix might(?) work as a stopgap, but it does redundant IO and needs to be tested so that there's no data loss for sure

williewillus:
the fix *seems* to work, it'll do until 1.15

PR #3186, opened 2020-02-24
Update bottle textures to 1.14+

Alwinfy:
because they irk me that's why


williewillus:
you deleted the ender air texture without replacing it?

williewillus:
oh, nvm, github's negative diff indicator confused me.

PR #3187, opened 2020-02-25
Don't reduce stack size when re-adding previous recipe in creative.

eutropius225:
I came across this earlier today and was thoroughly annoyed when attempting to recraft in creative proved more difficult that it should be. Curretnly, reinserting a recipe required and removed the items from the player's inventory, creative mode or not. With this change, reinserting a recipe, while still requiring at least one of each item, no longer removes those items from the inventory.

Runic Altar
Runic Altar 2
Petal Apothecary

Should reinserting a recipe from creative mode instead not require even the presence of materials?


eutropius225:
Reinserting recipes no longer requires any items on the creative player's part.

PR #3200, opened 2020-03-06
Move spreader magic-numbers into the enum

Alwinfy:


williewillus:
please retarget to 1.15

PR #3241, opened 2020-04-10
Make Magnet Lens code better but actually worse if you're a masochist.

eutropius225:
This should preserve the peculiarities of the previous logic, while fixing the unreported bug where bursts not fired from Mana Spreaders would not magnetize block around the origin.

### DIFF
--- a/src/main/java/vazkii/botania/common/item/lens/LensMagnet.java
+++ b/src/main/java/vazkii/botania/common/item/lens/LensMagnet.java
@@ -48,9 +48,12 @@ public class LensMagnet extends Lens {
 					entity.getPersistentData().getInt(TAG_MAGNETIZED_Z)
 			));
 			if (!predicate.test(tile)) {
-				return;
+				tile = null;
+				entity.getPersistentData().remove(TAG_MAGNETIZED);
 			}
-		} else {
+		}
+
+		if (tile == null) {
 			for (BlockPos pos : BlockPos.getAllInBoxMutable(basePos.add(-range, -range, -range),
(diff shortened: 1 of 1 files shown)


eutropius225:
The biggest improvement would be that a stored, but otherwise invalid, magnetised position no longer causes the full iteration.

The question now is whether an invalid position should be honoured, or rebound.

williewillus:
sorry for forgetting about this.

magnetized bursts should ignore invalid positions and resume looking for a new target (ideally, forget about the invalid pos)

src/main/java/vazkii/botania/common/item/lens/LensMagnet.java
```diff
@@ -30,50 +32,59 @@ public void updateBurst(IManaBurst burst, ThrowableEntity entity, ItemStack stac
...
+		TileEntity tile = null;
+		if (magnetized) {
+			tile = entity.world.getTileEntity(new BlockPos(
+					entity.getPersistentData().getInt(TAG_MAGNETIZED_X),
+					entity.getPersistentData().getInt(TAG_MAGNETIZED_Y),
+					entity.getPersistentData().getInt(TAG_MAGNETIZED_Z)
+			));
+			if (!predicate.test(tile)) {
+				tile = null;
+				entity.getPersistentData().remove(TAG_MAGNETIZED);
```
williewillus, line 52: should `magnetized` also be set to false here so the logic below runs?

eutropius225: You're right.

PR #3257, opened 2020-04-21
apothecary more compatible with other fluid containers

cech12:
let apothecary be more compatible with other fluid containers
- filling apothecary with lava works like filling with water
- emptying apothecary with fluid containers also possible now

Edit: Can be tested with mods like [Ceramic Bucket](https://www.curseforge.com/minecraft/mc-mods/ceramic-bucket) or [Big Buckets](https://www.curseforge.com/minecraft/mc-mods/bigbuckets).


cech12:
Question related to these lines of code:
Why is the possibility to get the fluids out of the apothecary deactivated when Garden Of Glass mod is loaded?

Alwinfy:
yeah, infinite water's supposed to be gated behind Rod of the Seas

cech12:
Ah okay! :)

cech12:
But this PR don't change anything of the behaviour when Garden Of Glass is loaded. :D

PR #3268, opened 2020-04-28
Fix #3232 and missing recipe in Lexica

### LINKED ISSUE #3232: 1.15.2 - Items on Tiny Potato face wrong way
# Version Information
Forge version:  31.1.36
Botania version:  r1.15-382

# Further Information
Items placed on the Tiny Potato block face the wrong way. This can be seen with blocks like the Furnace and Chest

What I expected to happen:
Blocks face the same way that the Tiny Potato is

What happened instead:
image
They face the opposite

Partonetrain:
Tiny ez fixes


williewillus:
can you post a screenshot of the fixes #3232

Partonetrain:
Yep. 


It was rotating them as if it was the back side on every side *except* the back - flipping the condition was all that was needed

PR #3300, opened 2020-05-21
Allow ender air to be picked up from mid-air clicks, closes #3265 on 1.12.2

### LINKED ISSUE #3265: End Air is only obtainable by right-clicking blocks in the End
# Version Information
Forge version:  14.23.5.2838
Botania version:  r1.10-363

# Further Information
Steps to reproduce:
1. Go to the End
2. Right click in the air with a glass bottle
3. Observe nothing happens
4. Right click on a piece of end stone
5. Observe you obtain end air

It's notable that this was only tested in the Galaxy-Craft Lite modpack, and not with Botania alone.

What I expected to happen:
Step 2 should produce end air, as the lexica entry states.

What happened instead:
Only step 4 produces end air.

Brycey92:
If dragon breath is found nearby, the ender air handling is skipped.

Backported from MC 1.15


williewillus:
you are welcome to apply this patch and build it yourself for private usage, but 1.12 is not receiving further releases

PR #3321, opened 2020-06-07
Fix all (non-deprecation) compiler warnings

Alwinfy:


src/main/java/vazkii/botania/common/core/BotaniaCreativeTab.java
```diff
@@ -709,7 +709,7 @@ public void fill(@Nonnull NonNullList<ItemStack> list) {
 
 	private void addTag(ResourceLocation tagId) {
 		ItemTags.getCollection().getOrCreate(tagId).getAllElements().stream()
-				.sorted(Comparator.comparing(ForgeRegistryEntry::getRegistryName))
+				.sorted(Comparator.comparing(ForgeRegistryEntry<?>::getRegistryName))
```
williewillus, line 712: TIL you can supply type parameters in method refs

PR #3345, opened 2020-07-05
Add alternative constructor to accept suppliers for EffectInstances

Alan19:
The current Brew class is unable to create Brews using Effects that are registered through DeferredRegisters. By adding a supplier and modifying various fields to only use the EffectaInstance when needed through suppliers, Minecraft will not crash when trying to register a Brew using effects from other mods. 

I ran into this issue when I was trying to add brews when Botania is installed for my mod and am currently using a workaround to bypass the crash.
https://github.com/MinerArcana/Astral/compare/feature/botania_compat#diff-f4145c9694231af724e6a15feceb81e8

### DIFF
--- a/src/main/java/vazkii/botania/api/brew/Brew.java
+++ b/src/main/java/vazkii/botania/api/brew/Brew.java
@@ -12,17 +12,19 @@ import com.google.common.collect.ImmutableList;
 
 import net.minecraft.item.ItemStack;
 import net.minecraft.potion.EffectInstance;
+import net.minecraft.potion.PotionUtils;
 import net.minecraftforge.registries.ForgeRegistryEntry;
 
 import java.util.List;
+import java.util.function.Supplier;
 
 /**
  * The class for a Brew definition, each one is a singleton.
  */
 public class Brew extends ForgeRegistryEntry<Brew> {
-	private final int color;
+	private final Supplier<Integer> color;
 	private final int cost;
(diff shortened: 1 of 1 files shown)


Alan19:
After thinking about it, I realized the supplier in the alternative constructor was added to fix an issue with my mom where the effect instances were being stored as a list where the effects may not have been initialized when the brews were being registered. I could also make a variation that doesn't use a supplier in the constructor, but the switching the field to a supplier is still necessary to not crash.

src/main/java/vazkii/botania/api/brew/Brew.java
```diff
@@ -35,9 +37,19 @@
 	 * @param effects A list of effects to apply to the player when they drink it.
 	 */
 	public Brew(int color, int cost, EffectInstance... effects) {
-		this.color = color;
+		this.color = () -> color;
 		this.cost = cost;
-		this.effects = ImmutableList.copyOf(effects);
+		this.effects = () -> ImmutableList.copyOf(effects);
```
williewillus, line 42: can you make the list once and return that in the lambda instead of creating a new copy every call to the lambda?

Alan19: Does this look better?
https://github.com/Vazkii/Botania/pull/3345/files#diff-03fa907e3726ed8c32f672d932dc8752R42-R43

PR #3374, opened 2020-07-31
Refine zappyrod targeting logic

Alwinfy:


williewillus:
can you add the horse check back in (with your logic)? horses (and llamas and mules, etc.) don't extend TameableEntity

and mods could add hostile horses :wacko:

PR #3386, opened 2020-08-07
Allow paintslinger lens to dye sparks

Alwinfy:
Currently checks against `EntitySparkBase` because `ISpark` and `ICorporeaSpark`, despite their names, are disjoint

when testing i noticed that horizontally-fired bursts are at "eye-level" with the bottoms of sparks, so if you'd like to nudge those downward for reliable collisions, i can do that

diagram of what i mean by above


williewillus:
make it so that the center of the spark rests halfway through the blockspace (y+0.5). iirc it's shifted one way or another right now

also, should the burst die when it hits a spark?

Alwinfy:
> also, should the burst die when it hits a spark?

i set it not to, since sparks are kinda transparent, and so you can dye multiple at once

PR #3392, opened 2020-08-10
Items and Block textures

Ridanisaurus:
A simple texture update to some of the out-dated textures used for Ingots, Nuggets, Quartz, and Storage Blocks. The new textures are done to fit the new Minecraft texture style.


williewillus:
Sorry, there's already a retexture of the entire mod in progress by Dani, so we won't be taking these. :(

In the future, please reach out to us before starting so you won't have to waste time doing something that won't get accepted.

Ridanisaurus:
Not a problem at all. These were done already as part of a Resource Pack I was maintaining. Glad to hear a Texture Update is in the work for Botania as it is much needed in this new age of MC Graphics. Have a good one.

PR #3406, opened 2020-08-16
Refine zappyrod logic even further

Alwinfy:
Avatars defend all players, players can hit mobs angry at other players when PvP is off


Alwinfy:
actually you know what i will leave dev to local

PR #3413, opened 2020-08-17
Update ItemInvisibilityCloak.java

Keralop:
Fixed logic mana burn and increase mana


williewillus:
revert the cost change please

PR #3418, opened 2020-08-18
Revise unethical TNT check

friendlyhj:


Alwinfy:
Usually you need all three in the vicinity, yes? Otherwise you run the risk of false positives

williewillus:
^

friendlyhj:
> Usually you need all three in the vicinity, yes? Otherwise you run the risk of false positives

I know it may cause false judgment, but in most cases, not all of three are in dump-TNT machines...

PR #3428, opened 2020-08-25
Allow mooshrooms to be 'milked' instead of being sheared

Alwinfy:


Hubry (changes requested):
Looks like you didn't import `Item` and `Entity`? The build fails on it

Hubry:
The new commit title is concerning.

Hubry (approved):

PR #3434, opened 2020-08-30
Improve TNT ethics, fix world-loading deadlock.

### LINKED ISSUE #3426: Deadlock due to entropinnyum fake tnt check
# Version Information
Forge version:  32.0.108
Botania version:  Botania-1.16-402

# Further Information
Server gets stuck on launch.
Profiler indicates the check to be the cause. 
image

eutropius225:
Fixes #3426.

The current solution is to:

- make the whole thing static instead of per Entro because that's terrible all over

- always write the tag, write false if the TNT is ethical, true if it's unethical
- check if the TNT already has the tag before checking whether it's ethical
That should fix the issue ever coming up again, in worlds after this update.

However:
- worlds with Botania just added/updated with TNTs already in the world may still run into the issue

To circumvent that another check has been added:
- if the fuse is already below the max, don't bother checking

However this will still have issues from updating worlds with TNT's that haven't ticked at all since spawning.

### DIFF
--- a/src/main/java/vazkii/botania/common/Botania.java
+++ b/src/main/java/vazkii/botania/common/Botania.java
@@ -66,6 +66,7 @@ import vazkii.botania.common.block.string.BlockRedStringInterceptor;
 import vazkii.botania.common.block.subtile.functional.SubTileDaffomill;
 import vazkii.botania.common.block.subtile.functional.SubTileLoonuim;
 import vazkii.botania.common.block.subtile.functional.SubTileVinculotus;
+import vazkii.botania.common.block.subtile.generating.SubTileEntropinnyum;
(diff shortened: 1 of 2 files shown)


eutropius225:
The new solution is to just check whether the chunk the entity is in is loaded.

williewillus:
Can you post your reproduction steps? This current solution is open to exploit by just spawning TNT in lazy chunks to skip the ethical check, then loading it.

eutropius225:
To reproduce, it's enough to place and ignite a TNT block, then reload the world.

williewillus:
I wasn't able to reproduce it this way.

williewillus:
re ^ reproduced by placing an entro first

as for this fix, yolo

PR #3444, opened 2020-09-04
Bind the Eye to a head to warp to that player

Alwinfy:


williewillus:
Significant things I see on first glance:
1. This interacts strangely with the new per-world bindings. This is bound to a player...but still per-world?
2. This keeps the bound position essentially chunkloaded to check if it's a skull, which is not good.

Alwinfy:
Re: 1; If the player's in another dimension, the `getEntityByUuid` will just return `null` and you'll warp to the skull, afaik
Not sure how to mitigate 2 though

Alwinfy:
Currently no blockpos is saved when bound to a player anymore; this is so you don't see a wireframe when a skull breaks and gets replaced with another block

Alwinfy:
you know what, i've been reasonably convinced by the discord™ that this poses too big of a security risk. closing.

PR #3447, opened 2020-09-06
Allow the Obedience Stick to be used in a dispenser

Alwinfy:


src/main/java/vazkii/botania/common/block/dispenser/BehaviourStick.java
```diff
@@ -0,0 +1,39 @@
...
+	protected ItemStack dispenseStack(IBlockSource source, ItemStack stack) {
+		World world = source.getWorld();
+		Direction facing = world.getBlockState(source.getBlockPos()).get(DispenserBlock.FACING);
+		BlockPos pos = source.getBlockPos().offset(facing);
+
+		if (ItemObedienceStick.applyStick(world, pos)) {
+			return stack;
+		}
+
+		return super.dispenseStack(source, stack);
```
williewillus, line 36: So this launches the stick out of the dispenser if it fails? I'm pretty sure vanilla behaviour is to do nothing if the target is not eligible (like buckets)
See OptionalDispenseBehaviour or whatever it's called. The one with a setSuccess method

williewillus (changes requested):

williewillus:
Extend OptionalDispenseBehavior and do `setSuccessful(<the original call>)`

optional dispense behavior has logic to play a different sound if the dispense fails

PR #3449, opened 2020-09-06
Let dispensers place sparks.

eutropius225:
example


williewillus:
for the dispense behavior: https://github.com/Vazkii/Botania/pull/3447#issuecomment-687921816

also, just make separate dispense behaviors for normal and corporea. tying the items together with inheritance just to save that small amount of copying is very bleh

eutropius225:
I'd argue that it makes sense for them to be tied by inheritance anyway, but as you wish.

PR #3458, opened 2020-09-13
Add method on bursts to cast to ThrowableEntity

Alwinfy:


src/main/java/vazkii/botania/api/mana/ILensEffect.java
```diff
@@ -49,7 +48,7 @@
 	/**
 	 * Gets the amount of mana to transfer to the passed in mana receiver block.
 	 */
-	public default int getManaToTransfer(IManaBurst burst, ThrowableEntity entity, ItemStack stack, IManaReceiver receiver) {
+	public default int getManaToTransfer(IManaBurst burst, ItemStack stack, IManaReceiver receiver) {
```
williewillus, line 51: isn't this an api break

Alwinfy: i mean technically yes but are there really any 1.16.3 botania addons right now

Alwinfy:
bunp

PR #3459, opened 2020-09-13
Use -Xlint:all in build.gradle

Alwinfy:
`classfile` and `processing` are suppressed b/c those are Gradle upstream afaik; `deprecation`'s suppressed because otherwise we get about twenty angry messages about ITEP


Hubry (approved):

williewillus:
Is this enforced? (Does it fail the build?)

If not, I don't see the point as cruft will just slowly accumulate again

Alwinfy:
now it is

williewillus:
haha yes

PR #3461, opened 2020-09-13
Remove IManaGivingItem and ItemGreaterAuraRing

Alwinfy:


src/main/java/vazkii/botania/api/mana/IManaGivingItem.java
```diff
@@ -1,18 +0,0 @@
...
-package vazkii.botania.api.mana;
-
-/**
- * An item that implements this is an item that can give mana to other items.
- * This is used in botania for the terra shatterer to not be constantly
- * receiving mana from these items..
- */
-public interface IManaGivingItem {
-
-}
```
williewillus, line 18: deprecate for removal, this is breaking

PR #3528, opened 2020-11-26
Add ability to use gog level type in server.properties

### LINKED ISSUE #3511: Cant seem to get GOG world on 1.16.3 server
I've used GOG in previous versions with no issues at all but I cant seem to get a GOG world on my 1.16.3 dedicated server. It always comes up as a standard minecraft world, even when only using garden of glass and the other necessary mods needed for gog to work. Using latest versions of necessary mods, I have leveltype=botania-skyblock in server properties and yes have deleted the world since changing the server properties to that. What am I missing? Thanks in advance for your help.

MelanX:
Creating a world on server with level type set to "botania-skyblock" will now create the GoG skyblock world. 
Closes #3511

### DIFF
--- a/src/main/java/vazkii/botania/common/Botania.java
+++ b/src/main/java/vazkii/botania/common/Botania.java
@@ -287,7 +287,7 @@ public class Botania {
 					+ " \"Do not Override\". Whoever had the brilliant idea of overriding it needs to go"
 					+ " back to elementary school and learn to read. (Actual classname: " + clname + ")");
 		}
-		if (Botania.gardenOfGlassLoaded) {
+		if (Botania.gardenOfGlassLoaded && event.getServer() instanceof DedicatedServer) {
 			WorldTypeUtil.setupForDedicatedServer((DedicatedServer) event.getServer());
 		}
 	}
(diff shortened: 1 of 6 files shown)


williewillus:
Use mixins instead of AT
And if that code is copied of modeled after vanilla code specify where with a // [VanillaCopy]

MelanX:
Done (and it's not copied or modeled after vanilla code)

src/main/java/vazkii/botania/common/Botania.java
```diff
@@ -285,6 +287,9 @@ private void serverAboutToStart(FMLServerAboutToStartEvent event) {
 					+ " \"Do not Override\". Whoever had the brilliant idea of overriding it needs to go"
 					+ " back to elementary school and learn to read. (Actual classname: " + clname + ")");
 		}
+		if (Botania.gardenOfGlassLoaded) {
+			WorldTypeUtil.setupForDedicatedServer((DedicatedServer) event.getServer());
```
williewillus, line 291: this seems unsafe, what if it's an integrated server?

MelanX: true, fixed it :)

MelanX:
I think the forge update 35.1.5 made this unnecessary.

Hubry:
Superceded by d04eb2362bec6f1849660496884f9ef4fd7ced22

PR #3531, opened 2020-11-27
You can't see it, but I'm making a very angry face right now

Yoghurt4C:
changes:
-updated loom to shorten compile time (and enlarge logspam)
-move renderer registration out of the modelappender (fixes https://github.com/williewillus/botania-fabric-issues/issues/28
-added hard dependencies to fabric.mod.json
-mostly prevented the lightningsegmentgenerator from crashing
-made lightning fx render again

PR #3533, opened 2020-11-28
Fix broken CI on fabric branch

codewarrior0:
The exception seen in current CI logs:

    > java.lang.Exception: You are running Spotless on JRE 8, which limits you to google-java-format 1.7
     If you upgrade your build JVM to 11+, then you can use google-java-format 1.9, which may have fixed this problem.

This is fixed by changing `FuzzyNBTIngredient.java` to actually contain a class. Now spotless can throw all of its formatting errors, which are fixed with `spotlessApply`.

The next error appears as various phrasings of "JVM out of heap space". Fixed by adding an Xmx arg to gradle.properties.

Finally, fabric-loom is upgraded to version 0.5 for the performance increase alleged in #3531.

Sorry about the 500 file commit. You're welcome to cherry-pick the other three commits and rerun spotlessApply yourself.


williewillus:
thank

PR #3535, opened 2020-11-29
serverside duct-tape

Yoghurt4C:
this allows an smp instance to start, generate a world and accept player connections

### DIFF
--- a/src/main/java/vazkii/botania/common/crafting/RecipeBrew.java
+++ b/src/main/java/vazkii/botania/common/crafting/RecipeBrew.java
@@ -144,7 +144,7 @@ public class RecipeBrew implements IBrewRecipe {
 		public RecipeBrew read(@Nonnull Identifier id, @Nonnull JsonObject json) {
 			String brewStr = JsonHelper.getString(json, "brew");
 			Identifier brewId = Identifier.tryParse(brewStr);
-			if (brewId == null || BotaniaAPI.instance().getBrewRegistry().get(brewId) == ModBrews.fallbackBrew) {
+			if (brewId == null || !BotaniaAPI.instance().getBrewRegistry().getOrEmpty(brewId).isPresent()) {
 				throw new JsonParseException("Unknown brew " + brewStr);
 			}
 			Brew brew = BotaniaAPI.instance().getBrewRegistry().get(brewId);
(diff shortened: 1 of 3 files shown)


src/main/java/vazkii/botania/common/crafting/RecipeBrew.java
```diff
@@ -143,7 +144,7 @@ public boolean equals(Object o) {
 		public RecipeBrew read(@Nonnull Identifier id, @Nonnull JsonObject json) {
 			String brewStr = JsonHelper.getString(json, "brew");
 			Identifier brewId = Identifier.tryParse(brewStr);
-			if (brewId == null || !BotaniaAPI.instance().getBrewRegistry().containsId(brewId)) {
+			if (brewId == null || BotaniaAPI.instance().getBrewRegistry().get(brewId) == ModBrews.fallbackBrew) {
```
williewillus, line 147: can you explain this and the one below? The existing check should work.

Yoghurt4C:
```java
@Environment(EnvType.CLIENT)
public abstract boolean containsId(Identifier id);
```

Yoghurt4C: I could replace these with `Registry.BLOCK.get(Registry.BLOCK.getDefaultId());` if you're unsure about the hardcoding

williewillus: Hmm, is there a method that returns null instead of the default value? If so, I'd rather use that and null check.

If there isn't, what you suggest in your comment is fine.

Yoghurt4C:
```suggestion
			if (brewId == null || !BotaniaAPI.instance().getBrewRegistry().getOrEmpty(brewId).isPresent()) {
```

src/main/java/vazkii/botania/common/crafting/StateIngredientHelper.java
```diff
@@ -111,7 +112,7 @@ public static BlockState readBlockState(JsonObject object) {
 		ItemNBTHelper.renameTag(nbt, "properties", "Properties");
 		String name = nbt.getString("Name");
 		Identifier id = Identifier.tryParse(name);
-		if (id == null || !Registry.BLOCK.containsId(id)) {
+		if (id == null || Registry.BLOCK.get(id) == Blocks.AIR) {
```
Yoghurt4C, line 115:
```suggestion
		if (id == null || !Registry.BLOCK.getOrEmpty(id).isPresent()) {
```

PR #3537, opened 2020-11-29
apply enchanting power to pylons

Yoghurt4C:
essentially a dirty fix for https://github.com/williewillus/botania-fabric-issues/issues/6.
I'm not sure if there are any frameworks for this, but this mixin simply runs logic similar to vanilla's own bookshelf check before said checks actually happen. `modifyvariable` is used over several `inject`s because I'm not *entirely* sure if it's possible to capture variables inside an `if` conditional inside a loop and do anything useful with them without breaking something
image
image


Yoghurt4C:
I've been directed to <https://github.com/FabLabsMC/Block-Extensions/blob/master/src/main/java/io/github/fablabsmc/fablabs/api/block/extensions/v1/BlockExtensions.java#L88>, which is meant to "become part of fapi soon", but I wouldn't hold out on that `soon` until 1.17

williewillus:
Adding a dep on that lib for now is fine

PR #3538, opened 2020-11-29
Fix exchange rod punch-to-replace

codewarrior0:
The exchange rod now works correctly in creative mode, survival mode,
singleplayer, and multiplayer. The price? It can no longer be used as
a digging tool.

Fixes williewillus/botania-fabric-issues#26

(Multiplayer testing was accomplished with #3535, so maybe merge that one too)


Hubry:
Uh, is this gonna be needed at all anymore? On the Forge I changed it to use the same thing as the debug stick does. Same caveat, but no mixins or hacky hooks needed.

PR #3539, opened 2020-12-01
added full compat for rei

Yoghurt4C:
the pros:
-mostly faithful (minor subjective improvements where possible)
-it just works

the cons:
-special crafting wrappers aren't reliable for viewing *specific* recipes, but still get the point across
-the orechid categories, while fully functional, aren't populated for the time being

image
image
image
image
image
image
image
image


williewillus:
delete the jei stuff in src_disabled

PR #3541, opened 2020-12-02
data-driven orechids

Yoghurt4C:
this pr:

1. unhardcodes the ore weight "registry" by moving it to `.json`
2. relocates the initial weight processing to serverside datapack re/loading
3. ensures things like REI can access the data clientside
4. forcibly tags vanilla ores under the "common" `#c` namespace with vanilla-ish naming conventions used by other mods
5. due to the prior change, orechids now properly work on the fabric branch without further input; ~~though this pr will most likely clash with https://github.com/Vazkii/Botania/pull/3535~~


ToMe25:
This seems like something similar(except for the removing imc part, for backwards compat) would also make sense for the main forge branch/version.

Yoghurt4C:
I'm having issues setting up a devenv for the master branch, and I'm not sure if the imcsender method removal is of any real concern unless there are mods which piggybacked off of those 2 methods

ToMe25:
> I'm having issues setting up a devenv for the master branch

The only issue i ever had with this, besides standard gradle issues, was the `output.tsrg`, which can be generated with `gradlew createMcpToSrg` not existing, so i don't think i am the right person to help you with that.

> and I'm not sure if the imcsender method removal is of any real concern unless there are mods which piggybacked off of those 2 methods

Idk whether anyone uses them tbh, i just thought "there might be someone using them, so maybe don't remove them".

Yoghurt4C:
the thing is, someone would have to mix into imcsender and either inject into enqueue() or make invokers for those 2 methods because they're private

ToMe25:
On fabric yes, but on forge this is a handler for a forge internal system(that might have been removed since 1.13, not sure).

williewillus:
not super sure what I want to do with this yet, but whatever happens it's probably gonna happen on forge first, just a heads up.

Hubry:
It kinda needs a way to define an ore as a single block and not a tag, to be honest. That would be a potential solution to nether gold appearing.

Yoghurt4C:
> It kinda needs a way to define an ore as a single block and not a tag, to be honest. That would be a potential solution to nether gold appearing.

that'd require some good refactoring of the orechid itself, it only supports tags

Hubry:
I believe this has been made obsolete by my version of this on the Forge branch, just needs a merge into the Fabric branch.

PR #3570, opened 2020-12-23
Add "Botanist Ready" Superflat preset

eutropius225:
This PR adds a superflat preset called "Botanist Ready".

It looks like this in the menu:

and like this in-game:


Hubry:
We talked about this already on discord lol.
make this a mod with this as one of default added presets smh

PR #3573, opened 2020-12-24
Prioritize Vitreous Pickaxe for the Ring of Correction when breaking glass

eutropius225:
This PR ensures that the Vitreous Pickaxe is swapped for by the Ring of Correction when breaking glass. The previous behaviour was that it would always be swapped out.

Here it is after the changes:

https://user-images.githubusercontent.com/11538216/103090080-ef54b500-45e7-11eb-8ea3-67cae6c33ab6.mp4

### DIFF
--- a/src/main/java/vazkii/botania/common/item/equipment/tool/ItemGlassPick.java
+++ b/src/main/java/vazkii/botania/common/item/equipment/tool/ItemGlassPick.java
@@ -75,9 +75,8 @@ public class ItemGlassPick extends ItemManasteelPick {
 	@Override
 	public boolean onBlockStartBreak(ItemStack itemstack, BlockPos pos, PlayerEntity player) {
 		BlockState state = player.world.getBlockState(pos);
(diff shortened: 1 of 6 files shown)


src/main/java/vazkii/botania/common/item/equipment/tool/ItemGlassPick.java
```diff
@@ -98,14 +97,19 @@ public void inventoryTick(ItemStack stack, World world, Entity player, int slot,
...
 	@Override
 	public int getManaPerDamage() {
 		return MANA_PER_DAMAGE;
 	}
 
 	@Override
-	public int getSortingPriority(ItemStack stack) {
-		return 0;
+	public int getSortingPriority(ItemStack stack, BlockState state) {
+		// if you have a Vitreous Pickaxe on you you probably want it to be chosen when breaking glass
```
williewillus, line 111: not needed

williewillus:
looks good besides the comment

PR #3588, opened 2021-01-03
Fix the Bloodthirst potion at long long last

Alwinfy:
slime checks are skipped tho

### DIFF
--- a/src/main/java/vazkii/botania/common/brew/potion/PotionBloodthirst.java
+++ b/src/main/java/vazkii/botania/common/brew/potion/PotionBloodthirst.java
@@ -26,7 +26,6 @@ public class PotionBloodthirst extends Effect {
 		super(EffectType.BENEFICIAL, 0xC30000);
 	}
 
-	// TODO: This runs a check for *every* player, would it be faster to just query players within the spawn AABB?
 	public static boolean overrideSpawn(IServerWorld world, BlockPos pos, EntityClassification entityClass) {
 		if (entityClass == EntityClassification.MONSTER) {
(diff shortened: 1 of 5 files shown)


src/main/java/vazkii/botania/common/brew/potion/PotionBloodthirst.java
```diff
@@ -26,19 +26,19 @@ public PotionBloodthirst() {
 		super(EffectType.BENEFICIAL, 0xC30000);
 	}
 
-	public static void onSpawn(LivingSpawnEvent.CheckSpawn event) {
-		if (event.getResult() != Event.Result.ALLOW && event.getEntityLiving() instanceof IMob) {
-			AxisAlignedBB aabb = new AxisAlignedBB(event.getX() - RANGE, event.getY() - RANGE, event.getZ() - RANGE,
-					event.getX() + RANGE, event.getY() + RANGE, event.getZ() + RANGE);
-			for (PlayerEntity player : event.getWorld().getPlayers()) {
+	// TODO: This runs a check for *every* player, would it be faster to just query players within the spawn AABB?
```
williewillus, line 29: No, because we used to do that and it was actually slower. Touching the chunks to query their entities for players has super high over head, while doing a distance check for at most 100 or so players is much cheaper.

williewillus:
Can you add comments on override spawn to explain why it needs to be hooked from two places?

Hubry:
Re: slime checks, I'm getting slime spawns outside of slime chunks when testing on 1.12.

PR #3597, opened 2021-01-07
API overhaul - Part 1: api/item package

### LINKED ISSUE #3594: Interface.Optional being gone means many parts of Botania's API are non-viable
Implementing some interfaces for compatibility is now impossible for soft-dependencies because it is no longer possible (to my knowledge) to specify that an interface should be implemented conditionally.

What I'd like to suggest is to move interface APIs to [Registry-like types](https://github.com/Da-Technomancer/Essentials/pull/49).
It may also be possible to make some APIs into capabilities, but such a thing is only viable on tile entities as there is to my knowledge no way to attach a capability to a block. 

Advisory on this problem is very welcome.

DaedalusGame:
Very partially resolves #3594 (See also for why this is PR exists).

The idea:
- Non-impeding, made to support as many things as possible.
- Non-destructive, mods using the old API should continue working unimpeded.
- Self-similar, made in the image of existing API classes.
- Portable, should still be compatible with fabric and forge

The concept:
- Each interface handler is stashed in a registry
- Calling `has(item)` on a registry checks if `item` implements the interface or has a registered handler in the registry.
- Calling `get(item)` on a registry retrieves the handler from the registry or `item` itself, preferring the implemented interface over the one stored in the registry.
- Registries can either be gotten directly from an instance of ItemAPI or from the handler interface (ex: `ISortableTool.registry()`)

**Why not capabilities?**
Attaching item capabilities to vanilla items or other mod's items is often tedious, impeding compatibility addon development. I'm not opposed to turning the item registries into item capabilities, but blocks must remain in this system as to my knowledge it's impossible to attach capabilities to blocks without tile entities.

**Non-functional interfaces should probably be tags**
Please advise on this point, it seems like it would be possible to turn ICosmeticBauble into a tag.

**Explicit empty checks**
Please advise on this point. Item registries could easily have get overloads added for ItemStacks that check for stack emptyness before querying the hashmap.

**Code example**
Old code:
```
if (stackItem instanceof IFlowerPlaceable) {
    success = ((IFlowerPlaceable) stackItem).tryPlace(this, ctx);
}
```
New code:
```
IFlowerPlaceable placeable = IFlowerPlaceable.registry().get();
if(placeable != null) {
    success = placeable.tryPlace(this, ctx);
}
```


src/main/java/vazkii/botania/common/Botania.java
```diff
@@ -197,6 +206,27 @@ private void commonSetup(FMLCommonSetupEvent event) {
...
+					}
+				}
+			}
+
+			@Override
+			public ResourceLocation getOverlayResource(IAvatarTile tile, ItemStack stack) {
+				return new ResourceLocation(LibResources.MODEL_AVATAR_DIRT);
+			}
+		});
+
```
Hubry, line 229: haha yes

DaedalusGame: oops

Vazkii:
begone bord

PR #3600, opened 2021-01-09
Fix fabric stridersash jump heights

Alwinfy:


williewillus:
explain? these constants should not be different from forge.

Alwinfy:
no idea how they're different but they work experimentally the same as forge

williewillus:
concern

Hubry:
Better idea: don't call this 3 times? It adds motion 3 times because the equip check uses an `instanceof`
https://github.com/Vazkii/Botania/blob/4c39b77de3e430ce90920202851fc0b39faf79c2/src/main/java/vazkii/botania/mixin/MixinLivingEntity.java#L89-L91

Alwinfy:
Yeah okay that ought to do it :wacko:

PR #3618, opened 2021-01-20
Fix logo (again)

Fogrits:
I made the logo a transparent square to fix the "logo must be square" exception on Fabric + used the website logo (I think it's better than the default logo)

PS: The image size and file size are larger and than intended because I had to upscale the logo to fit Fabric's Mod Menu screenimage
image


Fogrits:
dunno why the build test has failed, even though I just changed a image file

williewillus:
Not your fault, I accidentally committed something broken earlier this week and havent fixed it yet. This looks good.

PR #3630, opened 2021-01-30
Fix some Optional<> usages to be actually idiomatic

Alwinfy:


src/main/java/vazkii/botania/common/block/tile/TileRuneAltar.java
```diff
@@ -203,12 +203,9 @@ private void updateRecipe() {
 			if (currentRecipe != null) {
 				this.manaToGet = currentRecipe.getManaUsage();
 			} else {
-				Optional<IRuneAltarRecipe> maybeRecipe = world.getRecipeManager().getRecipe(ModRecipeTypes.RUNE_TYPE, getItemHandler(), world);
-				if (maybeRecipe.isPresent()) {
-					this.manaToGet = maybeRecipe.get().getManaUsage();
-					break getMana;
```
williewillus, line 209: remove the label if it's not used anymore

src/main/java/vazkii/botania/common/crafting/RecipeBrew.java
```diff
@@ -144,10 +144,7 @@ public boolean equals(Object o) {
 		public RecipeBrew read(@Nonnull ResourceLocation id, @Nonnull JsonObject json) {
 			String brewStr = JSONUtils.getString(json, "brew");
 			ResourceLocation brewId = ResourceLocation.tryCreate(brewStr);
-			if (brewId == null || !BotaniaAPI.instance().getBrewRegistry().getOptional(brewId).isPresent()) {
-				throw new JsonParseException("Unknown brew " + brewStr);
-			}
-			Brew brew = BotaniaAPI.instance().getBrewRegistry().getOrDefault(brewId);
+			Brew brew = BotaniaAPI.instance().getBrewRegistry().getOptional(brewId).orElseThrow(() -> new JsonParseException("Unknown brew " + brewStr));
```
williewillus, line 147: getOptional handles nulls?

src/main/java/vazkii/botania/common/item/ItemManaMirror.java
```diff
@@ -137,16 +135,10 @@ private static GlobalPos getBoundPos(ItemStack stack) {
...
-		}
-
-		BlockPos coords = pos.get().getPos();
-		if (coords.getY() == -1) {
-			return null;
-		}
-		return pos.get();
+		return GlobalPos.CODEC.parse(NBTDynamicOps.INSTANCE, ItemNBTHelper.get(stack, TAG_POS))
+				.result()
+				.filter(x -> x.getPos().getY() != -1)
```
williewillus, line 140: you inverted the condition

also name the lambda param `pos` since `x` is confusing

williewillus (changes requested):

src/main/java/vazkii/botania/common/crafting/RecipeBrew.java
Alwinfy: Signature in code is `public Optional<T> getOptional(@Nullable ResourceLocation id);`

src/main/java/vazkii/botania/common/item/ItemManaMirror.java
Alwinfy: filter returns a null optional when the condition is false. the condition is false when getY() == -1

PR #3641, opened 2021-02-05
Add lens splitting recipe

eutropius225:
This PR implements a recipe that allows splitting composite lenses into their component lenses.
example
The slime ball is lost, but the two lenses are recovered.


williewillus:
thank

PR #3654, opened 2021-02-10
The T in LGBT stands for Tater

SuicidalSteve:
-Add LGBTaters gotten by renaming tiny potato with certain names put between colons, originally based on the :pridetater: Discord emote.
Options are:
-Trans tater
-Nonbinary tater
-Gay tater
-Bi tater
-Pan tater
-Lesbian tater
-Genderfluid tater
-Asexual tater
-Aromantic tater

PR #3655, opened 2021-02-11
Arbitrary tater models

eutropius225:
This PR adds support for any model to be used for a tiny potato. Models of the form `botania:tiny_potato/<name>` will automatically be loaded, and will appear when a tiny potato's name, (lowercased and with `[ -]` replaced by `_`) matches `<name>`.

For example, it allows for the `pluraltater|manytater|systemtater|systater` to be added without code changes.

This PR reimplements the existing `kyle hyde` and `aureylion` taters, as well as all those added in #3654, using the new system, and makes the halloween tater available at any time as `halloween|spooky|spookytater`.

taters after
They looked like this before
taters before


src/main/resources/assets/botania/models/block/shapes/tiny_potato.json
```diff
@@ -0,0 +1,18 @@
+{
+  "parent": "minecraft:block/block",
+  "elements": [
```
williewillus, line 3: why not just set "particle": "#all" here so you don't have to worry about particle in any of the code or generated models above?

williewillus:
also some of the taters in your screenshot don't match, on the front row.

eutropius225:
> also some of the taters in your screenshot don't match, on the front row.

Third from the left was erroneously named `egg` instead of `eggtater`, the inconsistency has been fixed.
Grumm on the far right lost its name, that's been fixed too.
Botater's sides were on the wrong way around, that's been fixed.
(Far left is `spooky`, which is new)
ss

src/main/resources/assets/botania/models/block/shapes/tiny_potato.json
eutropius225: It's only really relevant for the `default` model. In any case, all of the models now inherit from `block/cube_all` which does have `particle` bound to `#all`.

williewillus:
pog

PR #3664, opened 2021-02-16
Fix tiny potato models not working with resource packs

eutropius225:
Somehow there was an extra `/` appearing in model names when iterating over the resource locations. This PR fixes that.

???
what
---
twontater


eutropius225:
Note that Forge doesn't re-fire the model registry event every resources are reloaded, so to see the effects the game has to be restarted. This is not true on Fabric.

PR #3665, opened 2021-02-17
Use a trie to resolve tiny potato shaders

eutro:
This PR likely offers negligible performance benefit, if any, because of the relatively high constant factor in looking up the character for each node: [incredibly detailed benchmarking](https://gist.github.com/eutropius225/e5d7688c81e6512671fe54c1f02b04b0) has shown that the trie-based lookups are usually a little slower than the hardcoded lookups.

Nevertheless, this system is more extensible (please) and has a much better asymptotic time complexity. Additionally, planting tries (maybe it was trees?) is good for the environment.


williewillus:
JIT-compiled String.equals/startsWith is extremely fast, at least more than ping-ponging around your cache trying to hit trie nodes. I don't see the point of this for 5 elements.

PR #3685, opened 2021-03-14
Fix issues with warp lenses on fabric

BenCat07:
This fixes two major issues with warp lenses on fabric, them destroying the force relays with bore, 

The below was intentional and removed from the pr. makes me sad though.

~~and the bursts no longer carrying any many after being warped.~~

Note that i auto formatted, so excuse me for formatting disrepancies, tell me how to fix them if desired.

### DIFF
--- a/src/main/java/vazkii/botania/common/item/lens/LensMine.java
+++ b/src/main/java/vazkii/botania/common/item/lens/LensMine.java
@@ -22,6 +22,8 @@ import net.minecraft.world.World;
 
 import vazkii.botania.api.internal.IManaBurst;
 import vazkii.botania.api.mana.IManaBlock;
+import vazkii.botania.common.Botania;
+import vazkii.botania.common.block.BlockPistonRelay;
 import vazkii.botania.common.block.ModBlocks;
 import vazkii.botania.common.core.handler.ConfigHandler;
 import vazkii.botania.common.item.ModItems;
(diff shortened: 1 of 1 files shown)


src/main/java/vazkii/botania/common/item/lens/LensMine.java
```diff
@@ -58,18 +61,17 @@ public boolean collideBurst(IManaBurst burst, HitResult rtr, boolean isManaBlock
...
+		if (!source.equals(collidePos) && !(tile instanceof IManaBlock) && neededHarvestLevel <= harvestLevel
+				&& hardness != -1 && hardness < 50F && (burst.isFake() || mana >= 24)) {
 			if (!burst.hasAlreadyCollidedAt(collidePos)) {
 				if (!burst.isFake()) {
 					List<ItemStack> items = Block.getDroppedStacks(state, (ServerWorld) world, collidePos, tile);
 
-					world.removeBlock(collidePos, false);
-					if (ConfigHandler.COMMON.blockBreakParticles.getValue()) {
-						world.syncWorldEvent(2001, collidePos, Block.getRawIdFromState(state));
+					if (!(world.getBlockState(collidePos).getBlock() instanceof BlockPistonRelay)) {
```
BenCat07, line 70: this is the actual change here

BenCat07:
Was notified that i may not have fixed the actual issue, will look into more of it when i have time. Draft in the meantime.

Hubry:
Remember to run Spotless next time because it failed your build on an unused import.

williewillus:
can you check the forge version? why does the existing code work there and not here? That may indicate a deeper underlying bug. Papering it over at the surface level risks hiding that underlying issue.

Also, please run spotlessJavaApply on your branch

BenCat07:
Thank you, will do the latter, checking the former later like i said

BenCat07:
Turns out, combining the lenses in the wrong order was the cause, and it does the same thing on forge. i will see it as "intended" behaviour and close this pull request.

PR #3688, opened 2021-03-18
JSONify orechid, with single block outputs too

### LINKED ISSUE #3667: CraftTweaker / Orechid support
I would like to tweak the orechid to:
* Disable a nether ore it is spawning in the overworld
* Add a nether ore to the ignem (although haven't tested whether the aforementioned block is there or not already)
* Disable a duplicate ore that from a mod that I've decided is not the canonical block

CraftTweaker support for the orechid would enable this.

Hubry:
Alternative to #3541, which targets the fabric branch and doesn't support anything but tags.
This approach also bakes the data into a form where unnecessary data (like empty tags, and blocks ignored because of the priority config option) is discarded. Some of this code is pretty ugly though and I had too many distractions around me to focus.

Will close #3667 once I finish CT support I started working on.


src/main/java/vazkii/botania/common/crafting/StateIngredientBlocks.java
```diff
@@ -37,6 +42,23 @@ public boolean test(BlockState state) {
 		return blocks.contains(state.getBlock());
 	}
 
+	@Override
+	public BlockState pick(Random random) {
+		if (blocks instanceof List) {
+			return ((List<Block>) blocks).get(random.nextInt(blocks.size())).getDefaultState();
+		}
+
+		// Slow path, iterating over the set
```
williewillus, line 51: could just store both the set and list forms so both `test` and `pick` are fast, at the cost of some ram. That would let you get rid of the extra ctor too.

src/main/java/vazkii/botania/common/crafting/StateIngredientHelper.java
```diff
@@ -62,16 +78,42 @@ public static StateIngredient deserialize(JsonObject object) {
...
+			}
+			return new StateIngredientBlocks((Collection<Block>) blocks);
 		default:
 			throw new JsonParseException("Unknown type!");
 		}
 	}
 
+	/**
+	*
+	*/
```
williewillus, line 96: ?

src/main/java/vazkii/botania/common/crafting/StateIngredientHelper.java
```diff
@@ -91,6 +133,42 @@ public static StateIngredient read(PacketBuffer buffer) {
...
+			List<Block> blocks = tag.getAllElements();
+			if (blocks.isEmpty()) {
+				return null;
+			}
+			return getFiltered(ingredient, function, blocks);
+
+		} else if (ingredient instanceof StateIngredientBlocks) {
+			Collection<Block> blocks = ((StateIngredientBlocks) ingredient).getBlocks();
+			if (!(blocks instanceof List)) {
+				return ingredient;
```
williewillus, line 156: why?

src/main/java/vazkii/botania/common/crafting/StateIngredientHelper.java
```diff
@@ -91,6 +133,42 @@ public static StateIngredient read(PacketBuffer buffer) {
 		}
 	}
 
+	/**
+	 * Resolves tag ingredients, returning null if their tag doesn't exist, then returns a new ingredient
+	 * if the replacement function returns a replacement, used to filter ores deprioritized with the config.
+	 */
+	@Nullable
+	public static StateIngredient resolveAndFilter(StateIngredient ingredient, @Nonnull UnaryOperator<List<Block>> function) {
```
williewillus, line 141: maybe make this an instance method on StateIngredient instead of a static one with if-elses?

williewillus (commented):
looks mostly fine, but from what I can tell you have a weird dichotomy in the stateingredients between sets and lists (some situations purposefully pass a set, others a list), and this logic seems implicit/undocumented. Can that be eliminated (just have both)? Otherwise, it needs to be clearly documented somewhere.

Hubry:
Well, I think I wanted to avoid the situation where a big list gets passed (either explicitly, or through filtering a tag down to a mod) and we go through the entire collection with the iterator because sets are not indexed. That said this probably is negligible compared to filtering the entire tag list down every single time you create ore with the orechid...

I'll swap it to store both I guess, after all every tag in the game has both and it's fine...

src/main/java/vazkii/botania/common/crafting/StateIngredientBlocks.java
Hubry: Looks like ImmutableSet has a nice method that converts it to a list (cached on the set), used that.

src/main/java/vazkii/botania/common/crafting/StateIngredientHelper.java
Hubry: Oops

src/main/java/vazkii/botania/common/crafting/StateIngredientHelper.java
Hubry: brain must have been tired at that point

Hubry:
Some of the things you pointed out seem too obvious in hindsight, glad that I actually went with making a PR lol

Hubry:
Pushed manually.

PR #3698, opened 2021-03-31
Remove the and slimeblock requirement for Unethical TNT

foxwhite25:
Slime block is not necessary to create a duped tnt , here is a image of that
图片
this setup is designed to not even use sticky piston , so literally doesnt need any slime or slimeblock 
If anyone want to check it out I will leave a world download


foxwhite25:
Just a notice , you can do whatever you want

williewillus:
What was the point of posting it here, then?

The point of the check is to block most of the lazy people copying and pasting solutions from reddit or youtube. The check cannot account for all the shitty ways in which pistons are broken, and I wish Mojang would just add movable BE's and remove this stupid exploit from the game.

PR #3700, opened 2021-04-02
Rod of Elytraboosting

00-Steven:
Idea by LeoBeliik
- Make the Rod of the Skies give an elytra boost which I believe is *slightly* worse than the lowest tier firework.
- Make Livingwood Avatars holding said rod give people flying by them a boost. This boost is a bit stronger than the handheld rod boost. No this doesn't stack infinitely, no human railguns for you.

So yeah, this is pr the for the videos I kept sending on Vazcord.
I believe the intensity of the boosts aren't all too unbalanced- the avatars are stationary, require some infrastructure, and its boosts aren't continuous like the rod's, so they can be a bit bigger.
Similarly, the normal rod boost is about on par with a regular lowest tier firework, if not slightly worse, so I believe that's at least decent.

### DIFF
--- a/src/main/java/vazkii/botania/common/item/rod/ItemTornadoRod.java
+++ b/src/main/java/vazkii/botania/common/item/rod/ItemTornadoRod.java
@@ -83,7 +83,10 @@ public class ItemTornadoRod extends Item implements IManaUsingItem, IAvatarWield
 					player.playSound(ModSounds.airRod, 0.1F, 0.25F);
 					for (int i = 0; i < 5; i++) {
 						WispParticleData data = WispParticleData.wisp(0.35F + (float) Math.random() * 0.1F, 0.25F, 0.25F, 0.25F);
-						world.addParticle(data, player.getPosX(), player.getPosY(), player.getPosZ(), 0.2F * (float) (Math.random() - 0.5), -0.01F * (float) Math.random(), 0.2F * (float) (Math.random() - 0.5));
+						world.addParticle(data, player.getPosX(), player.getPosY(), player.getPosZ(),
+								0.2F * (float) (Math.random() - 0.5),
+								-0.01F * (float) Math.random(),
+								0.2F * (float) (Math.random() - 0.5));
 					}
 				}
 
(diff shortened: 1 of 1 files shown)


src/main/java/vazkii/botania/common/item/rod/ItemTornadoRod.java
```diff
@@ -146,20 +151,45 @@ public void onAvatarUpdate(IAvatarTile tile, ItemStack stack) {
...
+							for (int j = 0; j < 5; j++) {
+								WispParticleData data = WispParticleData.wisp(0.35F + (float) Math.random() * 0.1F, 0.25F, 0.25F, 0.25F);
+								world.addParticle(data, p.getPosX() + lookDir.getX() * i
+										, p.getPosY() + lookDir.getY() * i
+										, p.getPosZ() + lookDir.getZ() * i
+										, 0.2F * (float) (Math.random() - 0.5) * (Math.abs(lookDir.getY()) + Math.abs(lookDir.getZ())) + -0.01F * (float) Math.random() * lookDir.getX()
+										, 0.2F * (float) (Math.random() - 0.5) * (Math.abs(lookDir.getX()) + Math.abs(lookDir.getZ())) + -0.01F * (float) Math.random() * lookDir.getY()					//FIX YOUR SHIT, MAKE PARTICLES GO THE RIGHT WAY
+										, 0.2F * (float) (Math.random() - 0.5) * (Math.abs(lookDir.getY()) + Math.abs(lookDir.getX())) + -0.01F * (float) Math.random() * lookDir.getZ());
+							}
+						}
```
williewillus, line 170: pull the elytra part out into a separate method for cleanliness (you can pull the non elytra side out too if you want, but you don't have to)

SuicidalSteve: Oh whoops, I see I left in some stupid comments-
Yeah I'll look into doing that when I get the time

SuicidalSteve: So did some cleanup, and changed the way the boosts are implemented to allow for stacked boosts to be a lot more smooth. Did this by increasing the hardcap, adding a short cooldown, and applying diminishing returns based on current speed.
-Though, actual values might need some tweaking, but that's something I imagine would require more actual survival play.

Hubry:
Please run the `spotlessApply` task so your formatting doesn't break the build.

SuicidalSteve:
This should fix that- probably, hopefully, maybe

src/main/java/vazkii/botania/common/item/rod/ItemTornadoRod.java
```diff
@@ -142,27 +152,101 @@ public void onAvatarUpdate(IAvatarTile tile, ItemStack stack) {
...
+	private CompoundNBT getAvatarCooldowns(TileEntity te) {
+		if (te.getTileData().contains(TAG_COOLDOWNS)) {
+			return ((CompoundNBT) te.getTileData().get(TAG_COOLDOWNS));
+		} else {
+			return new CompoundNBT();
+		}
+	}
+
+	private void setAvatarCooldowns(TileEntity te, CompoundNBT cooldownTag) {
+		te.getTileData().put(TAG_COOLDOWNS, cooldownTag);
```
williewillus, line 239: sorry for getting back to this late. can you not use getTileData and just call methods on TileAvatar directly? getTileData is not portable to Fabric.

williewillus:
Manually applied

PR #3723, opened 2021-04-30
Rearrange Categories of Lexica Botania

DasGerippe:
Rearranged the categories of the Lexica Botania since they were without logical order (except for Basics being first and Challanges being last).

The order I had in mind is:
- Basics
- Mana Manipulation
- Generating Flora
- Functional Flora
- Tools (Mystical Items)
- Devices (Natural Apparatus)
- Baubles
- Misc
- Alfhomancy
- Ender
- Challanges

Feel free to further adjust the order.

### DIFF
--- a/src/main/resources/data/botania/patchouli_books/lexicon/en_us/categories/alfhomancy.json
+++ b/src/main/resources/data/botania/patchouli_books/lexicon/en_us/categories/alfhomancy.json
@@ -2,5 +2,5 @@
   "name": "botania.category.alfhomancy",
   "description": "botania.category.alfhomancy",
   "icon": "botania:flight_tiara",
-  "sortnum": 5
+  "sortnum": 8
 }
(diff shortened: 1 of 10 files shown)


Hubry:
Uh, did I break something in Patchouli that made them go unsorted? Oops

Hubry (changes requested):
The original order from 1.12 and before was:

- Basics
- Mana
- Generating Flowers
- Functional Flowers
- Devices
- Tools
- Baubles
- Ender
- Alfhomancy
- Misc

Can you tweak it to exactly reproduce this? I'll merge afterwards.

DasGerippe:
Sure, done. Something else just popped into my mind: Since Challanges probably always should be the last category, do we set it's sortNumber to something very high, so categories of addons wont be after challanges? Or does the addon dev have to worry about this?

Hubry:
Not a bad idea, actually.

Hubry (approved):

PR #3735, opened 2021-05-24
Add API method to register IHornHarvestable instances

InfinityRaider:
This removes the need to explicitly implement the interface in block classes.

### DIFF
--- a/src/main/java/vazkii/botania/api/BotaniaAPI.java
+++ b/src/main/java/vazkii/botania/api/BotaniaAPI.java
@@ -152,8 +152,8 @@ public interface BotaniaAPI {
 	/**
 	 * Make Botania recognize a Block as IHornHarvestable without explicitly implementing the interface
 	 *
-	 * @param blockId		The block ID
-	 * @param harvestable	The harvestable
+	 * @param blockId     The block ID
+	 * @param harvestable The harvestable
 	 */
 	default void registerHornHarvestableBlock(ResourceLocation blockId, IHornHarvestable harvestable) {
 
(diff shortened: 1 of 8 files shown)


williewillus:
Awesome.
Could you migrate Botania's own blocks to use this way? Eventually we'd probably want to remove support for the old way entirely. (probably in 1.17)

InfinityRaider:
When I have some time this evening I'll get on it.
Do you have any preference on where to register them? I would propose to do it where you are registering the blocks.

williewillus:
That's fine, yes

InfinityRaider:
All done.

Sidequestions: 
 - Is the Drum of the Wild no longer a thing in 1.16, as I couldn't find any other usages of IHornHarvestable outside of the ItemHorn class?
 - Is there no mixin for vanilla crops / leaves that implements IHornHarvestable in them, or have I overlooked it?

Hubry:
* The drum calls the horn code here: https://github.com/Vazkii/Botania/blob/80b2c1e4711c0ecdda9f1648a99682e46f89d6a8/src/main/java/vazkii/botania/common/block/mana/BlockForestDrum.java#L97-L100
* If the block isn't horn harvestable it does different logic: https://github.com/Vazkii/Botania/blob/80b2c1e4711c0ecdda9f1648a99682e46f89d6a8/src/main/java/vazkii/botania/common/item/ItemHorn.java#L88-L94

InfinityRaider:
I see, thanks for clarifying that.

williewillus:
thanks

InfinityRaider:
Cool, thanks for the merge!
Could I get a ping (here or on discord) once this goes live, so i can get cracking on my end?

Hubry:
You can use a snapshot dependency for now, but builds using that won't be usable by the general public until we release. I have no idea when would that happen though - we don't really have any substantial fixes waiting for release, maybe it is a good idea to just push it out.

InfinityRaider:
Where would I find this snapshot release, I do not see it listed on the maven index.

Issue #150, opened 2014-07-11
Mapmaker Features


Vazkii:
From TheMadman: http://pastebin.com/DNKMuKdS

PR #3737, opened 2021-05-27
Removing unnecessary NBT tags

Keralop:


Hubry:
What is the reason for this? Leaving the description empty doesn't really help.

The accept/spare tags are there for mapmaker reasons, similarly to all the mapmaker overrides in spreaders, see #150. I have no idea if anyone actually used those though.

williewillus:
Please provide a summary for your diffs.

PR #3750, opened 2021-06-13
Added a missing doc param to the rune altar

jaredlll08:
Without it we can't generate an example for it, so on the docs it looks like:
```zenscript
RuneAltar.addRecipe(name as string, output as IItemStack, mana as int, inputs as IIngredient[]) as void
```

instead of
```
<recipetype:botania:runic_altar>.addRecipe("petal_apothecary_test", <item:minecraft:diamond>, 5000, <item:botania:rune_>, <item:botania:orange_petal>, <item:botania:red_petal>);
```


Hubry:
oh, nice

PR #3781, opened 2021-07-21
Make the request and requestExact methods in ManaItemHandlerImpl drain mana from multiple IManaItems at once.

noeppi-noeppi:
Also fixes that `requestManaExact` always left at least one mana in the mana item it was requesting from, so the request failed when requesting 500000 mana from a player with a completely full band of mana in their inventory.

See https://github.com/noeppi-noeppi/MythicBotany/issues/66


Hubry:
@noeppi-noeppi Looks like this causes *adding* mana instead of removing it.

PR #3789, opened 2021-08-02
Fix flower binding range not being centered at spreader/pool.

### LINKED ISSUE #3775: Binding range sphere not centered at mana pool / spreader
# Version Information

Forge version:  1.16.5-forge-36.1.32
Botania version:  Botania-1.16.5-418

# Further Information

This bug occurs for generating and functional flowers getting bound to a spreader / pool respectivly. I will only describe the behaviour for functional flowers.

The range for a mana pool in wich flowers can be bound to it is hard-coded as 10 blocks. However, placing a flower 10 blocks away from a mana pool only bounds it to the pool in positive x, y or z directions. In negative x, y, or z directions, the range is only 9 blocks. The resulting sphere in wich flowers can be bound to a pool is not centered at the pool. The same behaviour can be seen when trying to manually bind flowers with a wand of the forest.

Funnily though, when using a Floral Obedience Stick, the sphere in wich flowers will be bound to the pool is shifted in the **opposite** direction (shifted towards negative x, y, and z coordinates).

In the image you can see the green circle in wich flowers wil be automatically bound to the pool / can be manually bound to it, as well as the blue circle in wich flowers will be bound to the pool when using the Floral Obedience Stick.

As i said, the same applies for generating flowers and mana spreaders, just with a hard-coded range of 6 blocks.

DasGerippe:
Fixes #3775 

So basically the previously used method
BlockPos#distanceSq(BlockPos)
would center the coordinates (add 0.5) of the variable the method has been called on, but not of the passed parameter. That lead to the resulting binding range being shifted.

Calling
BlockPos#distanceSq(x, y, z, boolean center)
instead, allows to specify whether the coordinates should be centered or not.

Adding 0.5 to the range before squaring makes the resulting binding range a nice and clean circle, instead of a circle with dots on the flat sides (dunno how to describe it better).

### DIFF
--- a/src/main/java/vazkii/botania/api/subtile/TileEntityFunctionalFlower.java
+++ b/src/main/java/vazkii/botania/api/subtile/TileEntityFunctionalFlower.java
@@ -214,11 +214,11 @@ public class TileEntityFunctionalFlower extends TileEntitySpecialFlower {
 
 	@Override
 	public boolean bindTo(PlayerEntity player, ItemStack wand, BlockPos pos, Direction side) {
-		int range = 10;
-		range *= range;
+		double sqRange = (LINK_RANGE + 0.5) * (LINK_RANGE + 0.5);
 
-		double dist = pos.distanceSq(getPos());
-		if (range >= dist) {
+		BlockPos flower = getPos();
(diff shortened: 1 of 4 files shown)


Hubry:
Fixed in #3831

PR #3793, opened 2021-08-07
Use Patchouli's newer first-class l10n for page entries

Alwinfy:
This trips MinecraftForge/MinecraftForge#7396 on pages like the Manastar, unfortunately. At least it decouples Botania from "i18n": "true"?


Hubry:
There is an awaiting Forge PR for 1.16 fixing this now, and it is fixed on 1.17, btw.

artemisSystem:
Is this the PR you're referring to? https://github.com/MinecraftForge/MinecraftForge/pull/8050

Since it's now merged, does that mean this can be closed (or should it be merged?)

PR #3804, opened 2021-08-21
Add top margin to the carousel

Yash-Singh1:


williewillus:
Hi, thanks for the patch. Can you show a screenshot of the change?

Yash-Singh1:
It increases the top margin for the carousel. I thought there should be some spacing between the navigation bar and the carousel.

Before:

Screenshot from 2021-08-21 13-00-46

After:

Screenshot from 2021-08-21 13-00-55

williewillus:
Great, thanks.

PR #3809, opened 2021-09-01
Make the corporea index look better (+ some other stuff)

quat1024:
* Fix the math in `RenderTileCorporeaIndex` to add `partialTicks` to the global tick count *before* multiplying it by 2, this makes it not jitter so much
* Pass `ClientTickHandler.partialTicks` through in `TEISR`, everything that uses it now animates smoother. The Botanical Brewery looks a lot nicer now
* Allow the corporea index TEISR to rotate as well. Why not?
* Add a custom render bounding box to the corporea index, so the pink particles don't get prematurely culled when looking away. (I think this method is a forge extension unfortunately)
* Use more fitting particle textures for the corporea index (and corporea crystal cube) since it was bugging me

Also I wanted to fix the AO acting strange on corporea indices (try placing 9 of them in a square, it makes the floor turn black), but I couldn't figure out a better way than just giving them a non-full collision box. I thought the old collision box was kind of big anyways lol

### DIFF
--- a/src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaIndex.java
+++ b/src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaIndex.java
@@ -307,14 +307,14 @@ public class TileCorporeaIndex extends TileCorporeaBase implements ICorporeaRequ
 		super.onChunkUnloaded();
 		removeIndex(this);
 	}
-	
+
 	@OnlyIn(Dist.CLIENT)
 	@Override
 	public AxisAlignedBB getRenderBoundingBox() {
 		//The tile entity renderer can draw pink stars fairly far away from the index itself, this helps it not get culled too early.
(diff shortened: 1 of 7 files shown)


quat1024 (commented):
mfw i confused the names of corporea index and corporea crystal cube again

src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaIndex.java
```diff
@@ -308,6 +308,13 @@ public void onChunkUnloaded() {
 		removeIndex(this);
 	}
 
+	@OnlyIn(Dist.CLIENT)
```
Hubry, line 311: I don't think this OnlyIn is necessary but whatever

Hubry (approved):
lgtm

williewillus:
thank

PR #3812, opened 2021-09-02
Port banner stuff to banner++

MoriyaShiine:
This removes the crafting recipe way of making banners as it isn't in vanilla any more and no objections were made, if it is required then it can be added back


Hubry:
* The localization and texture location changes ideally should be reverted, see the PR willie did: kvverti/banner-plus-plus#8
  * Keep in mind we still have a potential Forge port to think about later, and having to sync these two changes will be annoying.
* I am really starting to think if the best option isn't really just making new banner pattern items instead of trying to make existing items work.

williewillus:
Yeah, let's do what Hubry suggested. It'll match vanilla better too. Basically add new banner pattern items and recipes to make them (paper + the old item), and use those instead.

williewillus:
Only the tater and terrasteel ones need a pattern item, the rest (fish + tools) just make available without a required pattern item.

MoriyaShiine:
Gonna redo this once bpp gets a new release

PR #3817, opened 2021-09-12
Fixed the issue with the built jar

CammiePone:
This fixes the mixin responsible for crashing the game at startup with built jars. This doesn't happen in dev environments, only with a built jar. I tested to ensure the behaviour is still correct.


Hubry:
this is what I get for being "fancy" and all that with mixin

PR #3826, opened 2021-10-01
Redo flower-binding logic, remove flower.forceCheck

quat1024:
~~**Please don't merge this yet it, it's scary internal stuff and needs some more testing**~~ Yolo

This PR cleans up flower binding logic a little bit.

* `cachedCollectorCoordinates` has been renamed to just `collectorCoordinates`, and it will not be set to `null` upon the flower being unable to find a collector.
* Same for functional flowers and pools.
* The `flower.forceCheck` option has been removed, since it seems to be leftover performance snake-oil at this point: disabling `forceCheck` caused flowers to only auto-bind to the nearest pool/collector if the amount of pools/collectors in the world was different than last time it tried, but all that logic was wrapped up in a `ticksExisted == 1` block, so it'd only ever try once anyways. This PR essentially locks the option to "on".
* Direct caching of tile entities has been removed from TileEntityGeneratingFlower and TileEntityFunctionalFlower, and a call to getTileEntity is made to look up the binding each time it's used instead. This makes the `xxxCoordinates` fields the authoritative source for the flower's binding. This is - technically - worse for performance, but I think it makes better code.
* I also added something similar to PR #3789 since this PR changes the flower logic so much that that PR isn't mergable anymore.
* The Endoflame consumes fuel when not bound to a spreader now. It's the only time a flower attempted to have different behavior when bound vs. unbound (i.e. the only reason the field was `protected` instead of `private`), and dates back to 2014 - I guess it's just cruft.

### DIFF
--- a/src/main/java/vazkii/botania/api/BotaniaAPI.java
+++ b/src/main/java/vazkii/botania/api/BotaniaAPI.java
@@ -296,8 +296,9 @@ public interface BotaniaAPI {
 	default void sparkleFX(World world, double x, double y, double z, float r, float g, float b, float size, int m) {}
 
 	/**
-	 * See config value "flower.forceCheck" for more information
+	 * Deprecated - the flower.forceCheck option has been removed.
 	 */
+	@Deprecated
 	default boolean shouldForceCheck() {
 		return false;
 	}
--- a/src/main/java/vazkii/botania/api/BotaniaAPI.java
+++ b/src/main/java/vazkii/botania/api/BotaniaAPI.java
@@ -300,7 +300,7 @@ public interface BotaniaAPI {
 	 */
 	@Deprecated
 	default boolean shouldForceCheck() {
-		return false;
+		return true;
 	}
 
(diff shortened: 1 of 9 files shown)


quat1024:
Things to explore:

* Is it possible to completely remove the `linkedCollector`/`linkedPool` fields?
  * These cache a direct reference to a TileEntity, which feels idk, kinda weird.
  * Doing this would make the `blablaCoordinates` fields truly authoritative sources on the binding.
  * You can also remove a lot of fragile cache-busting code that verifies that the tile is still real and in the world, etc
  * But now you have to work out issues like "the binding is valid, except the chunk is unloaded right now, so it's kinda invalid"

Also I initially made this PR because i was trying to throw out `IManaNetwork`; flower binding is one of the few things that use it.

quat1024:
>     * Is it possible to completely remove the `linkedCollector`/`linkedPool` fields?

Yes it is lol

quat1024:
Closing in favor of doing this on the 1.17-fabric branch because it's a large change.

PR #3831, opened 2021-10-03
(1.17.x-fabric) Rewrite and merge generating/functional flower binding logic, remove forceCheck, fix Wand/Obedience binding being off-center

quat1024:
Basically its #3826 but more fabric. Wheres the button to mark it as draft. Sorry for the huge PR - would really like to get some eyes on this before merging

- Flower bindings are just one BlockPos field and they don't cache their bound tile entity anymore.
- Remove `flower.forceCheck` option since it seemed to be mostly unused and did not contribute much performance benefit anymore.
- Fixes flower binding being off-center a little bit by implementing a more accurate `distSqr` method in `MathHelper`.
- Merges binding logic from `TileEntityGeneratingFlower` and `TileEntityFunctionalFlower` into a new common superclass, `TileEntityBindableSpecialFlower<T>`, where `T` is the class of tile entities the flower can be bound to. This isn't suuuuper pretty, but it's better than copypasting half the methods across the classes.
  - Would like feedback on whether using a generic here is appropriate btw
  - Binding positions are stored in NBT as `{binding: {X: ..., Y: ..., Z: ...}}` instead of `{poolX: ..., poolY: ..., poolZ: ...}`. For now, old binding NBT data is still accepted and converted.

Regression: If an impossible binding is created, like, out-of-range binding through NBT hacking, the experience is much worse.
- The UI will display an "x" (because `isBindingValid` returns false), but the binding will still *work* (because as written, a false result from `isBindingValid` is not a reason to actually de-establish a link - maybe the target chunk is not loaded? both checks are lumped into the same method).
- I didn't want to add half a dozen `isBindingValidDistance` `isTargetTileValidBinding` methods, because there's enough of them already, and that sounds error-prone.
- Flowers should not be *able to be* bound invalidly anyways, although the binding-radius centering might inadvertently create some. (hell, it's a 1.17 update anyways.)
- Still, is there a nicer way?


quat1024:
i had to rebase this and it wasn't pretty so if shit is fucked sorry

quat1024:
Might also be a good idea to stick `mana` into the superclass, since now it's the biggest chunk of duplicate code. (I'd rename the class though, maybe "mana special flower" or whatever.)

src/main/java/vazkii/botania/common/core/helper/MathHelper.java
```diff
@@ -50,4 +51,12 @@ public static int multiplyColor(int c1, int c2) {
 		return c1 & ~0xFFFFFF | r << 16 | g << 8 | b;
 	}
 
+	public static int distSqr(BlockPos a, BlockPos b) {
+		//Vec3i#distSqr, while convenient, offsets the second argument by (0.5, 0.5, 0.5).
+		//It also returns a double, which is unneeded; the squared distance between two integer coordinates is an integer.
+		int dx = a.getX() - b.getX();
+		int dy = a.getY() - b.getY();
+		int dz = a.getZ() - b.getZ();
+		return dx * dx + dy * dy + dz * dz;
```
Hubry, line 60: I can't help but worry that some sort of shenanigans is possible if you try to somehow bind something near the world border to spawn chunks. Please test.

src/main/java/vazkii/botania/api/subtile/TileEntityBindableSpecialFlower.java
```diff
@@ -0,0 +1,176 @@
...
+			if (cmp.contains("collectorX")) {
+				bindingPos = new BlockPos(cmp.getInt("collectorX"), cmp.getInt("collectorY"), cmp.getInt("collectorZ"));
+			} else if (cmp.contains("poolX")) {
+				bindingPos = new BlockPos(cmp.getInt("poolX"), cmp.getInt("poolY"), cmp.getInt("poolZ"));
+			}
+			//These versions of the mod also sometimes used a binding with a Y of -1 to signify an unbound flower.
+			//Currently, `null` is always used for unbound flowers. Coerce these positions to `null`.
+			if (bindingPos != null && bindingPos.getY() == -1) {
+				bindingPos = null;
+			}
```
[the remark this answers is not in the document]
Hubry: oh fuck i am blind and this is not relevant

src/main/java/vazkii/botania/api/BotaniaAPI.java
```diff
@@ -267,10 +267,11 @@ default boolean hasSolegnoliaAround(Entity e) {
...
 
 	/**
-	 * See config value "flower.forceCheck" for more information
+	 * Deprecated. The flower.forceCheck option has been removed.
 	 */
+	@Deprecated
 	default boolean shouldForceCheck() {
-		return false;
+		return true;
 	}
```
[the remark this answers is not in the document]
quat1024: 🦀

src/main/java/vazkii/botania/api/subtile/TileEntityBindableSpecialFlower.java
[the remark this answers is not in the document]
quat1024: Good catch though, i did not think about that at all 😅

src/main/java/vazkii/botania/common/core/helper/MathHelper.java
quat1024: Shenanigans are possible lol
image
i was able to bind a flower at 47000, 0 to a pool at 0, 0 because of the overflow

quat1024: Fixded by using longs which overflow at 3 billion, way bigger than a mc world. (I guess i could have used doubles like vanilla)

quat1024:
Ok think i'm done adding shit now

quat1024:
(Merge conflicts resolved with github.com user interface because i don't wanna mess with command line, not tested, don't merge pls)

quat1024:
Issue: flower auto-binding to nearest pool/spreader happens in `setPlacedBy`, which is only called when a player places the block, and not when `Level#setBlock` or `/setblock` does it. In the old system this was not an issue because auto-binding happened on the first tile entity tick, but here, it never happens in the tick method.

There is some kind of `Block#onPlace`, but it gets called before the BlockEntity exists, and other than `BlockEntity#setBlockState`, which is a totally inappropriate place to do it, there aren't any callbacks. Hm.

So I might have to bind on the first tick anyways.

(I noticed this only because of the flower-binding tests I'm working on :D)

src/main/java/vazkii/botania/api/subtile/TileEntityBindableSpecialFlower.java
```diff
@@ -0,0 +1,163 @@
...
+ * such as generating flowers to mana collectors, or mana-using functional flowers to pools.
+ * Implements bindability logic common to both types of flower.
+ */
+public abstract class TileEntityBindableSpecialFlower<T> extends TileEntitySpecialFlower implements IWandBindable {
+	protected @Nullable BlockPos bindingPos = null;
+	private final Class<T> tileClass;
+
+	private static final String TAG_BINDING = "binding";
+
+	public TileEntityBindableSpecialFlower(BlockEntityType<?> type, BlockPos pos, BlockState state, Class<T> tileClass) {
```
williewillus, line 39: tileClass -> bindClass and maybe document that it's the class that we want  to bind to

src/main/java/vazkii/botania/common/core/helper/MathHelper.java
williewillus: I don't really get why we're using longs here instead of double. Doesn't that lose precision?

williewillus (changes requested):
overall idea is okay but see comments

src/main/java/vazkii/botania/common/core/helper/MathHelper.java
williewillus: Oh you're multiplying integer inputs I'm stupid.

src/main/java/vazkii/botania/api/subtile/TileEntityBindableSpecialFlower.java
quat1024: how's this?

quat1024:
I'm not entirely happy with this but I gotta stop somewhere, lol.

 `findClosestTarget` should probably not be `abstract`, but it currently needs to leverage `IManaNetwork` which is different among the two flowers.

Also not super happy w/ wand of the forest overlays, the getBindingPos / getBinding distinction in this PR makes that look real ugly

williewillus:
Good enough yolo

PR #3833, opened 2021-10-04
(1.17.x-fabric) Use dedicated sound events instead of vanilla ones

### LINKED ISSUE #3829: Stop using vanilla sound events for proper subtitles
Throughout the Botania codebase, we do things like `playSound(SoundEvents.WOOL_PLACE, ...)` where the sound event is one of vanilla's. We should not be doing this.

The reason is that each sound event has its own unique subtitle, but we play a lot of these sounds outside of their original vanilla contexts. We were simply lazy and wanted to get the same sound for the least effort. Folks that are hard of hearing will see misleading vanilla subtitles instead of whatever botania related sound actually is happening.

Migrate all instances where we do this to use our own sound event we declare in our sounds.json, which piggybacks off of vanilla's sound event but with a proper Botania-fied subtitle.

Not prioritizing this myself since it's a huge slog, but if you have any questions please ask.

quat1024:
Closes #3829


quat1024:
remaining atm: flower sounds & the gaia guardian spawn/death/TP sounds. dont wanna burn out lol

quat1024:
I didn't add sound events for the Gourmaryllis because the subtitles are just "Eating" and "Burp", which i think are good enough.

The arcanerose also has a generic-enough "Experience gained" subtitle, but i added a sound event anyways so i could do "Experience absorbed". Idk.

Many of the item and block sounds are fairly clear-cut but a lot of these flower sounds are more judgement-calls.

src/main/java/vazkii/botania/common/item/rod/ItemMissileRod.java
```diff
@@ -84,7 +84,7 @@ public boolean spawnMissile(Level world, LivingEntity thrower, double x, double
 		missile.setPos(x, y, z);
 		if (missile.findTarget()) {
 			if (!world.isClientSide) {
-				missile.playSound(ModSounds.missile, 1F, 0.8F + (float) Math.random() * 0.2F);
+				missile.playSound(world.random.nextInt(100) == 0 ? ModSounds.missileFunny : ModSounds.missile, 1F, 0.8F + (float) Math.random() * 0.2F);
```
quat1024, line 87: hubyr wanted this it wasnt my idea

src/main/java/vazkii/botania/common/item/rod/ItemTerraformRod.java
```diff
@@ -127,9 +127,7 @@ private void terraform(ItemStack stack, Level world, Player player) {
 			}
 
 			if (!blocks.isEmpty()) {
-				for (int i = 0; i < 10; i++) {
-					world.playSound(player, player.getX(), player.getY(), player.getZ(), SoundEvents.SAND_STEP, SoundSource.BLOCKS, 1F, 0.4F);
-				}
+				world.playSound(player, player.getX(), player.getY(), player.getZ(), ModSounds.terraformRod, SoundSource.BLOCKS, 1F, 1F);
```
quat1024, line 130: This sound is still much more than loud enough, even if it's not played 10 times. I think it sounds the same when played 10x anyways

src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -3334,47 +3334,84 @@
...
   "botania.subtitle.agricarnation": "Agricarnation glimmers",
-  "botania.subtitle.altarCraft": "Petal Apothecary finishes",
+  "botania.subtitle.airRod": "Rod of the Skies whooshes",
+  "botania.subtitle.altarCraft": "Petal Apothecary twinkles",
   "botania.subtitle.arcane_rose_disenchant": "Rosa Arcana disenchants",
+  "botania.subtitle.astrolabeConfigure": "Worldshaper's Astrolabe clinks",
   "botania.subtitle.babylonAttack": "Key of King's Law unleashed",
   "botania.subtitle.babylonSpawn": "Key of King's Law activated",
-  "botania.subtitle.bellows": "Manatide Bellows hisses",
+  "botania.subtitle.bellows": "Manatide Bellows hiss",
```
williewillus, line 3345: should it still be hisses to match all the other ones?

src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -3334,47 +3334,84 @@
...
   "botania.subtitle.terraPickMode": "Terrasteel Pick powers up",
-  "botania.subtitle.terrasteelCraft": "Terrasteel finishes",
-  "botania.subtitle.thermalily": "Thermalily consumes lava",
+  "botania.subtitle.terraformRod": "Land terraforms",
+  "botania.subtitle.terrasteelCraft": "Terrasteel shines",
+  "botania.subtitle.thermalily": "Thermalily sips",
+  "botania.subtitle.thornChakramThrow": "Thorn Chakram thrown",
+  "botania.subtitle.tigerseyePacify": "Tigerseye petrifies",
   "botania.subtitle.unholyCloak": "Cloak of Sin activates",
+  "botania.subtitle.vineBallThrow": "Vine Ball flies",
```
williewillus, line 3412: seems a bit weird, maybe just `is thrown` like ender air?

src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -3334,47 +3334,84 @@
...
+  "botania.subtitle.virusInfect": "Virus infests",
```
williewillus, line 3413: maybe `infects` is better

src/main/resources/assets/botania/lang/en_us.json
quat1024: oops, missed that one

src/main/resources/assets/botania/lang/en_us.json
quat1024: took this wording from snowball subtitle i think

artemisSystem:
Not sure if it's within the scope of this PR, but during casual play we've noticed these two subtitles could be better:
- https://github.com/quat1024/Botania/blob/9df405fdaad956076c861731af12f91dba3de442/src/main/resources/assets/botania/lang/en_us.json#L3405 (Plays when terra shatterer powers up or down. Should probably at least say terra shatterer instead of terrasteel pick)
- Oh you already fixed the "player dashes forward" one

src/main/resources/assets/botania/lang/en_us.json
quat1024: I'll change it back for consistency with the rest of the lexica, but I'm not sure if it's actually correct - I interpreted "bellows" as a plural noun, but other parts of the lexica say things like "The Manatide Bellows is", which sounds wrong to my ears (i would use "bellows are")

PR #3837, opened 2021-10-14
Livingwood/Dreamwood rework

artemisSystem:
Thought i'd just open this draft PR to keep track of everything that needs to be done/taken care of, cause it's a lot. We realized there would actually have to be a decent amount of code changes, so Hubry suggested we just make the textures to begin with.

### Textures to do:

- [x] Livingwood log side
- [x] Livingwood log top
- [x] Stripped livingwood log top
- [x] Stripped livingwood log side
- [x] Dreamwood log side
- [x] Dreamwood log top
- [x] Stripped dreamwood log side
- [x] Stripped dreamwood log top
- [x] Livingwood planks
- [x] Dreamwood planks
- [x] Glimmering livingwood log side
- [x] Glimmering stripped livingwood log side
- [x] Glimmering dreamwood log side
- [x] Glimmering stripped dreamwood log side
- [x] Mana spreader
- [x] Pulse mana spreader
- [x] Elven mana spreader
- [x] Gaia mana spreader
- [x] Elven gateway core, natura pylon
- [x] Mossy livingwood planks
- [x] Mossy dreamwood planks
- [x] Framed livingwood
- [x] Pattern framed livingwood 
- [x] Framed dreamwood
- [x] Pattern framed dreamwood 
- [x] Livingwood avatar
- [x] Manatide bellows
- [x] Abstruse platform, Spectral platform, Infrangible platform
- [x] Drum of the wild
- [x] Drum of the gathering
- [x] Drum of the canopy
- [x] Manastorm charge
- [x] Mana blaster + lens clip
- [x] Spreader turntable
- [x] Corporea crystal cube
- [x] Livingwood twig, Dreamwood twig
- [x] All tools and rods (they have livingwood/dreamwood handles)
- [x] Wand of the forest
- [x] Open crate
- [x] Crafty crate
- [x] Shimmerwood planks

### Code changes to do/things to keep note of:

- [x] Blocks to add:
  - [x] Livingwood Log
  - [x] Glimmering Livingwood Log
  - [x] Stripped Livingwood Log
  - [x] Stripped Livingwood
  - [x] Stripped Livingwood Stair/Slab/Wall
  - [x] Glimmering Stripped Livingwood
  - [x] Glimmering Stripped Livingwood Log
  - [x] Dreamwood Log
  - [x] Glimmering Dreamwood Log
  - [x] Stripped Dreamwood Log
  - [x] Stripped Dreamwood
  - [x] Stripped Dreamwood Stair/Slab/Wall
  - [x] Glimmering Stripped Dreamwood
  - [x] Glimmering Stripped Dreamwood Log
- [x] Clean up various other models that use livingwood, and make sure nothing refers to the `livingwood.png` or `dreamwood.png` textures (which don't exist anymore).
- [x] Add the tags `#botania:livingwood_logs` and `#botania:dreamwood_logs`, which should be members of `#minecraft_logs_that_burn` (and transitively become members of `#minecraft:logs`).
- [x] Add the tags `#botania:glimmering_livingwood_logs` and `#botania:glimmering_dreamwood_logs`, which should be members of `#botania:livingwood_logs` and `#botania:dreamwood_logs`, respectively.
- [x] Change most (if not all) recipes that use `botania:livingwood` or `botania:dreamwood` to use `#botania:livingwood_logs` or `#botania:dreamwood_logs`, respectively. Add recipes for livingwood log -> livingwood, etc.
- [x] I believe the Pure Daisy recipe for Livingwood uses `#minecraft:logs`. That would create a loop where livingwood log would become itself. Possible solutions:
  - Add an ignore-list of sorts to the Pure Daisy recipes, so that it can ignore `#botania:livingwood_logs` and maybe `#botania:dreamwood_logs`.
  - Add a check in the Pure Daisy to make it not transform blocks into themselves. This would make it possible to "purify" livingwood, stripped livingwood/logs and glimmering stripped/regular livingwood/logs back into livingwood logs. (I like this one best from a player standpoint.)
  - Or just do the first option but only ignore `botania:livingwood_log`. More versatile for datapacks, but functionally the same as the second option.
- [x] Make the Pure Daisy retain all blockstates from the input block that apply to the output block. This would make logs retain their rotation when they're turned to livingwood logs.
- [x] Currently (as far as i'm aware), only `botania:dreamwood` stops the flight of minecarts coming from spectral rails, but not the stairs, slabs, walls, planks, etc. There should probably be a tag that controls what stops it, if there isn't already. Should at least contain `#botania:dreamwood_logs`, i'm not sure about stairs/slabs/walls/planks etc.
  - Currently only stopping on `#botania:dreamwood_logs` (which includes `#botania:glimmering_dreamwood_logs`).
- [x] As discussed in discord, instead of the elven portal requiring `botania:livingwood` and `botania:glimmering_livingwood`, it should just need `#botania:livingwood_logs` and `#botania:glimmering_livingwood_logs`. That way you could make the portal out of stripped livingwood if you wanted, or even make the whole portal glimmering. ~~Maybe add a stripped elven gateway core (and making `#botania:elven_gateway_cores`?) to match. Should probably then be added to the JEI/REI elven trade category.~~
- [x] ~~Because of how the blockstate multipart system works and is used for walls, to avoid z-fighting on dreamwood and stripped dreamwood walls, we will have to:~~ Nevermind, blockstate picks are consistent for a given block coordinate, so all the active multiparts will have the same index of model, meaning there will be no z-fighting.
  - ~~Create a new `dreamwood_template_wall_side.json`, `dreamwood_template_wall_post.json` and `dreamwood_template_wall_side_tall.json`, which are like `template_wall_side.json` etc. except it allows specifying a different texture to be used for the top and bottom faces than the rest of the faces.~~
  - ~~Use those templates in the models 1-6 for stripped and normal dreamwood walls, specifying the same texture every time for the tops/bottoms, using the textures 1-6 for the rest of the faces~~
- [x] The blockstates for all the slabs and stairs and walls etc should have `"uvlock": true`
- [x] Maybe rename Framed Livingwood etc. to Framed Livingwood Planks and add them to `#minecraft:planks`
- [x] Add livingwood, dreamwood and shimmerwood planks + stairs and slabs to the respective `#minecraft:planks/wooden_stairs/wooden_slabs` tags (i guess this was just missed when moving to 1.14+, but makes sense to include here).
- [x] Look through lexica botania to see if any changes are necessary.


src/main/java/vazkii/botania/common/block/ModBlocks.java
```diff
@@ -966,6 +969,25 @@ public static void addAxeStripping() {
...
+	public static Map<Block, Block> getCustomStripping() {
+		return Collections.unmodifiableMap(CUSTOM_STRIPPING);
+	}
+
+	public static BlockState getStrippedBlock(BlockState state) {
+		Block block = CUSTOM_STRIPPING.get(state.getBlock());
+		if (block == null) {
+			return null;
+		}
+		return block.withPropertiesOf(state);
```
williewillus, line 990: til withPropertiesOf

williewillus:
yay

I'll take a brief look today but won't be able to give it an in-depth look until Monday or Tuesday

Common/src/main/java/vazkii/botania/common/block/ModBlocks.java
```diff
@@ -206,20 +208,31 @@
...
 	public static final Block livingrockBrickMossy = new BlockMod(BlockBehaviour.Properties.copy(livingrock));
 
-	// TODO 1.16+ livingwood/dreamwood should support leaves?
-	public static final Block livingwood = new BlockMod(BlockBehaviour.Properties.of(Material.WOOD).strength(2).sound(SoundType.WOOD));
-	public static final Block livingwoodPlanks = new BlockMod(BlockBehaviour.Properties.copy(livingwood));
-	public static final Block livingwoodPlanksMossy = new BlockMod(BlockBehaviour.Properties.copy(livingwood));
-	public static final Block livingwoodFramed = new BlockMod(BlockBehaviour.Properties.copy(livingwood));
-	public static final Block livingwoodPatternFramed = new BlockMod(BlockBehaviour.Properties.copy(livingwood));
-	public static final Block livingwoodGlimmering = new BlockMod(BlockBehaviour.Properties.copy(livingwood).lightLevel(s -> 12));
+	public static final Block livingwoodLog = new RotatedPillarBlock(BlockBehaviour.Properties.of(Material.WOOD).strength(2).sound(SoundType.WOOD)); // TODO color
```
williewillus: what does this todo mean?

williewillus:
I'm merging this as it overall looks good but a couple things (this is the first time I've actually taken a look at the textures in detail so you can take this as first impressions feedback)

1. The todo I commented on below
2. The manasteel shovel looks a bit strange to me. The shading/shape of the head makes it so that it kinda looks like a jewel or something on a staff.
3. Similarly, I'm not sure about the elementium shovel. It's probably useless feedback in this case because I can't really put my finger on why, it just slightly bothers me. I think the handle feels too thin right before the head so it looks like a lollipop or something in-hand.

Either way, feel free to follow up with more fixes. Thanks for your hard work.

williewillus (commented):
.

williewillus:
There are also some minor code cleanups I spot but I can take care of them.

Common/src/main/java/vazkii/botania/common/block/ModBlocks.java
artemisSystem: Oh, that was from @Hubry, i had to ask about that as well a while ago. iirc it's for correctly setting the map color for the logs, since the vanilla logs have a different color depending on if they're vertical or horizontal. I guess we forgot about it

PR #3841, opened 2021-10-19
fix https://github.com/williewillus/botania-fabric-issues/issues/5

MoriyaShiine:


Hubry:
why the formatting change on mixin json

PR #3843, opened 2021-10-21
(hackily) fix https://github.com/williewillus/botania-fabric-issues/i…

MoriyaShiine:
…ssues/35


williewillus:
Meh we'll fix this mess when it blows up inevitably

PR #3844, opened 2021-10-22
Fixed mana infusion recipes and the pebble tag in gog

Voleil:


Voleil:
If nothing went wrong this time this one should be the good one  ^

PR #3861, opened 2021-11-16
Add cacophonium & flower binding tests, + some testing documentation and utilities [CI fails]

quat1024:
we do a little gametesting

CI currently fails because of broken flower-binding, #3831 fixes it

- Cacophonium test: fake player clicking on a cow with a cacophonium, then clicking on a note block
- Flower automatic binding tests: places a pool/spreader, places floating flowers around the target, checks that the ones that are close enough bound to it.
- Flower manual binding tests: places flowers then places a spreader, and a fake player uses a Wand of the Forest to try and bind each flower to the target.
- Obedience Stick test: same but uses a floral obedience stick.

One shortcoming of the flower binding tests is they only check 6 "it should work" positions and 6 "it shouldn't work" positions, all axis-aligned with the bind target. Mmm. Might be worth revisiting.


quat1024:
The flower-binding tests are going to fail by the way (sorry about your email accounts). I'm trying to make my flower-binding-new PR make the tests pass.

PR #3866, opened 2021-11-21
Improve corporea index messages

Alwinfy:
- Add special cases for "all [x]" and "count [x]"
 - Tweak localization to be simpler/more uniform


src/main/java/vazkii/botania/common/block/tile/corporea/TileCorporeaIndex.java
```diff
@@ -372,12 +374,25 @@ public static void clearCache() {
 		serverIndexes.clear();
 	}
 
+	private static IFormattableTextComponent formatRequest(int inputCount, ITextComponent requestName, ICorporeaResult res) {
```
williewillus, line 377: just put this on the result class?

williewillus: or on the request. The count can be saved from the initial count, the name can be gotten from the matcher.

Alwinfy: Problem is that the `CorporeaRequest` object never gets handled by TileCorporeaIndex, it's all inside CorporeaHelper.requestItem()

PR #3890, opened 2021-12-25
Fix hopperhock inserting into the wrong side (#3889)

### LINKED ISSUE #3889: Hopperhock inserts into opposite side (Fabric edition)
### Mod loader

Fabric

### Minecraft version

1.18.1

### Botania version

1.18.1-427-fabric

### Modloader version

Fabric Loader 0.12.11

### Modpack info

_No response_

### The latest.log file

...

### Issue description

When the hopperhock inserts an item into a sided inventory, it inserts into the opposite side from what's expected.

### Steps to reproduce

1. Place a furnace in the world.
2. Place a hopperhock above or below the furnace.
3. Throw an item on the ground.

Expected behaviour:
The item should be placed in the input slot if the hopperhock is above the furnace, or in the fuel slot if the hopperhock is below.

Actual behaviour:
The item is placed in the opposite slot from the expected one. If that slot is full, the item entity simply disappears.

### Other information

_No response_

atakle:
Fix the hopperhock inserting items into the opposite side from where it's positioned.

Fixes #3889


atakle:
Since this has happened before, I thought it might be useful to write some tests (I also wanted to try the GameTest system). I hope that's OK.

PR #3903, opened 2022-01-14
(1.18.x-fabric) Update zh-cn Localization

aquila-zyy:
Translate most untranslated entries. Update old descriptions. Add interlinks to the Lexicon. Replace most half-width punctuations with full-width ones. Rewrite many poorly translated sentences.


aquila-zyy:
Also may I ask why there were so many random whitespaces all over the place in the lang file. It's almost as if someone used to use those spaces as line breaks or something. 

Had to delete whitespace en masse which might have resulted in some missing line breaks in the json file.

williewillus:
> Also may I ask why there were so many random whitespaces all over the place in the lang file. It's almost as if someone used to use those spaces as line breaks or something.

I believe the original intent was to keep the line numbers synced with the en_us file. I don't really have an opinion one way or another, so if it's not useful to you feel free to get rid of it.

aquila-zyy:
Oh, I meant the white spaces were in the string literals, which made me feel like those places were where the line breaks in the lexicon were in old versions of the mod.

There were also some instances where a Chinese word would be splitted into two halves and marked by two separate markers. I would guess for some reasons those markers did not work across multiple lines in old versions so the original contributor had to do that as a work around. 

I've already removed most of those but I didn't reorder the items to make it in sync with en_us.

williewillus:
oh ok, cool. Also I saw you have lot of commits, is this ok to merge now or there's still more to do?

aquila-zyy:
Oh sorry for the huge amount of commits! Every time I thought I was done I would find another small error that I missed. 

Yes, I think it's fine to merge now. I've been playing with it for a few hours and have yet to find a thing to correct. 

Glad to contribute to this mod.

PR #3905, opened 2022-01-15
Make recipes for cracked and mossy things vanilla-consistent

artemisSystem:
- Changes cracked livingrock brick recipe to be smelting livingrock brick in furnace, instead of crafting it with cobblestone
- Changes recipes for mossy things to use vines or moss blocks instead of seeds
- This change is 7 years overdue (1.8) ~~except it's only 5 since that's how long ago botania 1.10 released, which was the next version after 1.7~~


williewillus:
Looks good but rebase to 1.18.x please

artemisSystem:
Will do

artemisSystem:
Don't merge yet i forgot to actually check if it builds and puts all the datagen in the right place after merging

artemisSystem:
Ok it seems good

PR #3909, opened 2022-01-17
(1.18.x-fabric) Make A Few Tweaks to the Starcaller

aquila-zyy:
- Stars produced by the Starcaller that materialize inside of blocks will now continue to fly until they reach open space, at which point they will become solid.
- Now summons a star only once per 12 ticks to prevent spam clicking.
- The projectile now ignores villagers, as trading with them while holding the sword will summon a star that could potentially kill them.


src/main/java/vazkii/botania/common/entity/EntityFallingStar.java
```diff
@@ -78,6 +91,10 @@ public void tick() {
 	protected void onHitEntity(@Nonnull EntityHitResult hit) {
 		super.onHitEntity(hit);
 		Entity e = hit.getEntity();
+		// Blacklisting villagers since trading with them counts as a "swing" and will summon a star.
```
williewillus, line 94: Trading with a villager counts as left clicking them?

src/main/java/vazkii/botania/common/item/equipment/tool/ItemStarSword.java
```diff
@@ -36,37 +40,49 @@ public ItemStarSword(Properties props) {
...
 
-			if (player.getMainHandItem() == stack && player.attackAnim == check && !world.isClientSide) {
-				BlockHitResult pos = ToolCommons.raytraceFromEntity(player, 48, false);
-				if (pos.getType() == HitResult.Type.BLOCK) {
-					Vec3 posVec = Vec3.atLowerCornerOf(pos.getBlockPos());
-					Vec3 motVec = new Vec3((0.5 * Math.random() - 0.25) * 18, 24, (0.5 * Math.random() - 0.25) * 18);
-					posVec = posVec.add(motVec);
-					motVec = motVec.normalize().reverse().scale(1.5);
+		// Initialize timer for new items or items given to another player.
+		if (!ItemNBTHelper.verifyExistance(stack, STAR_CHARGED) || ItemNBTHelper.getInt(stack, STAR_CHARGED, 0) > player.tickCount) {
```
williewillus, line 48: is `player.tickCount` persistent or it restarts at 0 every time you load the world?

src/main/java/vazkii/botania/common/entity/EntityFallingStar.java
aquila-zyy: Oddly enough yes. So does opening a chest or a wooden door. Basically all interactions.

I'll admit that I can't think of a decent solution to this given how it is implemented now.

Making a new (data) tag is a possible improvement if we want the blacklist to be modifiable I guess?

src/main/java/vazkii/botania/common/item/equipment/tool/ItemStarSword.java
aquila-zyy: I'm not sure about this one. Looks like it counts the number of ticks this entity has lived since creation, so I guess it will reset? 

But the code also checks if the value on the tag is greater than the bearer's tickCount, and it will always ensure the former is no greater than the latter, so the logic seems fine to me no matter what.

williewillus:
@aquila-zyy hi, sorry for the delay. Can you rebase your PR? Will try to get this in before releasing 1.18.2

aquila-zyy:
Uhh, so I just rebased it but it looks like there're still conflicts, and commits from my previous pr are also brought over. Have I done something wrong? Sorry for the trouble. I'm not really familiar with git operations.

williewillus:
ok yeah it's all messed up now, maybe try making a new PR from scratch?

aquila-zyy:
Will do. I'll close this one first.

PR #3825, opened 2021-10-01
Let corporea interceptors intercept any failed request (#3810)

artemisSystem:
Closes #3810

PR #3910, opened 2022-01-17
Allow Corporea Interceptors to intercept any failed request

### LINKED ISSUE #3810: Make corporea interceptors intercept any failed request if they don't have an item frame
Currently, an interceptor without an item frame (or an empty one) doesn't do anything.

artemisSystem:
Fixes #3810 

Supersedes #3825

### DIFF
--- a/src/main/java/vazkii/botania/test/block/InterceptorTest.java
+++ b/src/main/java/vazkii/botania/test/block/InterceptorTest.java
@@ -1,56 +1,64 @@
+/*
+ * This class is distributed as part of the Botania Mod.
+ * Get the Source Code in github:
+ * https://github.com/Vazkii/Botania
+ *
+ * Botania is Open Source and distributed under the
+ * Botania License: http://botaniamod.net/license.php
+ */
 package vazkii.botania.test.block;
 
 import net.minecraft.core.BlockPos;
 import net.minecraft.gametest.framework.GameTest;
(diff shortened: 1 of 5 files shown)


src/main/java/vazkii/botania/test/block/InterceptorTest.java
```diff
@@ -0,0 +1,64 @@
...
+			assertHasPendingRequest(helper, WILDCARD_RETAINER);
+			assertHasPendingRequest(helper, BOOTS_RETAINER);
+		}).thenSucceed();
+	}
+
+	private static void assertHasPendingRequest(GameTestHelper helper, BlockPos retainerPos) {
+		assertHasPendingRequest(helper, retainerPos, false);
+	}
+
+	private static void assertHasPendingRequest(GameTestHelper helper, BlockPos retainerPos, Boolean invert) {
```
williewillus:
```suggestion
	private static void assertHasPendingRequest(GameTestHelper helper, BlockPos retainerPos, boolean invert) {
```

williewillus (commented):
small nit. Looks good otherwise.

src/main/java/vazkii/botania/test/block/InterceptorTest.java
artemisSystem: Is this something that could be added to spotless?

PR #3913, opened 2022-01-21
Have the Kanade's regen apply to pets as well

Alwinfy:


Common/src/main/java/vazkii/botania/common/block/subtile/functional/SubTileFallenKanade.java
```diff
@@ -50,6 +53,11 @@ public void tickFlower() {
 		}
 	}
 
+	private static boolean canHeal(LivingEntity e) {
+		// heal pets and players
+		return e instanceof Player || e instanceof TamableAnimal && ((TamableAnimal) e).isTame() || e instanceof AbstractHorse && ((AbstractHorse) e).isTamed();
```
williewillus: you can use the fancy new pattern variable stufff now

`e instanceof TamableAnimal tamable && tamable.isTame() || e instanceof AbstractHorse horse && horse.isTamed()`

Hubry: This does not check if the entity is alive, nor does it check for spectator mode players.

PR #3932, opened 2022-01-27
Remove Pestle and mortar

artemisSystem:
Removes Pestle and Mortar item, the recipes that used it before are now just 1->1 of petal/mushroom to dye


williewillus:
Update lexicon?

williewillus (changes requested):
lexicon update needed

PR #3941, opened 2022-02-01
Fix mana prism right click interaction

artemisSystem:
Makes the mana prism not eat clicks when a right click doesn't do anything


Common/src/main/java/vazkii/botania/common/block/mana/BlockPrism.java
```diff
@@ -120,6 +120,8 @@ public InteractionResult use(BlockState state, Level world, BlockPos pos, Player
 		} else if (!lens.isEmpty()) {
 			player.getInventory().placeItemBackInInventory(lens);
 			prism.getItemHandler().setItem(0, ItemStack.EMPTY);
+		} else {
+			return InteractionResult.PASS;
 		}
 
 		return InteractionResult.SUCCESS;
```
williewillus, line 127: I'm surprised java or intellij did not complain about dead code on this line

Hubry: It's not dead, the if above the addition doesnt return

PR #3946, opened 2022-02-04
Spreader tweaks

artemisSystem:
- Changes the spreader padding model:
  - Makes it point upward by default (and just rotating it when it's rendered) so that if you point it up, all the side textures are pointing the same way
  - Added front, back, and side texture slots to allow for in the future making "spreader sweaters" that was briefly discussed a while ago. Basically custom textures for each padding color that each have their own pattern, think llama carpets.
- Shrinks mana spreaders to have a 12x12 hitbox, 14x14 with wool padding
- Adds the ability to surround a spreader with scaffolding
  - Changes its hitbox to a full block, which allows things like buttons and redstone dust to be placed on it
  - Also makes it so that water doesn't flow out when it's waterlogged for some reason (current spreaders with their full block hitbox also do this), but that's kinda neat so why not i guess.


Common/src/main/java/vazkii/botania/client/render/tile/RenderTileSpreader.java
```diff
@@ -83,20 +84,36 @@ public void render(@Nonnull TileSpreader spreader, float partialTicks, PoseStack
...
 		}
 
 		if (spreader.paddingColor != null) {
-			BakedModel model = getPaddingModel(spreader.paddingColor);
+			ms.pushPose();
+			// The padding model is facing up so that the textures are rotated the correct way
+			// It's simpler to do this than mess with rotation and UV in the json model
+			ms.translate(0.5F, 0.5F, 0.5F);
+			ms.mulPose(Vector3f.XP.rotationDegrees(-90));
+			ms.mulPose(Vector3f.YP.rotationDegrees(180));
```
williewillus, line 99: what's the purpose of this? is the padding model not rotated correctly around Y axis?

williewillus (commented):
looks good, just one question

Common/src/main/java/vazkii/botania/client/render/tile/RenderTileSpreader.java
artemisSystem: The padding model has the "hole"/opening for the spreader at the top, instead of pointing "forwards" like the spreader model itself. This is so the wool texture faces in a consistent direction on all the faces. As mentioned in the comment, it's possible to do this by rotating and plotting in UV values in the model itself, but that's annoying and error-prone. when the "hole" is pointing up, the textures on all the 4 sides around it are facing the same direction (up), and look consistent. (If you make the hole on one of the sides, like it is currently/before this pr, textures will be pointing in different directions). Images: (Red lines are to illustrate the "flow"/direction of the wool texture)

Old:
image

New:
image

williewillus: oh i was more confused about the Y 180 rotation specifically
the whole sub-branch of code i understand why it's there 
but to go pointing up -> pointing forward just needs an x rotation right?

artemisSystem: it needs the y rotation because: lining up all the textures requires rotating either the "front" face or the "back" face (really top and bottom in the model) 180 degrees. By just doing the x rotation, it turns out the face that needs the rotation to be upright when the spreader is facing forwards, is the front/`up` face. but that face is on 4 different cuboids and requires tweaking the UV mapping to rotate around, whlie the back face can be made to be on just 1 cuboid. so it's easier to rotate that and then rotate the whole model 180deg

PR #3947, opened 2022-02-05
Tweak models for framed livingwood/dreamwood

artemisSystem:
- Rotated all textures to face the right way, so they always match up with nearby planks (Before, it looked really weird when you'd get for example two lines of dark pixels next to eachother across a block border)
- Item model rotated sideways to more easily distinguish from pattern framed variant

### DIFF
--- a/Common/src/main/java/vazkii/botania/data/ItemModelProvider.java
+++ b/Common/src/main/java/vazkii/botania/data/ItemModelProvider.java
@@ -366,6 +366,11 @@ public class ItemModelProvider implements DataProvider {
 			consumer.accept(ModelLocationUtils.getModelLocation(i), new DelegatedModel(prefix("block/petal_block")));
 		});
 
+		takeAll(itemBlocks, ModBlocks.livingwoodFramed.asItem(), ModBlocks.dreamwoodFramed.asItem()).forEach(i -> {
+			String name = i == ModBlocks.livingwoodFramed.asItem() ? "livingwood" : "dreamwood";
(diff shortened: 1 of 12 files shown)


artemisSystem:
Some comparison images (first is new, second is old):

Hubry (approved):

Common/src/main/java/vazkii/botania/data/ItemModelProvider.java
```diff
@@ -366,6 +366,11 @@ private void registerItemBlocks(Set<BlockItem> itemBlocks, BiConsumer<ResourceLo
 			consumer.accept(ModelLocationUtils.getModelLocation(i), new DelegatedModel(prefix("block/petal_block")));
 		});
 
+		takeAll(itemBlocks, ModBlocks.livingwoodFramed.asItem(), ModBlocks.dreamwoodFramed.asItem()).forEach(i -> {
+			String name = i == ModBlocks.livingwoodFramed.asItem() ? "livingwood" : "dreamwood";
+			consumer.accept(ModelLocationUtils.getModelLocation(i), new DelegatedModel(prefix("block/framed_" + name + "_horizontal_z")));
+		});
```
williewillus, line 372: would rather just have two separate `accept` lines instead of a take + extra if check

williewillus (approved):
looks good but see the nit

artemisSystem:
It says build failed but spotless doesn't bring anything up locally, i think there's an issue with the ci sometimes

PR #3949, opened 2022-02-09
[1.19.x] Add french translation (fr_fr)

Aegide:
The translation is not complete, but it's just to give it more "visibility".

I used the current English translation as a new base, since it was so different from the old fr_FR file.
I replaced the entries already translated, and fixed some of those entries.

Currently i'm working on adding the missing translations.


Aegide:
@williewillus It's over. It's done.

williewillus:
how

PR #3958, opened 2022-02-16
Fix rafflowsia streak and buffer size

artemisSystem:
- `getMaxStreak` seemed to be wrong and hubry agreed
- Max mana was never adjusted for the max streak reward


Hubry:
Uh, I think I have to revert half of this after a closer look.

PR #3973, opened 2022-03-02
Modernize petal apothecary recipes

artemisSystem:
- Changes the apothecary recipe to not use slabs
- Removes the behavior of throwing a vine onto an apothecary to make it mossy
- Adds a recipe for the mossy apothecary using mossy cobblestone
- Changes the recipes for the metamorphic stone apothecaries to be like the other apothecary recipes


Common/src/main/java/vazkii/botania/data/recipes/RecipeProvider.java
```diff
@@ -876,22 +876,29 @@ private void registerMisc(Consumer<FinishedRecipe> consumer) {
...
+		ShapedRecipeBuilder.shaped(ModBlocks.mossyAltar)
+				.define('P', ModTags.Items.PETALS)
+				.define('C', Items.MOSSY_COBBLESTONE)
+				.pattern("CPC")
 				.pattern(" C ")
 				.pattern("CCC")
 				.unlockedBy("has_item", conditionsFromTag(ModTags.Items.PETALS))
 				.save(consumer);
 		for (String metamorphicVariant : LibBlockNames.METAMORPHIC_VARIANTS) {
-			Block altar = Registry.BLOCK.getOptional(prefix("apothecary_" + metamorphicVariant.replaceAll("_", ""))).get();
```
artemisSystem, line 887: Why was this `.replaceAll("_", "")` even here? None of the variant names have `_` and i don't see why you'd strip them anyway

williewillus: don't remember, check the blame

artemisSystem: Looks like @Hubry made it like that when setting up the datagen two years ago, doesn't seem like there was a particularly good reason for it

Hubry: Might have been just copypasting from an another spot? I have no idea anymore

PR #3997, opened 2022-04-01
Make A Few Tweaks to the Starcaller (re-PR)

aquila-zyy:
This is a re-PR of #3909 cus I messed up the git history on that one. 

Cherrypicked the commits from the old branch and fixed a few imports that no longer worked.


williewillus:
thanks! There's a couple tweaks I'd make, but I'll do that in trunk after applying the patch.

PR #4007, opened 2022-04-10
Migrate to FabricItemGroupBuilder on Fabric, fixing a crash on Quilt

quat1024:
Getting bugs for two modloaders is already so much fun, how about getting bugs for three???????????

Quilt's reimplementation of Fabric API does not include `net.fabricmc.fabric.impl.item.group.ItemGroupExtensions`. This causes the game to crash on startup because Botania uses it [right here](https://github.com/VazkiiMods/Botania/blob/1.18.x/Fabric/src/main/java/vazkii/botania/fabric/FabricBotaniaCreativeTab.java#L31-L34) when creating its creative tab.

Quilt's implementation [does include a reimplementation of `FabricItemGroupBuilder`](https://github.com/QuiltMC/fabric-api-qsl/blob/1.18.2/fabric-item-groups-v0/src/main/java/net/fabricmc/fabric/api/client/itemgroup/FabricItemGroupBuilder.java). Unfortunately FabricItemGroupBuilder is a too-opinionated half-baked API, it only accepts a ResourceLocation for the creative tab name instead of a simple String, concatenating the halves of the ResourceLocation with a period.

It's literally not possible to create a creative tab that doesn't have a period in the translation key with this API, so I went ahead and changed the translation key on Forge to match. Sigh. The second commit in this PR renames the language keys in non-`en_us` locales; I also took the opportunity to remove the language entry if it was the same as `en_us`.

### Alternatives

Creating a creative tab without using *any* loader-provided APIs would basically require pasting the ItemGroupExtensions mixin into the Botania source tree, which sounds like a fun source of bugs later on down the line.


quat1024:
An accompanying PR, fixing it from Quilt's end, was created here: https://github.com/QuiltMC/fabric-api-qsl/pull/2. If that gets merged, this PR can be closed.

williewillus:
FIne with fixing it on our end, though I think you're missing one callsite, the root advancement (advancements/main/root.json) also refers to this lang key.

quat1024:
God dammit

Running the data generator now puts all the recipe advancements under `generated/resources/data/botania/advancements/recipes/botania.tab/`

quat1024:
I cant fix it lmao, gonna try a different way

quat1024:
Was fixed via https://github.com/VazkiiMods/Botania/commit/371fb2e862a96870f37b724c1b18ec01750c314e

PR #4028, opened 2022-04-28
Tweak the Tiny Potato's inventory

quat1024:
### Nerf tiny potatoes

I noticed that you can use a hopper to insert and remove items from a Tiny Potato. This is, without a doubt, one of the features of all time. Possibly ever.

However! Automation can insert full stacks of 64 items into each of the six slots of the tiny potato inventory, but a player right clicking can only add one item per slot. I found this really confusing because if you hopped items into a Tiny Potato it looked like they were disappearing.

So the first commit only allows tiny potatoes to hold 1 item per slot. This makes Tiny Potatoes much less effective as a mass storage solution, and I predict it will drastically upset the Tiny Potato meta. But it brings their automation-side interface closer to what people would expect it to be, based on how it feels to interact with the block.

### Buff tiny potatoes

Ok, so now comparators can read the fill-level of the tiny potato inventory. Why the hell not

### DIFF
--- a/Common/src/main/java/vazkii/botania/common/block/decor/BlockTinyPotato.java
+++ b/Common/src/main/java/vazkii/botania/common/block/decor/BlockTinyPotato.java
@@ -60,6 +60,20 @@ public class BlockTinyPotato extends BlockModWaterloggable implements EntityBloc
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


williewillus:
why is the signal stored vs recomputing like how chests do it?

williewillus:
Applied locally

PR #4033, opened 2022-05-08
Update mana_alchemy.json

Aegide:
Make the entry for "potato_unpoison" about "potato_unpoison"


williewillus:
So is this recipe already on another page?

Aegide:
Yes.

https://github.com/VazkiiMods/Botania/blob/master/Common/src/main/java/vazkii/botania/data/recipes/ManaInfusionProvider.java#L104

https://github.com/VazkiiMods/Botania/blob/master/Common/src/main/resources/assets/botania/patchouli_books/lexicon/en_us/entries/devices/mana_alchemy.json#L84

PR #4070, opened 2022-06-08
[1.18.2] Add EMI Integration

emilyploszaj:
Adds pretty comprehensive integration with EMI, including recipe defaults and tag translations/exclusions. Hopefully this follows style guidelines I was not able to get the task to run.


Fabric/src/main/java/vazkii/botania/fabric/integration/emi/OrechidEmiRecipe.java
```diff
@@ -0,0 +1,61 @@
...
+import vazkii.botania.api.recipe.IOrechidRecipe;
+import vazkii.botania.api.recipe.StateIngredient;
+
+public class OrechidEmiRecipe extends BotaniaEmiRecipe {
+	private final EmiIngredient orechid;
+	private static MethodHandle handle;
+
+	static {
+		try {
+			handle = MethodHandles.lookup().findVirtual(IOrechidRecipe.class, "getOutput", MethodType.methodType(StateIngredient.class));
```
williewillus: ?? why are we reflecting?

williewillus:
Overall looks good, but can you explain what the json files in assets are for? The `defaults.json` seems kinda random.

Fabric/src/main/java/vazkii/botania/fabric/integration/emi/OrechidEmiRecipe.java
emilyploszaj: Oh wow, I forgot I did this, integration was originally written in yarn, and there was a signature conflict that wouldn't let me call this method, I'll remove it now that it's on mojmap proper.

emilyploszaj:
> Overall looks good, but can you explain what the json files in assets are for? The `defaults.json` seems kinda random.

The tag exclusions json is for telling EMI which tags should not even be considered for recipes, or displayed in UI. I listed some tags I thought were particularly non-user-facing instead of translating them.

The recipe defaults json indicates the default way to achieve an item, simply listing a recipe is shorthand for "apply to all outputs", which in most cases, is only 1. Recipe defaults let EMI automatically expand recipe trees, as well as pick the correct recipes to focus when doing a lookup.

The defaults json was created by me going through every item in Botania and determining if it had a reasonable recipe to set as its default. Base resources like manasteel don't have a default recipe, but crafted resources such as runes, trinkets, or functional blocks have their main method of acquisition defaulted.

emilyploszaj:
I have figured out how to run spotless, sorry for the style violations, they are now fixed.

emilyploszaj:
Default recipes sorted alphabetically at the request of hubry

PR #4085, opened 2022-07-04
Add support for Ears

TJT01:
This PR does the following when [Ears](https://ears.unascribed.com/) is installed:
- If phantom-inked armor is worn; Ears will act as if there isn't anything in that slot (feet claws will show with invisible boots, for example).
- If a Flügel tiara with wings is worn, Ears' wings will be hidden.
# Testing skin
This skin can be used to test compatibility. Simply use this as a normal Minecraft skin.
test

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/client/integration/ears/EarsIntegration.java
+++ b/Xplat/src/main/java/vazkii/botania/client/integration/ears/EarsIntegration.java
@@ -1,19 +1,26 @@
 package vazkii.botania.client.integration.ears;
 
+import com.unascribed.ears.api.EarsFeatureType;
 import com.unascribed.ears.api.OverrideResult;
+import com.unascribed.ears.api.registry.EarsInhibitorRegistry;
 import com.unascribed.ears.api.registry.EarsStateOverriderRegistry;
 
+import net.minecraft.world.Container;
 import net.minecraft.world.entity.EquipmentSlot;
 import net.minecraft.world.entity.player.Player;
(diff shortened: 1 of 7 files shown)


Xplat/src/main/java/vazkii/botania/client/integration/ears/EarsIntegration.java
```diff
@@ -0,0 +1,37 @@
...
+
+import net.minecraft.world.entity.EquipmentSlot;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+
+import vazkii.botania.api.item.IPhantomInkable;
+
+public class EarsIntegration {
+	public void register() {
+		EarsStateOverriderRegistry.register("botania", (state, peer) -> {
```
williewillus: use LibMisc.MOD_ID instead of hardcoding the string

Xplat/src/main/java/vazkii/botania/client/integration/ears/EarsIntegration.java
```diff
@@ -0,0 +1,37 @@
...
+			if (!(peer instanceof Player player))
```
williewillus: braces on ifs (the linter/github action will yell for these)

williewillus (changes requested):
Check the lint errors (run `gradlew pmd spotlessCheck` locally). Also, what compatibility does this add? Can you put a brief summary?

williewillus:
lgtm thanks

PR #4092, opened 2022-07-17
Fix "Elves" being highlighted with item macro on elfResources page

Aegide:
I also did other minor changes.


Aegide:
Imo, it would be better to remove all the macros about elves, they are not really a botania-specific concept.

But keeping the macros about Alfheim would be fine.

PR #4112, opened 2022-08-14
Munchdew eating order weighted by distance from logs

TheRealWormbo:
This change makes the Munchdew randomly select leaves to eat from among those that are furthest away from logs (as specified via their distance property), rather than strictly iterating over X/Y/Z coordinates. The selection weighting contains an added amount of randomness to deviate a bit from strict Manhattan distance. The requirement for leaves to have air exposure before being selected remains untouched.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/subtile/generating/SubTileMunchdew.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/subtile/generating/SubTileMunchdew.java
@@ -13,6 +13,7 @@ import net.minecraft.core.Direction;
 import net.minecraft.nbt.CompoundTag;
 import net.minecraft.tags.BlockTags;
 import net.minecraft.world.level.block.Block;
+import net.minecraft.world.level.block.LeavesBlock;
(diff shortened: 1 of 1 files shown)


williewillus:
Can you add a video or something comparing the behaviours?

TheRealWormbo:
I couldn't get Github to recognize my video upload attempts, so here are a couple of YouTube links instead:

- [New behavior for this PR](https://youtu.be/DRaGUEbXekk) – eating from the outside inwards
- [Old behavior](https://youtu.be/ojC8qQccszA) – visually not very appealing IMHO, also starts leaf decay at the end, when leaves are no longer connected to any logs
- For comparison: [Munchdew in 1.12](https://youtu.be/VKsuUoBCeEw) – likely to disconnect leaves, causing saplings to drop eventually
  (That's basically what commit [ba3ea10](https://github.com/VazkiiMods/Botania/pull/4112/commits/ba3ea100fb3ad6a13c2706a9c4af1f69165f9f3a) restores, in case you prefer that behavior instead)

williewillus:
Sounds good. I'm going to apply this to 1.19 only though.

PR #4129, opened 2022-08-25
Disables entry about banners

Aegide:
In the version 1.18 of Botania, banner patterns are disabled.

So i feel like it would be better to hide this entry, otherwise players will think that the Lexica Botania is lying.

<br>

**Examples** :
- https://github.com/VazkiiMods/Botania/blob/1.18.x/Xplat/src/main/java/vazkii/botania/common/block/ModPatterns.java
- https://github.com/VazkiiMods/Botania/blob/1.18.x/Xplat/src/main/java/vazkii/botania/common/item/ItemModPattern.java


williewillus:
sure

PR #4134, opened 2022-08-27
[1.19] Add bifrost pane recipe

Aegide:
Bifrost pane had no Patchouli tooltip, even if the recipe existed.


williewillus:
Target 1.19 please.

Aegide (commented):
Done

PR #4139, opened 2022-08-28
Update drop tooltip for mana infusion

Aegide:
I didnt touch at "dropTip1" in case it was already translated (like in spanish)


Xplat/src/main/java/vazkii/botania/client/patchouli/processor/ManaInfusionProcessor.java
```diff
@@ -84,7 +84,8 @@ public IVariable process(String key) {
 			case "drop":
 				Component q = Component.literal("(?)").withStyle(ChatFormatting.BOLD);
 				return IVariable.from(Component.translatable("botaniamisc.drop").append(" ").append(q));
-			case "dropTip0":
+			case "dropTip2":
+			case "dropTip3":
 			case "dropTip1":
 				Component drop = Component.keybind("key.drop").withStyle(ChatFormatting.GREEN);
 				return IVariable.from(Component.translatable("botaniamisc." + key, drop));
```
williewillus, line 90: is it okay that you're passing a format arg even when dropTip3 doesn't need one?

Aegide: true, true

PR #4145, opened 2022-08-31
[1.19] Update lenses

Aegide:
We know it's a mana lens, it would be redundant to re-specify it.

<br>

image

Now the item names match the entries.

<br>

On top of that, it would be possible to change the "text" of the "crafting" page, so it's less repetitive.


williewillus:
Those crafting entries are missing nerdy references, but yeah this is fine.

PR #4146, opened 2022-08-31
[1.19] Update blackHoleTalisman entry

Aegide:
> if i right click on a chest, the chest opens :clueless:


Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -3036,7 +3036,7 @@
...
   "botania.page.blackHoleTalisman1": "Right-clicking a block with an empty Talisman will set it to store that type of block, and sneak-right clicking the Talisman in the air will enable or disable it. When enabled, a Talisman will absorb any blocks of its given type from its user's inventory.$(p)Placing the item in a crafting grid will yield its stored blocks, a stack at a time.",
-  "botania.page.blackHoleTalisman2": "Any blocks stored in a Talisman can be placed by simply right-clicking with it.$(p)Right-clicking the Talisman on an inventory will dump up to a stack of blocks into that inventory. Finally, a Talisman can provide a $(l:tools/exchange_rod)$(item)Rod of the Shifting Crust$(0)$(/l) with the blocks it contains.",
+  "botania.page.blackHoleTalisman2": "Any blocks stored in a Talisman can be placed by simply right-clicking with it.$(p)Sneak-right clicking the Talisman on an inventory will dump up to a stack of blocks into that inventory. Finally, a Talisman can provide a $(l:tools/exchange_rod)$(item)Rod of the Shifting Crust$(0)$(/l) with the blocks it contains.",
```
williewillus, line 3039: Should be `Sneak right-clicking`, not `Sneak-right clicking`

williewillus (changes requested):

Xplat/src/main/resources/assets/botania/lang/en_us.json
Aegide: image

Then why are there multiple entries that use `Sneak-right clicking` and zero entries that use `Sneak right-clicking` ?

williewillus: ok, this is fine for now then, though I think that's the wrong version to use.

Aegide: Yeah true, but I didn't want to go against the standard that was already established.

PR #4149, opened 2022-09-02
[1.19] Update Ender Air entry

Aegide:
The End is not an item.


williewillus:
No need to prefix everything with "[1.19]", by the way.

PR #4153, opened 2022-09-11
Move JEI integration to XPlat to enable it on Fabric

shartte:
While testing on Fabric with JEI, I noticed the JEI integration is currently Forge-only.

Luckily it doesn't seem to make use of any platform-specific API so moving it over to Xplat just worked (tm).


Fabric/src/main/resources/fabric.mod.json
```diff
@@ -55,6 +55,9 @@
     ],
     "rei_client": [
       "vazkii.botania.fabric.integration.rei.BotaniaREIPlugin"
+    ],
+    "jei_mod_plugin": [
+      "vazkii.botania.client.integration.jei.JEIBotaniaPlugin"
```
williewillus, line 60: is this actually necessary? I thought JEI uses annotations to identify plugins

Technici4n: The annotation is only used on Forge. Since Fabric doesn't do jar scanning for annotations, JEI uses entrypoints there.

PR #4164, opened 2022-09-24
Update redString entry

Aegide:
<br>

# Current

current_a
current_b
current_c

<br>

# New

new_a
new_b
new_c


Aegide (commented):
I didnt change any non-english translation, but with that variant it should be do-able.

PR #4167, opened 2022-10-08
Remove orphaned translation keys

Quezler:
Remove(s/d) translation entries who’s keys are no longer present in en_us.json

Includes the nodejs script i wrote & used to automate this task, since doing this by hand would be slightly insane.


Aegide:
I already removed these keys in the translation i'm working on, but it's nice that you fixed all of these translation files.

PR #4208, opened 2022-11-05
Added FORGE suffix to forge builds

### LINKED ISSUE #4169: Release forge builds with a FORGE suffix like fabric
We should release forge versions with a forge suffix to be symmetrical: for example `Botania-1.18.2-435-FORGE.jar` to match `Botania-1.18.2-435-FABRIC.jar`. Or change to something else that's consistent. (currently, forge releases look like this: `Botania-1.18.2-435.jar`)

LeoBeliik:
Fixes #4169


artemisSystem:
Thanks!

PR #4209, opened 2022-11-06
Makes Items inside corporea cubes don't go outside the model

### LINKED ISSUE #4192: Item placed too high for Corporea Crystal Cube
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-436

### Modloader version

Fabric 0.14.9

### Issue description

Perhaps the "item" could be placed lower.

image

LeoBeliik:
Closes #4192

A little explanation for this: 

`ms.translate(0F, (Mth.sin(((float)entity.getAge() + f) / 10.0F + entity.bobOffs) * 0.1F + 0.1F) / -7F, 0F);` this is the mayor change that "syncs" the items and the cube
`Mth.sin(((float)entity.getAge() + f) / 10.0F + entity.bobOffs) * 0.1F + 0.1F` is the calculation vanilla uses for the bobbing in `ItemEntityRender#render`

Using only that makes that the cube moves as fast as a normal item does, but also prevents the item inside it to move, so I divided it by a negative number (you can also multiply to make it move faster)
`-1` almost syncs the item, I used `-7` to make the cube slightly slower.

This still keeps the item and the cube in sync, the cube just doesn't move as much as the item, but the item never goes outside the cube (based on the tests I made; some edge cases could exist).


williewillus:
Can you put your explanation into the code as comment?

LeoBeliik:
yes

PR #4210, opened 2022-11-06
Makes manabursts visible after they warp

### LINKED ISSUE #3636: Mana burst is invisible after warping
# Version Information
Forge version:  1.16.4-35.1.37
Botania version:  1.16.4-411

# Further Information
Steps to reproduce:

Set up as in image, with pulse mana spreader (warp lens) aimed at force relay, force relay linked to block above stone block, and hourglass set to 1 second.

What I expected to happen:
The mana burst would teleport to above the stone block and continue until it strikes the mana pool. This occurs as expected when using a bore-warp lens instead: 

What happened instead:
The mana burst strikes the relay and seems to disappear. However the burst still exists as the mana detector still activates, as seen in first image.

LeoBeliik:
Closes #3636

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/lens/WarpLens.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/lens/WarpLens.java
@@ -11,6 +11,7 @@ package vazkii.botania.common.item.lens;
 import net.minecraft.core.BlockPos;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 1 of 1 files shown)


williewillus:
Can you explain how this fixes the problem?

LeoBeliik:
The change here is `return shouldKill;` -> `return false;` that made the burst in the client to be invisible, the format is based on the bore lens for consistency.

williewillus:
I trust you tm

PR #4212, opened 2022-11-06
Fix flower shrinking heading

Aegide:
petite

### DIFF
--- a/Xplat/src/main/resources/assets/botania/lang/en_us.json
+++ b/Xplat/src/main/resources/assets/botania/lang/en_us.json
@@ -2118,6 +2118,7 @@
   "botania.tagline.flowerShrinking": "Smaller flowers with a smaller range",
   "botania.page.flowerShrinking0": "Some functional flowers can have a bit too much reach for certain use-cases. Tossing one of those into a $(l:mana/pool)$(item)Mana Pool$(0)$(/l) with an $(l:devices/mana_alchemy)$(item)Alchemy Catalyst$(0)$(/l) will (at the cost of some $(thing)Mana$(0)) shrink it, making its area of effect smaller. Every other function remains the same.",
   "botania.page.flowerShrinking1": "Like a nendoroid",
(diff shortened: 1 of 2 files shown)


Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -2118,6 +2118,7 @@
   "botania.tagline.flowerShrinking": "Smaller flowers with a smaller range",
   "botania.page.flowerShrinking0": "Some functional flowers can have a bit too much reach for certain use-cases. Tossing one of those into a $(l:mana/pool)$(item)Mana Pool$(0)$(/l) with an $(l:devices/mana_alchemy)$(item)Alchemy Catalyst$(0)$(/l) will (at the cost of some $(thing)Mana$(0)) shrink it, making its area of effect smaller. Every other function remains the same.",
   "botania.page.flowerShrinking1": "Like a nendoroid",
+  "botania.page.flowerShrinking2": "Petite Flower",
```
artemisSystem: Maybe make the heading say "Petite Flowers" instead?

Aegide: I was also very hesitant about that choice.

image

Here for example:
- there are multiple petals (2)
- there are multiple type of petals (16)

But it's still "Mystical Petal", so i wanted to use pre-existing naming logic.

However that logic could be changed, if the rules are clearly defined.

artemisSystem: I think the "Mystical Petal" name here would be better off as "Mystical Petals" too, for both the reasons you specified. I can change that myself now, since it's in a separate file and doesn't really belong in this PR.

So to be clear, yeah change it to "Petite Flowers" please.

Aegide: Sure, done.

artemisSystem:
Thanks!

PR #4214, opened 2022-11-07
Makes mana blasters don't have recoil if player is wearing a Tectonic Girdle

LeoBeliik:
New mechanic, asked Willie and he say yes (he actually said sure)

 
willie_said_yes

This is undocumented.
Do I just need to update it the en_us.json or does Botania has a different way to generate this?

### DIFF
--- a/Xplat/src/main/resources/assets/botania/lang/en_us.json
+++ b/Xplat/src/main/resources/assets/botania/lang/en_us.json
@@ -3132,7 +3132,7 @@
 
   "botania.entry.knockbackBelt": "Tectonic Girdle",
   "botania.tagline.knockbackBelt": "Prevent all knockback",
-  "botania.page.knockbackBelt0": "By harnessing the (relative) stability of tectonic forces, the $(item)Tectonic Girdle$(0) negates any $(thing)Knockback$(0) applied to its wearer from outside attacks and all recoil from the $(item)Mana Blaster$(0).",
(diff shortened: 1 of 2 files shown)


artemisSystem:
> Do I just need to update it the en_us.json or does Botania has a different way to generate this?

To update the entry in the book, just edit `en_us.json`, yeah.

Aegide:
Nice

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -3132,7 +3132,7 @@
 
   "botania.entry.knockbackBelt": "Tectonic Girdle",
   "botania.tagline.knockbackBelt": "Prevent all knockback",
-  "botania.page.knockbackBelt0": "By harnessing the (relative) stability of tectonic forces, the $(item)Tectonic Girdle$(0) negates any $(thing)Knockback$(0) applied to its wearer from outside attacks.",
+  "botania.page.knockbackBelt0": "By harnessing the (relative) stability of tectonic forces, the $(item)Tectonic Girdle$(0) negates any $(thing)Knockback$(0) applied to its wearer from outside attacks and all recoil from the $(item)Mana Blaster$(0).",
```
williewillus: link to the entry?

LeoBeliik: you'll need to be more specific on what you mean by this :p

williewillus: lexicon entries can link to other ones, and we try to make it so that items that reference other items have a clickable link to that entry. I was asking if you could add a link to the Mana Blaster entry.
Should be something like `$(l:botania:mana_blaster)Mana Blaster$(/l)`, or whatever the entry ID is.

LeoBeliik: ah, I see, yes I'll do that

PR #4216, opened 2022-11-11
Update banner pattern icon

Aegide:
image


artemisSystem:
Hmm, it seems like you're using an arbitrary one of the vanilla banner patterns. Which ones exist, and do they all just have that texture?

Not sure if there's a better solution if we want to show that texture, but i'd like to know what the names of the other patterns are first, i think

Aegide:
Yes they do all have that texture.

image

<br>

> "i'd like to know what the names of the other patterns are"

Sure, here is a detailed list.

https://minecraft.fandom.com/wiki/Banner_Pattern

artemisSystem:
Ok, i think using the flower one here is fine then. Thanks!

PR #4217, opened 2022-11-11
Fix advancements

Aegide:
image
image


Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -398,7 +398,7 @@
   "advancement.botania:flowerPickup": "Sky of Beginning",
   "advancement.botania:flowerPickup.desc": "Pick up a Mystical Flower from the world",
   "advancement.botania:lexiconUse": "A Realistical Logical Ideologist",
-  "advancement.botania:lexiconUse.desc": "Open a Lexica Botania and have a look through the _index",
```
williewillus, line 401: pretty sure this was an intended reference at some point in the past

Aegide: I have no idea.

I just know that this message appears in the advancement menu, which is not the most interactive place when it comes to "macro".

PR #4218, opened 2022-11-12
Added small cooldown to pickaxes when they place torches

LeoBeliik:
Pickaxes has now a small cooldown when used to play torches.
This also fixes picks not letting use shields in the offhand when there was torches in the inventory.
The cooldown is the minimum amount of time required to trigger the shield (or other off hand non block item) use.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/equipment/tool/manasteel/ManasteelPickaxeItem.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/equipment/tool/manasteel/ManasteelPickaxeItem.java
@@ -40,6 +40,8 @@ public class ManasteelPickaxeItem extends PickaxeItem implements CustomDamageIte
 
 	private static final int MANA_PER_DAMAGE = 60;
 
+	private static final int TIME = 5;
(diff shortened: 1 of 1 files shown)


Xplat/src/main/java/vazkii/botania/common/item/equipment/tool/manasteel/ManasteelPickaxeItem.java
```diff
@@ -72,11 +74,13 @@ public InteractionResult useOn(UseOnContext ctx) {
 					if (did.consumesAction() && !ctx.getLevel().isClientSide) {
 						ItemsRemainingRenderHandler.send(player, displayStack, TORCH_PATTERN);
 					}
-					return did;
+					if (did != InteractionResult.FAIL) {
```
williewillus: why not just did.isSuccess()? (or whatever it's called)

LeoBeliik: I'm small brain

PR #4219, opened 2022-11-13
Makes warped entropic burst not explode immediately after being warped.

LeoBeliik:
manaburst from warp + entropic lenses explode immediately after being warped instead of continuing like all other warped bursts do, this changes that.
~~The thing to be consider is that this could break any farms that rely on this behaviour, although it's simply solvable by using a piece of obsidian to make the burst explode.~~
I've been informed that this is indeed a bug.


Xplat/src/main/java/vazkii/botania/common/item/lens/EntropicLens.java
```diff
@@ -8,27 +8,35 @@
...
+
 			if (!entity.level.isClientSide && !burst.isFake() && !isManaBlock) {
 				entity.level.explode(entity, entity.getX(), entity.getY(), entity.getZ(),
 						burst.getMana() / 50F, Explosion.BlockInteraction.BREAK);
 			}
 			return true;
 		}
 
-		return shouldKill;
+		return false;
```
artemisSystem: Why did you change this line?

LeoBeliik: `return shouldKill;` makes the warped burst invisible in the client side, 
https://github.com/VazkiiMods/Botania/pull/4210#issuecomment-1304808496

williewillus:
Please add a test

Xplat/src/main/java/vazkii/botania/common/item/lens/EntropicLens.java
artemisSystem: i feel like there's a deeper issue here. i suspect the function gets passed a false on the server and a true on the client...

LeoBeliik:
Sorry I don't know how to do that

Xplat/src/main/java/vazkii/botania/common/item/lens/EntropicLens.java
LeoBeliik: you're correct

artemisSystem:
If you wanna have a go anyway, see [here](https://github.com/VazkiiMods/Botania/blob/1.19.x/CONTRIBUTING.md#working-with-gametest) for an introduction, feel free to ask for help on discord (for example in #modders) also.

I imagine a test for this would look something like: Fire a burst (using pulse spreader. you can place a button on it, there's a method to press a button in a test, see InterceptorTest), then after it's warped (i think there's a method that lets you wait until a certain thing is true, you can probably check the burst entity's warped property, i think it has something like that), assert that it still exists at the expected position, then kill it so the explosion doesn't make a mess

Xplat/src/main/java/vazkii/botania/test/item/lens/EntropicWarpLensTest.java
```diff
@@ -0,0 +1,40 @@
...
+	public void testWarp(GameTestHelper helper) {
+		var spreaderPos = new BlockPos(3, 3, 3);
+		var lampPos = new BlockPos(2, 4, 3);
+		var buttonPos = new BlockPos(4, 3, 3);
+		var relayPos = new BlockPos(3, 2, 3);
+		var bindPos = new BlockPos(3, 10, 3);
+		var player = helper.makeMockPlayer();
+
+		var data = ForceRelayBlock.WorldData.get(helper.getLevel());
+		data.mapping.put(helper.absolutePos(relayPos), helper.absolutePos(bindPos));
```
artemisSystem, line 27: you could probably do the binding ahead of time, instead of doing it at test run time. unless the binding doesn't work because of position stuff?

LeoBeliik: I copied it from the relay test, dunno how to do the binding ahead of time as you say

artemisSystem: You would just do it after building the structure, before saving it. But i don't know if the coordinates behave or not. You can try it if you want, but if another test does it like this then it's probably not a big deal

LeoBeliik: Oh I tried that before and it didn't work, that's why I did it this way

williewillus: > unless the binding doesn't work because of position stuff?

binding data is stored in absolute coords, tests can be loaded anywhere in the world. So doing the binding at runtime is necessary.

williewillus:
Sorry for dragging on this. I took a closer look and I think there's a better solution hiding that isn't adding yet-another-special-case for the force relay. On the other hand, things are tricky.

Basically, whenever a warp lens warps, it adds the destination position to the burst's "have I collided here already" set. 

However, the problem is that checking this flag is up to each individual lens, as it's an open question whether you want a composite effect to activate immediately on the warp destination (if there's something there) or not. Explosion? maybe not. Block breaking or weight lensing, maybe? Let's not get into whether a blanket check on this flag is appropriate or not, and just assume we'll just add the check to the entropic lens.

So adding a `!burst.hasCollidedAtPos(hitResult.getBlockPos())` check should work right? No, because the `HitResult` still has the position of the old collision, and that position has not been added to the set yet (this is the right thing to do, we don't want to add it until the entire collision process is complete).

So we need some way for warp lenses to "reroute" the `HitResult` of the collision to point to the new position. This is tricky, because `collideBurst` can only return one thing, and `HitResults` are immutable. It would require enriching the result of `collideBurst` to more than just a plain `shouldKill` boolean.

Overall lenses are the most shaky part of Botania's core and changes to it are really tricky to get right.

artemisSystem:
Yeah, i agree, lenses are super shaky. There's still also that issue of the warp lens being invisible after warping that i still quite haven't pinned down. 

The shouldKill is also just in general confusing to work with, i think. Revisiting the whole thing would probably be nice but it's probably too late to get that in for 1.19.

With that said, i'm not sure what to do with this PR. merge for now and clean up lenses later? I'll leave the decision up to you willie.

williewillus:
The disappeared-after-warp thing, assuming it's a desync as you conjectured, can be sidestepped entirely by only killing the burst on the server (I have a patch up for that). With that patch, bursts seem to be visible after warping just fine.

@LeoBeliik I think for now, just add a TODO to the entropic file where you made the fix pointing back to my big comment above, and I'll revisit it later.

artemisSystem:
Yup, saw the patch after posting my comment. Will review it when i have time

LeoBeliik:
wait, there's a miss understanding here, the warp burst not rendering in the client was already fixed in #4210, or I'm being :clueless:

artemisSystem:
it was still broken for me when i tested it a couple days ago, im pretty sure i'd included that commit then

LeoBeliik:
I just tested it with various lens combinations and it does render as it should, both fabric and forge

PR #4228, opened 2022-11-22
Fix two autocrafting bugs regarding Crafty Crate & Black Hole Talisman

### LINKED ISSUE #4227: Black hole talisman when attempting to extract using a crafty craft while empty deletes the talisman instead of dropping as is
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-436-FABRIC-SNAPSHOT

### Modloader version

Quilt: Loader v0.18.1-beta.10, Quilted Fabric API 4.0.0-beta.19+0.64.0-1.19.2

### Modpack info

https://github.com/Oliver-makes-code/gaycraft-pack

### The latest.log file

https://gist.github.com/KJP12/50795617495920bf3cbaa0a9cc2bebee

### Issue description

When sending an empty black hole talisman through a crafty crate, instead of dropping it or holding onto it as expected, it instead deletes the talisman entirely.

### Steps to reproduce

1. Obtain talisman, crafty crate and hopper
2. Place crafty crate and feed hopper into it.
   - Optionally give it a 1x1 placeholder.
3. Give the crafty crate the talisman via the hopper.
4. If not using the placeholder, right click with a wand of the forest.

Note: The talisman has to be empty for the bug to show, it'll otherwise extract the contents and drop itself as normal

### Other information

Reproduced with a minimal set on the client of just QSL, Botania, Patchouli, LazyDFU and Trinkets, happens on server as well.

Ampflower:
Two bugs that together caused an item deletion glitch has been fixed.

 - Fixed Crafty Crate allowing an empty result on a matched recipe to continue crafting, which vanilla would not allow.
 - Fixed Black Hole Talisman matching when empty, then returning air for all, allowing Crafty Crate and other non-compliant crafting interfaces to delete it entirely.

Fixes #4227

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CraftyCrateBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CraftyCrateBlockEntity.java
@@ -170,6 +170,14 @@ public class CraftyCrateBlockEntity extends OpenCrateBlockEntity implements Wand
 		matchingRecipe.ifPresent(recipe -> {
 			craftResult = recipe.assemble(craft);
 
+			// Given some mods can return air by a bad implementation of their recipe handler,
+			// check for air before continuting on.
+			if (craftResult.isEmpty()) {
+				// We have air, do not continue.
+				matchFailed = true;
+				return;
+			}
+
 			Container handler = getItemHandler();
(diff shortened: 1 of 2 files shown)


williewillus:
Thanks!

PR #4229, opened 2022-11-23
Makes TerraShatterer only activate when sneak + right click in the air

LeoBeliik:
TerraShatterer only activate when sneak + right click in the air making blocks and items in the off hand usable.


williewillus:
Applied manually as 4c2251b96aba4f7fd8c18328ba2e21141ff9608d

PR #4249, opened 2022-12-20
Corporea string item loading is done client-side

### LINKED ISSUE #4248: Corporea index doesn't localize
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-436-FABRIC

### Modloader version

Quilt Loader 0.17.7 / Quilted Fabric API 4.0.0-beta.24+0.68.0-1.19.2

### Modpack info

Spooled (by Moss)

### The latest.log file

(there is no crash)

### Issue description

Corporea indices can't request anything from the network with their localized name.

Given this setup (there's 60 obsidian in the barrel):

- Requesting `obsidian` won't find any
- Requesting `block.minecraft.obsidian` will get you 1 obsidian
- Same with requesting `*`

### Steps to reproduce

1. Make a minimal corporea network with a storage chest and an index
2. Try to request something
3. Find it only works with the I18n string, not the translation

### Other information

I'm sorry, Artemis, for being mean to you.

gamma-delta:
This closes #4248 and also has some other nice effects:
- We no longer load the I18n on the server (which shouldn't be happening anyways, strictly speaking)
- Players in different languages can now request items in their language (tested, but the special-cased `stack of` and such are still Englishy)
- Requesting `this` now uses a `CorporeaItemStackMatcher` instead of searching for the name of the item

I've done some tests and the following works:
- Requesting items in various different languages
- Intercepting foreign-language requests
- Maintaining foreign language requests across lang changes:
  - Request `64 stone` with nothing in the system
  - Have an interceptor get it
  - Switch language to Toki Pona
  - Put stone in the system and replay the request with the retainer, and it arrives
  - Do the same in reverse; requesting `64 kiwen` in Toki Pona gives you the stone after switching back to English

Todo:
- [x] Check you can request items with custom names
- [x] Check keybind requests still work
- [ ] Check that intercepted requests from old versions still work when loaded (I made some changes to the storage format and added a legacy check to still try its best; is this a goal we need to test?)


gamma-delta:
whoops I very cleverly put backticks in the last commit name so it tried to interpret them as commands. it should say "make matching `this` use NBT in the same way the `c` keybind does"

gamma-delta:
Say I have an iron ingot in the system called `Foobar`. If I request an `iron ingot`, is it expected behavior to recieve the renamed ingot?

(I'll run spotless at the end)

gamma-delta:
One problem I thought of: it won't work nicely with items that have a different display name per stack irregardless of the display name, like hexcasting scrolls...

artemisSystem:
> Check that intercepted requests from old versions still work when loaded (I made some changes to the storage format and added a legacy check to still try its best; is this a goal we need to test?)

Yes, i'd say this is a goal. Currently not feeling very well but if i get better in the next few days i can test whether it works currently

> One problem I thought of: it won't work nicely with items that have a different display name per stack irregardless of the display name, like hexcasting scrolls...

Can you rephrase? I don't get what you mean

> Say I have an iron ingot in the system called Foobar. If I request an iron ingot, is it expected behavior to recieve the renamed ingot?

I think you already got the answer to this but, for posterity: No, it shouldn't return the renamed item.

Alwinfy:
> > One problem I thought of: it won't work nicely with items that have a different display name per stack irregardless of the display name, like hexcasting scrolls...
> 
> Can you rephrase? I don't get what you mean

Consider an item that changes its display name by its NBT, such as a fluid tank that may be named "Empty Tank" or "Tank of {X liquid}"

TheRealWormbo:
Can anything from this PR be salvaged?

chuangyuan1:
Even today, in all non-English language servers, Corporea index still can't recognize localized item names. It seems like this problem will take a long time to fix, or maybe it has been permanently put on hold   


 XD

TheRealWormbo:
It's not a particularly simple change, and it will very likely not happen in 1.20.1 anymore.

PR #4283, opened 2023-02-03
Full translation into Spanish (update and reworked)

Bksp:
More than 4 hours, with the help of the AI ​​to be able to make him respect all his work, which by the way, I love


artemisSystem:
This does not seem very looked through, and a machine translation is not going to be very good anyway. For example, it's translated the roman numeral "I" to "Yo", and I spotted some inconsistencies from just briefly looking through. Sorry, but I'm not going to merge this

PR #4299, opened 2023-02-20
Add gaia arena check to api

justliliandev:
This PR adds 2 methods to the api to check if something is currently inside a gaia arena


williewillus:
Thanks. Can you do the following:

1. Check the lint/format errors
2. Instead of duplicating the logic, change the Botania Doppleganger class to call the API method instead.

Xplat/src/main/java/vazkii/botania/api/BotaniaAPI.java
```diff
@@ -220,4 +221,12 @@ default void sparkleFX(Level world, double x, double y, double z, float r, float
 	default void registerCorporeaNodeDetector(CorporeaNodeDetector detector) {
 
 	}
+
+	default boolean isInGaiaArena(Entity entity) {
```
williewillus: I would omit this to keep the api lean. Callers can write the get calls themselves.

Xplat/src/main/java/vazkii/botania/api/BotaniaAPI.java
```diff
@@ -220,4 +221,12 @@ default void sparkleFX(Level world, double x, double y, double z, float r, float
...
+		return isInGaiaArena(entity.getLevel(), entity.getX(), entity.getY(), entity.getZ());
+	}
+
+	default boolean isInGaiaArena(@Nullable Level level, double x, double y, double z) {
```
williewillus: Why Nullable, let's just enforce nonnull

justliliandev:
I wouldn't recommend doing that, as the gaia logic would be as follows:
1. Gaia tick
2. Gaia searches for close Players 
3. API check if player is inside arena by looking for nearby gaia guardian entities
4. API check if the player is inside the arena of that guardian

this would result in the guardian looking essentially for itself to disable players.
I've changed the api to use the arena AABB which is used in the player lookup to do the arena check, as this is what is used for looking for nearby players and disabling their flight/sleeping/removing effects

TheRealWormbo:
We should probably incorporate this change in the 1.21 port.

Issue #4287, opened 2023-02-08
Suggestions on the balance of the Entropinnyum


kangzaig:
### Mod Loader

Forge

### Minecraft Version

1.12.2-1.19

### Botania version

1.12.2-1.19

### Modloader version

forge

### The latest.log file

suggest

### Issue description

I tried to publish these problems in discord at the beginning,But it doesn't seem to have any result.My original words on discord:
"kangzaig — 今天09:04
First of all, thank you for your interesting mod, but there are some things to say. In China, Botania was called "dye magic" and "magic in science and technology" when it was first introduced into China, but now it is the consensus of all people to call it "tnt magic". I personally think the colorful flower integrated production magic I made is a pioneering invention. But everyone disagrees, and therefore accuses me of not putting forward this useless method of producing magic. Everyone thinks shulker flowers and hot popcorn are good. My tutorial is useless, and those who want to try new methods also give up, because tnt is better. The measures taken by your Botania team to prevent the use of tnt in the higher version are useless. Some integration packages have deleted the craft of tnt flowers and used the magic of throwing tnt into the pool. But most integration packages have changed the magic consumption to very high because of the simple magic of tnt production, and the magic production of shuiker flowers is too large. With the help of various mods, it is very easy to make a large number of shulker sprayers to produce shulker, The problem is that the shulker production mana is far higher than the cost, which are too simple and powerful. If you can't change the tnt flower, why not delete the tnt flower directly? It's also good to add new magic flowers. The tutorial I made is useless because no one will see it. On the contrary, the people who watch the tnt flower tutorial without slime block brush tnt are very supportive. In this regard, if I do not make effective changes to the tnt flower, I can only delete my video, and I will not make any tutorials about the magic of Botania production from now on. https://www.bilibili.com/video/BV1zD4y1N7e6
I have to say
Hot popcorn is tnt flower.
Can you see this? My English level still needs to be improved
Thank you.
I really hope to pay attention to this serious problem
I don't want to ruin Botania's interest because of tnt flowers
But tnt is a game vulnerability, which almost destroys the interest of botania
This is too pessimistic. It's even better to take unlimited mana from your backpack
Wrong, everyone cheats, this is not cheating, it is legal.
[09:33]
There should be more reasonable and interesting ways than tnt flower.
This repair can be said to have little effect. Since we want to change this situation, we should have a thorough solution.
Like most people who play plant magic, he may not understand red stone at all, but he may be very familiar with tnt without mucus.
[...]

PR #4307, opened 2023-03-04
Improve detection of unethical TNT

TheRealWormbo:
This change replaces Botania's unreliable TNT duplication detection (moving block + detector rail + slime/honey block = unethical TNT) to cover all piston-based TNT duplication methods, including approaches based on dust redirection and any other methods that don't use coral blocks. Block structures moved by a force relay or a mana burst with the force lens are also covered now.

The new detection method works as follows:
- When a piston is about to start moving blocks, a tracking logic for spawned TNT entities starts.
- Whenever a TNT entity is spawned while tracking is active, the entity is added to a temporary list.
- When the piston finished converting all blocks it wants to move into moving block entities, the tracking logic stops and all potentially spawned TNT entities are validated. If a moving block entity is found next to the TNT entity (moving blocks are created at the destination position) that represents a TNT block moving away from the TNT, the TNT is assumed to have been duplicated and flagged accordingly.
- TNT entities spawned outside the tracking period are considered "ethical", since they cannot have originated from piston-based TNT duplication.

This PR was inspired by issue #4287, but doesn't implement any of the feature changes suggested there.

As far as I can tell, my changes are compatible to the piston behavior modifications by both Quark and Carpet mod, and hopefully any other mods that modify vanilla piston behavior in a sensible manner. Block structure movement through other means than vanilla pistons or the force lens logic are not covered, but chances are they weren't really covered before this change either.


kangzaig:
> 这一变化取代了Botania不可靠的TNT重复检测（移动块+探测器轨道+粘液/蜂蜜块=不道德的TNT），以涵盖所有基于活塞的TNT复制方法，包括基于灰尘重定向的方法和任何其他不使用珊瑚块的方法。由力继电器移动的方块结构或用力透镜的法力爆发现在也被覆盖。
> 
> 新的检测方法的工作原理如下：
> 
> * 当活塞即将开始移动方块时，生成的TNT实体的跟踪逻辑将启动。
> * 每当在跟踪处于活动状态时生成 TNT 实体时，该实体都会添加到临时列表中。
> * 当活塞完成将其想要移动到移动块实体的所有块转换后，跟踪逻辑将停止，并且所有可能生成的TNT实体都将得到验证。如果在TNT实体旁边找到一个移动块实体（在目标位置创建移动块），该实体代表TNT块远离TNT，则假定TNT已被复制并相应地标记。
> * 在追踪期之外产生的TNT实体被认为是“道德的”，因为它们不可能来自基于活塞的TNT复制。
> 
> 此 PR 的灵感来自问题 #4287，但没有实现那里建议的任何功能更改。
> 
> 据我所知，我的更改与夸克和地毯模组的活塞行为修改兼容，并希望任何其他以合理方式修改原版活塞行为的模组兼容。通过原版活塞或力透镜逻辑以外的其他方式进行的块结构运动没有被涵盖，但很可能在此更改之前也没有真正涵盖它们。

As far as I know, some detection methods may be similar to those in the article, but they all have detection scope. When the tnt replication is far enough from the detection scope, making the tnt entity enter the work scope of the flower will still cause the replicated tnt entity to generate mana

Alwinfy:
Oh no, our detection method is global (triggers every time a piston pushes something anywhere) and we don't have any checks for flower range. That's not a problem.

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/EntropinnyumBlockEntity.java
```diff
@@ -29,54 +27,77 @@
...
+import java.util.ArrayList;
 import java.util.List;
 
 public class EntropinnyumBlockEntity extends GeneratingFlowerBlockEntity {
 	private static final int RANGE = 12;
 	private static final int EXPLODE_EFFECT_EVENT = 0;
 	private static final int ANGRY_EFFECT_EVENT = 1;
 
+	private static boolean trackTntEntities = false;
+	private static final List<PrimedTnt> trackedTntEntities = new ArrayList<>();
```
Alwinfy: Do we want to wrap this in a [ThreadLocal](https://docs.oracle.com/en/java/javase/18/docs/api/java.base/java/lang/ThreadLocal.html) to be defensive?

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/EntropinnyumBlockEntity.java
```diff
@@ -29,54 +27,77 @@
...
-	public static boolean isUnethical(Entity e) {
-		BlockPos center = e.blockPosition();
-		if (!e.level.isLoaded(center)) {
-			return false;
+	/**
+	 * A force relay, force lens mana burst or vanilla piston is about to start moving blocks.
+	 */
+	public static void startTrackingTntEntities() {
+		assert (!trackTntEntities && trackedTntEntities.isEmpty());
+		trackTntEntities = true;
```
Alwinfy: Similarly, should this be a counter (in case some crazy mixin makes moveBlocks reentrant)?

kangzaig:
> 哦，不，我们的检测方法是全局的（每次活塞将东西推到任何地方时都会触发），我们没有任何花范围检查。这不是问题。

That would be very good,another small suggestion is that you can also make the tnt box ignite automatically to realize the generation of mana.

kangzaig:
> 哦，不，我们的检测方法是全局的（每次活塞将东西推到任何地方时都会触发），我们没有任何花范围检查。这不是问题。

That would be very good. Another small suggestion is that you can also make the tnt box ignite automatically to realize the generation of mana

Alwinfy (commented):
Looks pretty good, but see comments:

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/EntropinnyumBlockEntity.java
kangzaig: > 

Yes, if the tnt block is activated after updating, a tnt entity will be generated when the tnt block is pushed. For the case of piston driven tnt, this method should be feasible. When the tnt is copied, it will become a 36 block. Perhaps you can simplify the detection idea from here?

kangzaig:
The power lens also has the function of pushing the block. I mean, if you don't use the piston, you may bypass it.

kangzaig:
> 电动镜头还具有推动块的功能。我的意思是，如果你不使用活塞，你可能会绕过它。

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/EntropinnyumBlockEntity.java
williewillus: I'm always wary of storing entities in global state like this. It makes memory leaks easy to slip in. How does `WeakHashMap<Level, IntOpenHashSet>` holding the entity id's sound, instead?

TheRealWormbo: `ThreadLocal` sounds like a good idea in case dimensions' tick loop run in parallel, and I'll look into referencing entities by ID instead.

TheRealWormbo:
> The power lens also has the function of pushing the block. I mean, if you don't use the piston, you may bypass it.

Yes, this PR also covers the Force lens for mana spreaders and pushing blocks via Force Relay block.

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/EntropinnyumBlockEntity.java
TheRealWormbo: I can see the entire thing falling apart in a bad way if someone really goes that deep, but it's a counter now.

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/EntropinnyumBlockEntity.java
TheRealWormbo: Both done. I also moved the entire logic into its own helper class, because it got somewhat complex and felt messy.

Alwinfy (approved):

williewillus:
Also, this is not blocking, but it'd be great if you could write a test for this

artemisSystem (changes requested):
Some of the checking goes a bit above my head, but the code looks fine otherwise. Am I understanding correctly that the PistonBaseBlockMixin also accounts for the force relay's movements?

Also, I agree with willie. I want to have some tests for this. (which is the reason I marked this as "changes requested")

TheRealWormbo:
> Am I understanding correctly that the PistonBaseBlockMixin also accounts for the force relay's movements?

The force relay uses `ForceLens.moveBlocks`, not the vanilla piston's logic. The mixin only applies to the piston implementation. However, your question raises an interesting point – a moving force relay might actually be a reentrant case, as `ForceLens.moveBlocks` is called in `ForceRelayBlock.onRemove`, which is likely called from within the piston push logic. I'll have to investigate that.

> Also, this is not blocking, but it'd be great if you could write a test for this

I assume you mean GameTests? I'll see what reasonable scenarios I can come up with.

artemisSystem:
> I assume you mean GameTests? I'll see what reasonable scenarios I can come up with.

Yup. What i had in mind was just setup some tnt duping setups and confirm that they are caught

TheRealWormbo:
There are now game tests for the three major duping variants the detection covers:
- an "entityless" vanilla duper that drops TNT onto a platform
- a slime/detector rail duper that is moved by two force lens mana bursts (first one assembles the duper, second burst causes a duped TNT entity to drop onto the test platform)
- another slime/detector rail duper that is moved twice by a force relay (first push to assemble, second to shoot forward a TNT entity into a catching structure)
- for comparison there's also a fourth setup that just ignites a TNT block with a button, which (unlike the first three setups) produces the full amount of TNT from the Entropinnyum used to test the "legitimacy" of the TNT entity

kangzaig:
> This change replaces Botania's unreliable TNT duplication detection (moving block + detector rail + slime/honey block = unethical TNT) to cover all piston-based TNT duplication methods, including approaches based on dust redirection and any other methods that don't use coral blocks. Block structures moved by a force relay or a mana burst with the force lens are also covered now.
> 
> The new detection method works as follows:
> 
> * When a piston is about to start moving blocks, a tracking logic for spawned TNT entities starts.
> * Whenever a TNT entity is spawned while tracking is active, the entity is added to a temporary list.
> * When the piston finished converting all blocks it wants to move into moving block entities, the tracking logic stops and all potentially spawned TNT entities are validated. If a moving block entity is found next to the TNT entity (moving blocks are created at the destination position) that represents a TNT block moving away from the TNT, the TNT is assumed to have been duplicated and flagged accordingly.
> * TNT entities spawned outside the tracking period are considered "ethical", since they cannot have originated from piston-based TNT duplication.
> 
> This PR was inspired by issue #4287, but doesn't implement any of the feature changes suggested there.
> 
> As far as I can tell, my changes are compatible to the piston behavior modifications by both Quark and Carpet mod, and hopefully any other mods that modify vanilla piston behavior in a sensible manner. Block structure movement through other means than vanilla pistons or the force lens logic are not covered, but chances are they weren't really covered before this change either.

There is another solution that may be considered, which is based on the ignition lens to ignite tnt to produce magic instead of ordinary tnt.

kangzaig:
> 

Kindle lens，I mean, if we can't find a perfect solution to the problem of replicating tnt to generate mana, perhaps we can start from the way of generation.

Xplat/src/main/java/vazkii/botania/common/helper/EthicalTntHelper.java
```diff
@@ -0,0 +1,129 @@
...
+
+		BlockPos.MutableBlockPos blockPos = new BlockPos.MutableBlockPos();
+		for (final var dir : Direction.values()) {
+			blockPos.setWithOffset(center, dir);
+			if (!entity.getLevel().isLoaded(blockPos)) {
+				continue;
+			}
+
+			final var blockState = entity.getLevel().getBlockState(blockPos);
+			if (blockState.getBlock() == Blocks.MOVING_PISTON && blockState.hasBlockEntity()) {
```
williewillus: .is() is the convention over ==, and the hasBlockEntity check will always be true

```suggestion
			if (blockState.is(Blocks.MOVING_PISTON)) {
```

Xplat/src/main/java/vazkii/botania/common/internal_caps/EthicalComponent.java
```diff
@@ -11,20 +11,24 @@
...
-import vazkii.botania.common.block.flower.generating.EntropinnyumBlockEntity;
+import vazkii.botania.common.helper.EthicalTntHelper;
 
 public class EthicalComponent extends SerializableComponent {
 	protected static final String TAG_UNETHICAL = "botania:unethical";
 	protected boolean unethical;
 
 	public EthicalComponent(PrimedTnt entity) {
-		unethical = EntropinnyumBlockEntity.isUnethical(entity);
+		EthicalTntHelper.addTrackedTntEntity(entity);
```
williewillus: this should probably only call the tracker on the server (!level.isClientSide)

williewillus (changes requested):
almost there, a couple more things

TheRealWormbo:
> almost there, a couple more things

Both done.

Issue #4062, opened 2022-06-05
Corporea spark on red stringed container detaches unexpectedly


TheRealWormbo:
### Mod Loader

Forge

### Minecraft Version

1.18.2

### Botania version

1.18.2-432

### Modloader version

Forge: 40.1.25

### Modpack info

https://www.curseforge.com/minecraft/modpacks/the-phoenixs-magics-and-goodies

### The latest.log file

https://gist.github.com/TheRealWormbo/99aba996a806b06e3df7798a1d1c7c31

### Issue description

Corporea sparks cannot be attached to a red stringed container block while that block is not linked to anything, and will pop off on various unexpected occasions if there actually is a link to an inventory block.

### Steps to reproduce

1. Place a sideway facing Red Stringed Container without any inventories in range.
2. Attempt to place a corporea spark on the container. (expected: spark attaches properly, exposing nothing to the corporea network; observed: unable to do so)
3. Place a shulker box in range of the red stringed container. (should link properly, as confirmed with a wand of the forest)
4. Attempt to place a corporea spark on the container. (works now)
5. Save and Quit to Title, then reopen the world. (expected: same state as after step 4; observed: spark popped off on reload)
6. Place the corporea spark on the red stringed container again.
7. Place a hopper pointing into the shulker box on the opposite side (from the perspective of the red stringed container, the hopper is behind the shulker box)
8. Break the shulker box in some way, e.g. with a piston, a bore lens mana burst, or pickaxe. (expected: corporea spark stays attached, now exposing the content of the hopper to the network; observed: spark pops off)

### Other information

This might also be a regression from #2790.

williewillus:
@Hubry can you take a look when you can

artemisSystem:
Solution might be as simple as adding red stringed container (and dispenser) to the corporea spark override tag?

TheRealWormbo:
As far as I can tell, previous versions (here: Botania for 1.12.2) ensured to always return an item handler capability, which would represent an empty, non-insertable inventory if the redstringed container is not bound to anything. I assume that behavior was lost in some kind of refactoring.
I feel like adding the block to the corporea spark override tag might intermittently report "not an inventory" to other blocks or entities interacting with it, which could cause weird interactions with other mods.

artemisSystem:
Good point, i agree the old behavior of reporting an empty inventory would be the best course of action.

TheRealWormbo:
[...]

PR #4308, opened 2023-03-05
Ensure red-stringed containers always report an accessible inventory

TheRealWormbo:
Fix for Botania issue #4062, as suggested.

For Forge, the red stringed container reports an empty inventory even when it's not bound to an inventory.

For Fabric, this was already the case, but the inventory reported that it absolutely never supports extraction, which is what Corporea sparks check for the up direction. Since even vanilla Minecraft supports dynamically adding and removing inventory blocks (shulker boxes) in the line of sight of the red stringed container, the block's inventory access options could change at any time and cause e.g. some kind of pipe-like connection to break. The obvious downside is that a player could make pipe-like connections or place a Corporea spark on a red stringed container without the bound inventory actually allowing access from that side.

### DIFF
--- a/Fabric/src/main/java/vazkii/botania/fabric/internal_caps/RedStringContainerStorage.java
+++ b/Fabric/src/main/java/vazkii/botania/fabric/internal_caps/RedStringContainerStorage.java
@@ -66,7 +66,8 @@ public class RedStringContainerStorage implements Storage<ItemVariant> {
 
 	@Override
 	public boolean supportsInsertion() {
-		return getStorage().supportsInsertion();
+		// Since the binding target could change we can't know if "absolutely always 0" is correct.
+		return true;
 	}
 
 	@Override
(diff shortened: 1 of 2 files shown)


Fabric/src/main/java/vazkii/botania/fabric/internal_caps/RedStringContainerStorage.java
```diff
@@ -66,7 +66,8 @@ public Iterator<StorageView<ItemVariant>> iterator() {
 
 	@Override
 	public boolean supportsInsertion() {
-		return getStorage().supportsInsertion();
+		// Since the binding target could change we can't know if "absolutely always 0" is correct.
```
williewillus, line 69: I'm not sure what this comment means by "absolutely always 0"

TheRealWormbo: From the description in the Storage interface:

> Return false if calling `insert` will absolutely always return 0, or true otherwise or in doubt.
> Note: This function is meant to be used by pipes or other devices that can transfer resources to know if they should interact with this storage at all.

Basically, `supportsInsertion` and `supportsExtraction` determine whether there could ever possibly be any interaction with the storage. My reasoning is that for a red stringed container this assessment is not fixed, as the binding target could change at any time. Unlike e.g. a Thermal Expansion machine, where the player could cause that through configuration changes, the red stringed container can vary in functionality without player interaction (or, turns out, even when just loading the world), so the two "supports" methods would need to pick the "in doubt" option.

williewillus: Got it, thanks

PR #4311, opened 2023-03-11
Update for Entropinnyum documentation

TheRealWormbo:
Removed the hint about avoiding rails, sticky blocks and pistons from language files that contained the paragraph. Randomly having those blocks nearby should be completely fine now.


artemisSystem:
Thanks!

PR #4313, opened 2023-03-17
Loonium Structure Feature Implementation

zacharybarbanell:
Added a feature where the Loonium generates different loot when it's placed inside different structures.  Currently only partially configurable by datapack.


Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -50,7 +53,6 @@
 public class LooniumBlockEntity extends FunctionalFlowerBlockEntity {
 	private static final int COST = 35000;
 	private static final int RANGE = 5;
-	private static final String TAG_LOOT_TABLE = "lootTable";
```
williewillus, line 53: Removing this is not ideal as it was added to allow mapmakers to override the loonium's loot table.
Unfortunately, all existing-placed flowers will have the dungeon loot table set here too.

How about we:
1. Rename the `lootTable` field to `customLootTable`, make it @Nullable, and default it to `null` for new flowers (don't write it to NBT in that case). Keep the tag key the same though for compatibility.
2. If nonnull, use this table, otherwise use the structure selection logic.
3. In the changelog (please add an entry in web/changelog.md), note that this change only applies to newly crafted looniums (old ones are going to keep their override even when broken because that's how it's supposed to be).

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -60,7 +62,49 @@ public class LooniumBlockEntity extends FunctionalFlowerBlockEntity {
...
+		}
+		if (totalWeight > 0) {
+			int roll = level.random.nextInt(totalWeight);
+			int i = 0;
+			while (true) {
+				roll -= validLoot.get(i).weight();
+				if (roll < 0) {
+					break;
+				}
+			}
```
williewillus: No need to roll your own weighted select, vanilla has utilities for it. Look around for stuff called `WeightedRandom` or similar

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -60,7 +62,49 @@ public class LooniumBlockEntity extends FunctionalFlowerBlockEntity {
...
+			return level.getServer().getLootTables().get(validLoot.get(i).loc());
+		} else {
+			return level.getServer().getLootTables().get(defaultLootTable);
+		}
```
williewillus, line 98: what happens if the table doesn't exist? Does .get() return the Empty table if the ID isn't found?

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -60,7 +62,49 @@ public class LooniumBlockEntity extends FunctionalFlowerBlockEntity {
 			Zombie.class
 	);
 
-	private ResourceLocation lootTable = new ResourceLocation("minecraft", "chests/simple_dungeon");
+	private record StructureLoot(ResourceLocation loc, int weight) {
+		public StructureLoot(ResourceLocation loc) {
+			this(loc, 1);
+		}
+	}
```
williewillus: see below, vanilla has a utility WeightedRandom.Item<T> (or something like that) that is a pair of item and int weight already, no need to remake it.

Xplat/src/main/java/vazkii/botania/common/helper/StructureHelper.java
```diff
@@ -0,0 +1,17 @@
...
+import net.minecraft.resources.ResourceLocation;
+import net.minecraft.server.level.ServerLevel;
+import net.minecraft.world.level.levelgen.structure.Structure;
+
+public class StructureHelper {
+	public static Structure getStructure(ServerLevel level, ResourceLocation key) {
+		return level.registryAccess().registryOrThrow(Registry.STRUCTURE_REGISTRY).get(key);
+	}
+
+	public static boolean isInStructureBounds(ServerLevel level, BlockPos pos, Structure structure) {
```
williewillus: @Nullable Structure structure

Xplat/src/main/java/vazkii/botania/common/helper/StructureHelper.java
```diff
@@ -0,0 +1,17 @@
...
+		return structure == null ? false : level.structureManager().getStructureAt(pos, structure).isValid();
```
williewillus:
```suggestion
		return structure != null && level.structureManager().getStructureAt(pos, structure).isValid();
```

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -2228,7 +2228,7 @@
 
   "botania.entry.loonium": "Loonium",
   "botania.tagline.loonium": "Conjures dungeon loot",
-  "botania.page.loonium0": "Any adventurer knows that $(thing)Dungeons$(0) can hold valuable goodies. The $(item)Loonium$(0) will, when fed quite a bit of $(thing)Mana$(0), summon these items for said adventurers to collect.$(p)There's a catch, though: each item is held by a monster protecting it. These monsters are extra-strong, but will drop their precious $(thing)dungeon loot$(0) when killed.",
+  "botania.page.loonium0": "Any adventurer knows that $(thing)Dungeons$(0) can hold valuable goodies. The $(item)Loonium$(0) will, when fed quite a bit of $(thing)Mana$(0), summon these items for said adventurers to collect.$(p)There's a catch, though: each item is held by a monster protecting it. These monsters are extra-strong, but will drop their precious $(thing)dungeon loot$(0) when killed.$(p)If placed somewhere with enough history, the $(item)Loonium$(0) will draw loot from its surroundings instead.",
```
williewillus: Hm, seems a bit weird for flavor text. Summoning flavor text master @Alwinfy

williewillus (changes requested):
some comments inline

Xplat/src/main/java/vazkii/botania/common/helper/StructureHelper.java
zacharybarbanell: taken straight from vanilla code, but I guess that works too

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
zacharybarbanell: Yep

Xplat/src/main/java/vazkii/botania/common/helper/StructureHelper.java
williewillus: If there's vanilla code, why not call that directly?

zacharybarbanell: The vanilla code is part of a method that's specific to nether fortresses

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
zacharybarbanell: Just to confirm - you want the nbt tag to be lootTable, not customLootTable, so that existing looniums with a set loot table retain it?

zacharybarbanell: In the currently released version of botania, if a loonium with a custom loot table set is broken, the dropped item doesn't have that NBT saved, and reverts to the normal pool if placed again - is that intentional?

zacharybarbanell:
done everything but the flavor text

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
williewillus: Huh, that's strange. Let me check.

williewillus: Ok, it doesn't, but that's a separate bug (the generated loot table needs logic to preserve the tag). As for your other question, yeah, let's keep the nbt tag as `lootTable`, but rename the code variable to express what it really means.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -59,8 +66,38 @@ public class LooniumBlockEntity extends FunctionalFlowerBlockEntity {
...
+	private static final ResourceLocation defaultLootTable = new ResourceLocation("minecraft", "chests/simple_dungeon");
+
+	//Returns an empty loot table if the selected ResourceLocation is invalid
+	private LootTable getLootTable(ServerLevel level) {
+		List<WeightedEntry.Wrapper<ResourceLocation>> validLoot = new ArrayList<WeightedEntry.Wrapper<ResourceLocation>>();
+		for (var entry : lootTables.entrySet()) {
+			ResourceLocation key = entry.getKey();
+			WeightedEntry.Wrapper<ResourceLocation> value = entry.getValue();
+			Structure structure = StructureHelper.getStructure(level, key);
+			if (StructureHelper.isInStructureBounds(level, getEffectivePos(), structure)) { //could be getBlockPos
```
williewillus: We should probably decide on one and remove the comment.

Issue #4178, opened 2022-10-24
Pushing BlockEntities with the Force lens makes them drop their contents


LeoBeliik:
### Mod Loader

Both Fabric and Forge (I confirm that I have tested both loaders and will specify both loader versions below)

### Minecraft Version

1.19.2

### Botania version

436

### Modloader version

fabric 0.14.10

### The latest.log file

nai

### Issue description

Moving block entities with the force lens makes the items inside the blocks to drop (or disappear if can't be dropped, like mana pools)
This doesn't happen using pistons

### Steps to reproduce

place a chest in front of a spreader
put the force lens in the spreader
make the spreader shoot mana
chest is moved but items are now in the floor

### Other information

I tested this both in fabric (carpet) and forge (quark) and both 1.18.2 and 1.19.2

Up to you if this is a relevant bug or not (we all know how jank moving be is)

williewillus:
I didn't really intend for this to move block entities, not sure how that's happening lol

williewillus:
oh, I see that you're using it with another mod that adds MBE. This could be because I copy the vanilla moving code which those mods mixin into, so my copy doesn't have their changes.

PR #4314, opened 2023-03-17
Reuse original piston pushing logic instead of a copy (fix for #4178)

TheRealWormbo:
This allows Botania to inherit Movable Block Entity implementations from mods like Quark or Carpet for block structures pushed by a Force Relay or by a mana burst with the Force lens behavior.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/helper/ForcePushHelper.java
+++ b/Xplat/src/main/java/vazkii/botania/common/helper/ForcePushHelper.java
@@ -1,0 +1,20 @@
+package vazkii.botania.common.helper;
+
+import org.apache.commons.lang3.mutable.MutableInt;
+
+public class ForcePushHelper implements AutoCloseable {
+	private static final ThreadLocal<MutableInt> forcePushCounter = ThreadLocal.withInitial(MutableInt::new);
+
+	public static boolean isForcePush() {
+		return forcePushCounter.get().intValue() > 0;
+	}
+
+	public ForcePushHelper() {
+		forcePushCounter.get().increment();
+	}
+
+	@Override
+	public void close() {
+		forcePushCounter.get().decrement();
+	}
+}
--- a/Xplat/src/main/java/vazkii/botania/mixin/PistonBaseBlockMixin.java
+++ b/Xplat/src/main/java/vazkii/botania/mixin/PistonBaseBlockMixin.java
@@ -8,17 +8,27 @@ import net.minecraft.world.level.block.piston.PistonBaseBlock;
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
 import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.ModifyVariable;
 import org.spongepowered.asm.mixin.injection.callback.CallbackInfoReturnable;
 
 import vazkii.botania.common.helper.EthicalTntHelper;
+import vazkii.botania.common.helper.ForcePushHelper;
 
-@Mixin(PistonBaseBlock.class)
-public class PistonBaseBlockMixin {
+/**
+ * Additional logic for pistons:
+ * <ul>
+ * <li>Detection for "unethical" (duplicated) TNT</li>
+ * <li>Hooks for Force Lens and Force Relay pushing behavior. (Reusing piston pushing code instead of copying it allows
+ * us to inherit Movable Block Entities modifications by mods like Quark or Carpet.)</li>
+ * </ul>
+ */
+@Mixin(value = PistonBaseBlock.class, priority = 999 /* before MBE implementations like Quark or Carpet */)
+public abstract class PistonBaseBlockMixin {
 	@Inject(
 		at = @At("HEAD"),
 		method = "moveBlocks(Lnet/minecraft/world/level/Level;Lnet/minecraft/core/BlockPos;Lnet/minecraft/core/Direction;Z)Z"
 	)
-	private void preMoveBlocks(Level level, BlockPos pos, Direction dir, boolean retract,
+	private void preMoveBlocks(Level level, BlockPos pos, Direction dir, boolean extending,
 			CallbackInfoReturnable<Boolean> cir) {
 		if (!level.isClientSide()) {
 			EthicalTntHelper.startTrackingTntEntities();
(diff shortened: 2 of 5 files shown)


TheRealWormbo:
The way this is supposed to work is as follows:

* The `PistonBaseBlock::moveBlocks` method is made accessible to the Force Lens code, such that it is always called with that boolean parameter (`extending`) being true.
* The copy of the `moveBlocks` code previously used by the Force Lens and Force Relay did not contain three `if` statements present in `moveBlocks`. The first of those expects the `extending` parameter to be false (and thus is skipped automatically), the other two expect it to be true. To skip those, the value of the `extending` parameter is changed to false just before the first of those `if` statements, if the method is being called in the context of a force lens/relay push. This should be safe to do, as all relevant vanilla uses of the parameter happen before that point. The same is true for its uses in the modifications applied both by Quark and Carpet mod.

Situations I can think of where this might break things:
* A mod completely replaces the `moveBlocks` method. – Both Carpet and Quark are able to apply extensive changes to piston logic without doing that, so I'd argue that hypothetical other mod is probably Doing It Wrong. This change's modifications to the method code would fail, but I think the method call should still be possible, probably leaving behind a random non-sticky piston head I'd expect to break when it gets updated.
* A mod uses a mix-in with lower priority to access the value of the `extending` parameter in a place before the second code fragment this change intends to disable. – Depending on the nature of the modifications this could cause blocks to move the wrong way, and either they end up at the correct location (just moving in from the opposite direction) or actually move the wrong way altogether.
* A mod uses a mix-in to access the value of the `extending` parameter after this logic changed it. – The mod's access would be off by one access location, which could break that mod's intended functionality. Best case here would be if it attempted to do a similar thing, which could still be compatible which what this change is trying to do.

Maybe a better way instead of modifying the actual value of `extending` would be to replace the access to its value with the expression used to set its new value now. Then again, that also has a chance of failing in a different way if another mod inserts additional code using that parameter. Those two `if` statements being skipped are responsible for creating the extending piston head block and giving adjacent blocks a shape update notification.

Xplat/src/main/java/vazkii/botania/common/helper/ForcePusher.java
```diff
@@ -0,0 +1,9 @@
+package vazkii.botania.common.helper;
+
+import net.minecraft.core.BlockPos;
+import net.minecraft.core.Direction;
+import net.minecraft.world.level.Level;
+
+public interface ForcePusher {
```
williewillus: Usually, even if we have an existing Mixin, we make a separate Accessor mixin to access internal methods. Can you just make this an Accessor instead of a custom interface that's mixed in?

Xplat/src/main/java/vazkii/botania/common/helper/ForcePushHelper.java
```diff
@@ -0,0 +1,26 @@
...
+
+public class ForcePushHelper {
+	private static final ThreadLocal<AtomicInteger> forcePushCounter = ThreadLocal.withInitial(AtomicInteger::new);
+
+	public static boolean isForcePush() {
+		return forcePushCounter.get().get() > 0;
+	}
+
+	public static ForcePushContext createForcePushContext() {
+		return new ForcePushContext();
```
williewillus: why not just make the constructor public and call it directly? This is just a needless indirection.

Xplat/src/main/java/vazkii/botania/common/helper/ForcePushHelper.java
```diff
@@ -0,0 +1,26 @@
+package vazkii.botania.common.helper;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+public class ForcePushHelper {
+	private static final ThreadLocal<AtomicInteger> forcePushCounter = ThreadLocal.withInitial(AtomicInteger::new);
```
williewillus: If this is ThreadLocal, then it doesn't need to be atomic anymore (there would be no possible race conditions). Just use `Integer`?

Xplat/src/main/java/vazkii/botania/mixin/PistonBaseBlockMixin.java
```diff
@@ -35,4 +47,24 @@ private void postMoveBlocks(Level level, BlockPos pos, Direction dir, boolean re
...
+	@Override
+	public boolean callMoveBlocks(Level level, BlockPos pistonPos, Direction direction) {
+		return moveBlocks(level, pistonPos, direction, true);
+	}
+
+	@Shadow
+	protected abstract boolean moveBlocks(Level level, BlockPos pistonPos, Direction direction, boolean extending);
+
+	/**
+	 * When force-pushing, clear the boolean method argument right before the if block that creates
```
williewillus:
```suggestion
	 * When force-pushing, set <pre>extending</pre> to false right before the if block that creates
```

williewillus (changes requested):
Couple comments

Xplat/src/main/java/vazkii/botania/common/helper/ForcePushHelper.java
TheRealWormbo: I'M not sure on this, but I think using an AtomicInteger here prevents potentially creating new Integer instances. Then again, there might be a way using Integer.valueOf to prevent that as well.

TheRealWormbo: Hmm, looks a bit ugly:
```java
	private static final ThreadLocal<Integer> forcePushCounter = ThreadLocal.withInitial(() -> 0);

	public static boolean isForcePush() {
		return forcePushCounter.get() > 0;
	}

	public ForcePushHelper() {
		forcePushCounter.set(forcePushCounter.get() + 1);
	}

	@Override
	public void close() {
		forcePushCounter.set(forcePushCounter.get() - 1);
	}
```

TheRealWormbo: Oh nevermind, MutableInt is a thing.

Xplat/src/main/java/vazkii/botania/mixin/PistonBaseBlockMixin.java
```diff
@@ -29,10 +39,19 @@ private void preMoveBlocks(Level level, BlockPos pos, Direction dir, boolean ret
...
 		if (!level.isClientSide()) {
 			EthicalTntHelper.endTrackingTntEntitiesAndCheck();
 		}
 	}
+
+	/**
+	 * When force-pushing, set the {@code extending} parameter to {@code false} right before the if block that creates
+	 * a moving piston head. This will skip all the code we don't want to execute.
+	 */
+	@ModifyVariable(method = "moveBlocks", at = @At(value = "LOAD", ordinal = 4), argsOnly = true)
```
williewillus: is there a way to further filter the At load, for example by type? The narrower you make it the less brittle this mixin is.

TheRealWormbo: Filtering by variable type happens via the parameter type of the mixin method. The `@ModifyVariable` annotation's `argsOnly` parameter narrows it down to arguments instead of all local variables. It can only be narrowed down further via `ordinal = 0` on the `@ModifyVariable` annotation itself, meaning "first boolean argument".
The `@At` annotation selects the variable access within the target method, here it's read access ("LOAD") number 4. Interestingly, `@At` counts ordinal as 1-based (0 meaning all locations), while `@ModifyVariable` counts 0-based (0 being the first variable with matching type).

williewillus:
hopingthisworks tm

PR #4316, opened 2023-03-24
Only process neighbor updates for new block when exchanging blocks with the Rod of the Shifting Crust

jpenilla:
This allows torches placed on exchanged blocks to not break for example


jpenilla:
I realized my previous solution (delaying updates with the collecting updater) was overkill and effectively the same thing as just disabling updates when removing and enabling them when placing. So I simplified the implementation to do that instead. Also passes the player into the break method for use in the break context now, which should improve the parity with drops when a player breaks the block directly. Of course I've tested the new code and it still works, I exchanged some stone with torches on it for planks and the torches persisted, reverting to the `1.19.x` branch caused them to break.

williewillus:
Wow, nice elegant solution. Thanks and sorry for the delay.

PR #4320, opened 2023-03-27
Add wand hud for mana sparks to show augments, similar to how spreaders show their lens

jpenilla:
This is initially more of a PoC to gauge interest/get feedback from maintainers and others than a proper implementation (hence draft PR).

A proper implementation would probably want to use an Entity capability similar to the WandHUD system for BlockEntities. This way it could easily be expanded to other Entity types.

Also added highlighting of the attached block like for Bound BlockEntities. But not sure if this is really needed since they will always be attached to something directly below afaik. But could be nice when you have a lot of sparks and pools next to each other.


williewillus:
Not sure about the bound block render, but if you make this just the augment HUD (properly done as a capability) I'll take it. I'd also add the color of the network in between the "Spark" text and the augment text, as a nice accessibility option.

jpenilla:
Moved it to capabilities, added color, removed the bound block outline, and tested on forge and fabric.

For some reason, checkSyntax/spotlessJavaApply refuse to run on my machine...

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
```diff
@@ -362,4 +368,44 @@ public boolean areIncomingTransfersDone() {
...
+				final Component upgradeName = upgradeItem.getHoverName();
+				int width = 16 + mc.font.width(upgradeName) / 2;
+				int x = mc.getWindow().getGuiScaledWidth() / 2 - width;
+				int y = mc.getWindow().getGuiScaledHeight() / 2 + 50;
+
+				mc.font.drawShadow(ms, upgradeName, x + 20, y + 5, color);
+				mc.getItemRenderer().renderAndDecorateItem(upgradeItem, x, y);
+			}
+
+			RenderSystem.setShaderColor(1F, 1F, 1F, 1F);
```
williewillus: is this needed? iirc none of the methods we're calling here should clobber the global color

williewillus (commented):
looks great, just one more question

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
jpenilla: no it's not, oops

williewillus:
thanks!

PR #4325, opened 2023-04-11
1.19 Orechid weights in recipe displays

TheRealWormbo:
Added back displaying Orechid and Orechid Ignem output chances to JEI, REI and EMI.
Since determining the total weight for all recipes with the same input is not possible at the time the recipe displays are built, a global cache for those sums is built after loading recipes from data packs or receiving them via the network.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/client/integration/shared/OrechidUIHelper.java
+++ b/Xplat/src/main/java/vazkii/botania/client/integration/shared/OrechidUIHelper.java
@@ -1,0 +1,169 @@
+package vazkii.botania.client.integration.shared;
+
+import it.unimi.dsi.fastutil.ints.IntIntImmutablePair;
+import it.unimi.dsi.fastutil.ints.IntIntPair;
+
+import net.minecraft.ChatFormatting;
+import net.minecraft.client.Minecraft;
+import net.minecraft.core.BlockPos;
+import net.minecraft.network.chat.Component;
+import net.minecraft.util.Mth;
+import net.minecraft.world.item.crafting.RecipeType;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+import vazkii.botania.api.recipe.OrechidRecipe;
+import vazkii.botania.common.handler.OrechidManager;
+
+import java.text.NumberFormat;
+import java.util.stream.Stream;
+
+/**
+ * Shared helper methods for the various recipe display mod plugins.
+ */
+public class OrechidUIHelper {
+	/**
+	 * How far off from the actual chance an approximated outputs/inputs ratio calculated by
+	 * {@link #getRatioForChance(double)} should be at most.
+	 */
+	private static final float MAX_ACCEPTABLE_RATIO_ERROR = 0.05f;
+
+	/**
+	 * How many input blocks an approximated outputs/inputs ratio calculated by
+	 * {@link #getRatioForChance(double)} should have at most if the
+	 * number of outputs in the ratio is greater than 1.
+	 */
+	private static final int MAX_NUM_INPUTS_FOR_RATIO = 16;
+
+	@NotNull
+	public static Component getPercentageComponent(double chance) {
+		final String chanceText = LocaleHelper.formatAsPercentage(chance, 1);
+		return Component.literal(chanceText);
+	}
+
+	@NotNull
+	public static Component getRatioTooltipComponent(@NotNull IntIntPair ratio) {
+		final NumberFormat formatter = LocaleHelper.getIntegerFormat();
+		return Component.translatable("botaniamisc.conversionRatio",
+				formatter.format(ratio.secondInt()),
+				formatter.format(ratio.firstInt()));
+	}
+
+	@NotNull
+	public static Component getBiomeChanceTooltipComponent(double chance, @NotNull String biomeTranslationKey) {
+		return Component.translatable("botaniamisc.conversionChanceBiome",
+				getPercentageComponent(chance),
+				Component.translatable(biomeTranslationKey).withStyle(ChatFormatting.ITALIC)
+		).withStyle(ChatFormatting.GRAY);
+	}
+
+	@NotNull
+	public static Stream<Component> getBiomeChanceAndRatioTooltipComponents(double chance, OrechidRecipe recipe) {
+		final var biomeTranslationKey = OrechidUIHelper.getPlayerBiomeTranslationKey();
+		final var player = Minecraft.getInstance().player;
+		if (biomeTranslationKey == null || player == null) {
+			return Stream.empty();
+		}
+
+		final var biomeChance = OrechidUIHelper.getChance(recipe, player.blockPosition());
+		if (biomeChance == null || Mth.equal(chance, biomeChance)) {
+			return Stream.empty();
+		}
+
+		final var biomeRatio = OrechidUIHelper.getRatioForChance(biomeChance);
+		return Stream.of(
+				OrechidUIHelper.getBiomeChanceTooltipComponent(biomeChance, biomeTranslationKey),
+				Component.literal("(")
+						.append(OrechidUIHelper.getRatioTooltipComponent(biomeRatio))
+						.append(")")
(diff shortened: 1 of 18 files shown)


Xplat/src/main/java/vazkii/botania/common/handler/OrechidManager.java
```diff
@@ -64,4 +77,133 @@ public static <T extends OrechidRecipe> Collection<T> getMatchingRecipes(
...
+		int bestQ = (int) Math.round(1 / actualChance);
+		double bestError = calcError(actualChance, bestP, bestQ);
+
+		// now try to bring the error below an acceptable margin, but only with relatively small integer ratios
+		if (bestQ < MAX_RATIO_QUOTIENT && bestError > MAX_ACCEPTABLE_RATIO_ERROR) {
+			// continued fractions ahead:
+			int pNm1 = 1;
+			int pNm2 = 0;
+			int qNm1 = 0;
+			int qNm2 = 1;
```
williewillus: please name these something meaningful

the formulas probably make sense in your head right now but it's not going to to whoever is looking at this in 2 years

Xplat/src/main/java/vazkii/botania/mixin/RecipeManagerMixin.java
```diff
@@ -0,0 +1,33 @@
...
+	@Inject(at = @At("RETURN"), method = "replaceRecipes")
+	private void onRecipesReplaced(Iterable<Recipe<?>> recipes, CallbackInfo ci) {
+		RecipeManager manager = (RecipeManager) (Object) this;
+		OrechidManager.calculateTotalDisplayWeights(manager);
+	}
+
+	@Inject(at = @At("RETURN"), method = "apply(Ljava/util/Map;Lnet/minecraft/server/packs/resources/ResourceManager;Lnet/minecraft/util/profiling/ProfilerFiller;)V")
+	private void onRecipesLoaded(Map<ResourceLocation, JsonElement> jsonRecipes, ResourceManager resourceManager, ProfilerFiller profilerFiller, CallbackInfo ci) {
+		RecipeManager manager = (RecipeManager) (Object) this;
+		OrechidManager.calculateTotalDisplayWeights(manager);
```
williewillus: can we just do this lazily instead of adding a mixin?

TheRealWormbo: The question is. When can we do this? Accessing the `RecipeManager` requires a `Level` reference from somewhere, but one of the three recipe display mods provide that or a complete list of recipes to the code that builds the display panel.
Maybe the event handler that clears the `BY_TYPE` cache could trigger something, or maybe all three mods (i.e. JEI, REI and EMI) provide some kind of global initialization hook that I missed so far. (The Marimorphosis displays could actually benefit from having a reference to the player, that way they could display location-specific chances.)

Xplat/src/main/java/vazkii/botania/common/handler/OrechidManager.java
TheRealWormbo: If that code survives, I'll fix the names and even point towards some article that explains it. Currently that entire method is not used, since it was supposed to generate a ratio from the percentage chance. I just kept it around in case I find another application for the numbers it produces (e.g. some kind of tooltip).

TheRealWormbo: Code moved to `OrechidUIHelper` class, including hopefully better variable names.

Xplat/src/main/java/vazkii/botania/mixin/RecipeManagerMixin.java
TheRealWormbo: The answer appears to be yes, via `Minecraft.getInstance().level` or even `Minecraft.getInstance().player`, if the player location is needed.

TheRealWormbo:
Current state of the recipe display implementations:
- For Orechid and Orechid Ignem it displays the chance as a percentage with one digit after the decimal point, right-aligned such that the point is roughly centered above the output block. Pointing the mouse at the percentage shows a tooltip that approximates the conversion ratio with relatively small integer numbers, if possible.
Orechid and Orechid Ignem
- For Marimorphosis, the percentage display shows the conversion chance without any biome bonus. If the player opens the recipe view while in a biome with bonus weights, the tooltip also includes the chance and approximated conversion ratio for that biome.
Marimorphosis
- The approximation logic for the ratio to display in the tooltip aims to reduce the error between the ratio and the actual chance below a certain threshold. It first attempts to use the "1 out of N" ratio, but if that is too far off, it calculates the first few convergents of the continued fraction, stopping either if the error gets low enough or the denominator gets too large. The maximum allowed error and convergent denominator are defined as constants in the `OrechidUIHelper` class.
- The new chance display is implemented in JEI, REI and EMI, with as much code as reasonably possible being shared via the `OrechidUIHelper` class. Some shared code that might be useful for other features that need to display numbers in a localized format is in the class `LocaleHelper`. (Candidates would be any numbers displayed with a decimal point, thousands separator or percentage sign, as the default formatting would otherwise use the system locale instead of one matching the Minecraft language.)

Xplat/src/main/java/vazkii/botania/client/integration/shared/OrechidUIHelper.java
```diff
@@ -0,0 +1,171 @@
...
+				getRatioTooltipComponent(ratio)
+		).withStyle(ChatFormatting.GRAY);
+	}
+
+	@Nullable
+	public static Stream<Component> getBiomeChanceAndRatioTooltipComponents(double chance, OrechidRecipe recipe) {
+		final var biomeTranslationKey = OrechidUIHelper.getPlayerBiomeTranslationKey();
+		final var player = Minecraft.getInstance().player;
+		if (biomeTranslationKey == null || player == null) {
+			return null;
```
williewillus: I would just return Stream.empty() from this method to make the result NotNull and avoid the long ternaries at the callsites (you could just blindly concat the two streams)

Xplat/src/main/java/vazkii/botania/common/handler/OrechidManager.java
```diff
@@ -10,24 +10,31 @@
...
 import vazkii.botania.xplat.XplatAbstractions;
 
 import java.util.*;
+import java.util.function.ToIntFunction;
 
 import static vazkii.botania.common.lib.ResourceLocationHelper.prefix;
 
 public class OrechidManager implements ResourceManagerReloadListener {
 	private static final Map<RecipeType<? extends OrechidRecipe>, Map<BlockState, List<? extends OrechidRecipe>>> BY_TYPE = new IdentityHashMap<>();
+	private static final Map<RecipeType<? extends OrechidRecipe>, Object2IntOpenHashMap<BlockState>> TOTAL_WEIGHTS = new IdentityHashMap<>();
```
williewillus: I would make it clear from the name and in the below code that this is for positionless lookups only

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -39,6 +39,8 @@
   "botaniamisc.lexiconcover2": "A Book by Vazkii",
   "botaniamisc.shiftinfo": "Hold %s for more info",
   "botaniamisc.ratio": "%sx Zoom (hover to zoom out)",
+  "botaniamisc.conversionRatio": "about %s out of %s",
```
williewillus: Capitalize
```suggestion
  "botaniamisc.conversionRatio": "About %s out of %s",
```

williewillus (commented):
seems legit, minor comments

Xplat/src/main/java/vazkii/botania/client/integration/shared/OrechidUIHelper.java
TheRealWormbo: Done. Also removed the redundant calls to `getRatioForChance` and `getRatioTooltipComponent` in `getBiomeChanceTooltipComponent`. (The translation string there only expects two arguments, as the biome ratio has its own tooltip line.)

williewillus:
Applied manually (oh just saw you push again lmao oops)

Issue #4273, opened 2023-01-23
Ender Air Bottling Interaction blocks bottle block interactions in the End


DaFuqs:
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-436

### Modloader version

Any

### The latest.log file

---

### Issue description

The ender air bottling logic is implemented via `UseItemCallback.EVENT.register(EnderAirItem::onPlayerInteract);`. It returns sucess() if the player is in the End and no dragon breath would be picked up.

Returning pass at this stage will skip any bottle on block interaction, since the interaction has already been handled, preventing players to interact with blocks using a bottle in the end (like bottling fluids).

A suggested fix would be to only collect Ender Air is the player is not targeting a block, while otherwise calling `pass()`.

### Steps to reproduce

1. Go to the End
2. Right Click with a bottle anywhere that is not near dragon's breath (in the air, on water, on stone, ...)
3. You get Ender Air

TheRealWormbo:
Issue also exists on Forge. However, on both platforms it only applies to bottling water. Interaction with e.g. bee hives works fine since that's a separate event that fires before Botania's ender air pickup attempt.
Might be a relatively simple fix by also checking whether the player targets a fluid.

TheRealWormbo:
I believe the change from #5040 will solve this in 1.21, but we won't adjust bottling behavior any further for 1.20 anymore.

PR #4327, opened 2023-04-13
Fix water bottle filling not working in the end dimension

TheRealWormbo:
Fix for #4273. Botania's logic for creating ender air bottles is skipped if the player aims at water.
(Grabbing fluids from blocks, e.g. getting honey bottles from full bee nests, was not affected.)

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/material/EnderAirItem.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/material/EnderAirItem.java
@@ -8,6 +8,7 @@
  */
 package vazkii.botania.common.item.material;
 
+import net.minecraft.core.BlockPos;
 import net.minecraft.core.particles.ParticleTypes;
 import net.minecraft.sounds.SoundEvents;
 import net.minecraft.sounds.SoundSource;
@@ -19,8 +20,12 @@ import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
+import net.minecraft.world.level.ClipContext;
 import net.minecraft.world.level.Level;
+import net.minecraft.world.level.gameevent.GameEvent;
 import net.minecraft.world.phys.AABB;
+import net.minecraft.world.phys.BlockHitResult;
+import net.minecraft.world.phys.HitResult;
 
 import org.jetbrains.annotations.NotNull;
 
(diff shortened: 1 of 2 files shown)


Xplat/src/main/java/vazkii/botania/common/item/material/EnderAirItem.java
```diff
@@ -43,22 +49,32 @@ public static InteractionResultHolder<ItemStack> onPlayerInteract(Player player,
...
 		}
 
 		return InteractionResultHolder.pass(stack);
 	}
 
+	private static boolean notAimingAtFluid(Level world, Player player) {
+		BlockHitResult hitResult = getPlayerPOVHitResult(world, player, ClipContext.Fluid.SOURCE_ONLY);
+		if (hitResult.getType() == HitResult.Type.BLOCK) {
+			BlockPos pos = hitResult.getBlockPos();
+			return world.mayInteract(player, pos) && !world.getFluidState(pos).is(FluidTags.WATER);
```
williewillus: why is this mayInteract here? Pretty sure mc internals already do this check before calling any item specific stuff.

TheRealWormbo: The `BottleItem` also does a `mayInteract` check first. I believe the "global" `mayInteract` check applies to the player's location, while this one here applies to the block the bottle is potentially being used on.
The situation this covers is as follows: The player stands in an unprotected area, but aims at a water block in a protected area. The question is, should that mean the player can't use the bottle at all (without the `mayInteract` call) or should it fall back to picking up air from the end (with the call).

TheRealWormbo:
Just wondering, that Discord discussion about potentially using empty managlass vials or even alfglass flasks (or either?) for bottling ender air instead of using regular empty bottles – is that something you folks would want something prototyped for?

Xplat/src/main/java/vazkii/botania/common/item/material/EnderAirItem.java
williewillus: also, is this meant to only check for water or absence of any fluid? the logic doesn't match the method name

TheRealWormbo: It was supposed to be water (covering the vanilla glass bottle logic), but the bug report mentioned fluids in general. Which way do you prefer?

TheRealWormbo: Changed it to check for absolutely any fluids (even flowing ones) for maximum compatibility. You need to use the bottle in the air or against some block now.

Issue #4306, opened 2023-02-28
rannu can't place multiple candles on a same block


LeoBeliik:
### Mod Loader

Forge

### Minecraft Version

1.19.2

### Botania version

Botania-1.19.2-437-FORGE

### Modloader version

Forge 1.19.2 - 43.1.65

### The latest.log file

-

### Issue description

rannuncarpus can't place multiple items like candles or pickles on a same blockspace 
![2023-02-28_17 44 57](https://user-images.githubusercontent.com/11804768/221975442-6e639f32-cad3-47ba-bd6c-6660c3bb3e79.png)

### Steps to reproduce

place a rann
throw candles or pickles at it
observe

PR #4329, opened 2023-04-13
Rannuncarpus can now add more candles and sea pickles to matching existing blocks

TheRealWormbo:
Fix for #4306. Rannuncarpus will do the complete replacement check that block items do when a player attempts to place them, not just against replaceable materials. That allows them to place multiple sea pickles or matching candles on the same block, and should be compatible with any other block that would allow such placement.


williewillus:
Looks good but please add a test

williewillus:
very nice

williewillus:
I thought about this more. Isn't this going to make e.g. placing single candles a lot more annoying?

Let's say I want to use the rannun to automate some magic mod's ritual which requires candles in the corners of a 3x3.

Now, instead of dropping 4 candles and guaranteeing each position gets one candle, I have to worry about the rannun choosing to double/triple/quad stack the candles in one spot. So now I have to use pistons or other jank to move the matching underneath to something else (which might not be possible even).

So, the UX of this needs more work I think.

artemisSystem:
I would still like there to be a way to place multiple candles in one block, at least. And i think the now new behavior of being able to place multiple is the most natural option. 

Or maybe it should prioritize empty/air blocks, and then if there are none, then it checks for otherwise placeable locations?

PR #4330, opened 2023-04-14
Fix redstone still breaking when placed on exchanged blocks after #4316

jpenilla:
This reverts back to the original implementation I had for preventing blocks attached to replaced blocks from breaking.

After further testing I found that the current solution only works for simple things like torches, but not more complex things like redstone dust.


williewillus:
hope this doesn't break anything tm

Issue #4312, opened 2023-03-12
[Feature Request] Allow the Sextant to create Spheres


f-raZ0R:
The Worldshaper's Sextant is cool and all, but creating spheres/domes is just as annoying as creating a circle, if not more so. The sextant is meant to be an easy way to create circular things in builds without having to tab out to a guide, but the fact that it doesn't do spheres/domes means that half of the time you'll be doing that regardless.

How this would work:
Shift-left clicking (or some other keybind) with the worldshaper's sextant, will swap it between "circle mode", and "sphere mode". In circle mode, it will act as it normally does, creating a circle. In sphere mode, it will create a sphere (wow who would've guessed that), again, centered at the block chosen, with a radius defined at where you look. Switching between circle mode and sphere mode while an imaginary circle/sphere is already created, will also swap the current selection from a circle to a sphere, or vice versa.

As a cute touch, maybe the sextant in sphere mode could look a little different? idk what it would look like though.

I have gotten express permission from williewillus to post this feature request here.

PR #4333, opened 2023-04-17
Worldshaper's Sextant sphere mode and finer radius control

TheRealWormbo:
Implementation for #4312.
- Sphere mode can be toggled by sneak+right-clicking while the sextant doesn't show a preview.
- Both the sphere *and* circle mode generate a thin shape now, containing only blocks that are visible from the outside. (The old version of the circle mode didn't necessarily do that all the time.) Since there are multiple reasonable ways to represent a voxel sphere or circle of the same general radius, the sextant now also accepts higher precision radii, which can cause the shape to vary slightly. Radius selection is still restricted to discrete values by snapping to a block-sized grid. (It did that before already, but also rounded down.)
- And finally, this PR restores the particle effects that should be displayed while selecting a radius and makes them slightly fancier. There also is a new version of the effect to be displayed while selecting a sphere radius:
Sphere mode selection in progress

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/WorldshaperssSextantItem.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/WorldshaperssSextantItem.java
@@ -50,8 +50,8 @@ import vazkii.patchouli.api.PatchouliAPI;
 
 import java.math.RoundingMode;
 import java.text.NumberFormat;
+import java.util.Arrays;
 import java.util.HashMap;
-import java.util.Locale;
 import java.util.Map;
 import java.util.stream.Stream;
 
@@ -63,7 +63,7 @@ public class WorldshaperssSextantItem extends Item {
 	private static final String TAG_SOURCE_X = "sourceX";
 	private static final String TAG_SOURCE_Y = "sourceY";
 	private static final String TAG_SOURCE_Z = "sourceZ";
-	private static final String TAG_SPHERE_MODE = "sphereMode";
+	private static final String TAG_MODE = "mode";
 
 	public WorldshaperssSextantItem(Properties builder) {
 		super(builder);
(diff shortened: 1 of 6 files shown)


Xplat/src/main/java/vazkii/botania/common/item/WorldshaperssSextantItem.java
```diff
@@ -112,32 +165,67 @@ public void releaseUsing(ItemStack stack, Level world, LivingEntity living, int
...
+		} else {
+			ItemNBTHelper.setInt(stack, TAG_SOURCE_Y, Integer.MIN_VALUE);
+		}
 		if (world.isClientSide) {
 			Proxy.INSTANCE.clearSextantMultiblock();
 		}
 	}
 
+	private static void setSphereMode(ItemStack stack, boolean sphereMode) {
+		ItemNBTHelper.setBoolean(stack, TAG_SPHERE_MODE, sphereMode);
```
williewillus: let's just store this as a string in `TAG_MODE` in case we want to add more shapes in the future

Xplat/src/main/java/vazkii/botania/common/item/WorldshaperssSextantItem.java
```diff
@@ -219,4 +307,35 @@ public static void render(PoseStack ms, Player player, ItemStack stack) {
...
+		var format = NumberFormat.getInstance(getLocale());
+		format.setRoundingMode(RoundingMode.HALF_UP);
+		format.setMaximumFractionDigits(numDigits);
+		format.setMinimumFractionDigits(numDigits);
+		return format;
+	}
+
+	private static Locale getLocale() {
+		final String languageCode = Minecraft.getInstance().getLanguageManager().getSelected().getCode();
+		return Locale.forLanguageTag(languageCode);
```
williewillus: this touches client code. I know it's actually safe in both the usages here, but it's easy to introduce easily-missed server-only bugs. I would move getLocale to `Proxy` and have it return en_us by default, then override it in `ClientProxy` to look up the client locale.

Xplat/src/main/java/vazkii/botania/common/item/WorldshaperssSextantItem.java
```diff
@@ -219,4 +307,35 @@ public static void render(PoseStack ms, Player player, ItemStack stack) {
...
+		return "botaniamisc.sextantMode." + (getSphereMode(stack) ? "sphere" : "circle");
+	}
+
+	private static String getRadiusString(double radius) {
+		NumberFormat format = getNumberFormat(1);
+
+		return format.format(radius);
+	}
+
+	// these two methods should probably move to a Locale helper class:
```
williewillus: fine with keeping this one here, though I'd just inline the param value since it seems like you only pass `1`.

williewillus (changes requested):
looks good, some adjustments

PR #4336, opened 2023-04-20
Delay Runic Altar picking up it's outputs, and allow Hopperhocks to pick up outputs quicker

jpenilla:
Brings behavior more in line with Mana infusions, making automation easier by not needing an Influence lens on a pulse spreader to move the items before the Altar picks them back up. Looking at older YouTube videos it seems like this is how it worked in the past, and at some point something changed to where you now need an Influence lens.

edit: I've noticed that Runic Altars already have their own cooldown, but I still think this change is good both for allowing Hopperhocks to pick up outputs more quickly and for preventing nearby altars from grabbing each other's outputs accidentally. It also might make sense to reduce that cooldown from the current 60 ticks along with this change?

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/internal_caps/ItemFlagsComponent.java
+++ b/Xplat/src/main/java/vazkii/botania/common/internal_caps/ItemFlagsComponent.java
@@ -29,7 +29,12 @@ public class ItemFlagsComponent extends SerializableComponent {
 	 * Used so certain mechanics don't interact with items immediately after they're produced.
 	 */
 	private int manaInfusionCooldown = 0;
+	/**
+	 * Similar to {@link #manaInfusionCooldown} but for the Runic Altar.
+	 */
+	private int runicAltarCooldown = 0;
 	public static final int INITIAL_MANA_INFUSION_COOLDOWN = 25;
+	public static final int INITIAL_RUNIC_ALTAR_COOLDOWN = 15;
 
 	@Override
 	public void readFromNbt(CompoundTag tag) {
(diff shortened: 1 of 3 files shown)


Xplat/src/main/java/vazkii/botania/common/internal_caps/ItemFlagsComponent.java
```diff
@@ -29,14 +29,20 @@ public class ItemFlagsComponent extends SerializableComponent {
 	 * Used so certain mechanics don't interact with items immediately after they're produced.
 	 */
 	private int manaInfusionCooldown = 0;
+	/**
+	 * Similar to {@link #manaInfusionCooldown} but for the Runic Altar.
+	 */
+	private int runicAltarCooldown = 0;
 	public static final int INITIAL_MANA_INFUSION_COOLDOWN = 25;
+	public static final int INITIAL_RUNIC_ALTAR_COOLDOWN = 15;
```
williewillus: any rationale behind this number? if not, let's align it with the mana infusion number for consistency.

jpenilla: Basically, I felt like it's slightly more likely someone would want to intentionally reuse outputs as inputs for the next craft compared to the mana pool. For example, a machine that produces one rune type. So I chose a lower number to not slow down such automation as much, while still giving the desired faster hopperhock pickup. 

...although for that to make any difference we would need to also lower the existing altar cooldown from 60 ticks (which I think a smaller cooldown makes sense and is fine now that items also have their own cooldown).

jpenilla: Ended up just making it the same.

PR #4337, opened 2023-04-21
Improve Heisei Dream (partially fixes #4064)

### LINKED ISSUE #4064: Some mobs are unaffected by Heisei Dream and Charm of the Diva
### Mod Loader

Forge

### Minecraft Version

1.18.2

### Botania version

1.18.2-432

### Modloader version

Forge: 40.1.25

### Modpack info

https://www.curseforge.com/minecraft/modpacks/the-phoenixs-magics-and-goodies

### The latest.log file

https://gist.github.com/TheRealWormbo/5ea9f086a04ae59dbdf036c84a8d5498

### Issue description

Some mob pairings exhibit weird behaviors while under the effect of a Heisei Dream. Some mobs seem to "punch themselves", while others will not attack each other in any way, but you can see the flower consuming mana again and again.

### Steps to reproduce

Place an Everlasting Guilty Pool (for convenience) and a Heisei Dream, then use spawn eggs create pairs of mobs within range of the flower:

* Two shulkers – nothing appears to happen, except the flower occasionally draining mana. (Can be observed via looking at the flower while holding a Wand of the Forest.)
* Two husks – one of them immediately starts taking damage, even though the other is definitely not in melee range yet.
* Two blazes – one of them again immediately starts taking damage (punching itself? it doesn't last long), while the other watches and prepares to fire. It may shoot eventually.
* Any pairing of piglins and/or piglin brutes – like with shulkers, nothing happens. If not testing this in the nether, they will eventually convert to zombified piglins, but only start attacking each other when both have converted. If they are far enough apart from each other, you can likely see one punching itself again as in the husk case. In fact, both piglin types don't appear to attack (or get attacked by) any other hostile mobs, except wither skeletons, which they would attack anyway.

Other zombie variants, and surprisingly also silverfish or endermites exhibit the self-punching behavior as well, while wither skeletons or endermen don't. Mixing mob types among the "self-punchers" results in no self-punching behavior, as far as I can tell. Guardians and zoglins both are among the mobs that won't attack others of their own kind.

### Other information

While I first noticed the issue in the linked mod pack, I could also reproduce it with a plain Forge installation in GDLauncher that contains nothing but Botania, Patchouli and Curios. The log file is from a test with that profile.

jpenilla:
Overrides checks in TargetingConditions that prevented Illagers from attacking each other

Originally wanted to also fix it for mobs using the new AI system (ie Piglins), but that turned out to be a lot more complicated than I was hoping it would be (so I'm leaving it for a future PR).

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/HurtByTargetGoalAccess.java
+++ b/Xplat/src/main/java/vazkii/botania/common/HurtByTargetGoalAccess.java
@@ -1,0 +1,5 @@
+package vazkii.botania.common;
+
+public interface HurtByTargetGoalAccess {
+	void botania$setBrainwashed();
+}
--- a/Xplat/src/main/java/vazkii/botania/common/TargetingConditionsAccess.java
+++ b/Xplat/src/main/java/vazkii/botania/common/TargetingConditionsAccess.java
@@ -1,0 +1,5 @@
+package vazkii.botania.common;
+
+public interface TargetingConditionsAccess {
+	void botania$setBrainwashed();
+}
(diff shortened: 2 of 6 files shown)


TheRealWormbo:
We will likely reimplement this PR after the 1.21 port, using MixinExtras features.

PR #4339, opened 2023-04-23
Sort recipe categories in EMI

### LINKED ISSUE #4326: Botania categories are not sorted in EMI
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-438

### Modloader version

Loader 0.14.17, API 0.76.0+1.19.2

### Modpack info

_No response_

### The latest.log file

not applicable

### Issue description

Botania does not specify any particular sorting order for its EMI categories, while it does so for its JEI categories. The default sorting order seems to be mostly stable between restarts, but doesn't appear to follow any specific logic.

For reference, the JEI sorting definitions from `JEIBotaniaPlugin#registerRecipes`:
- Brewery, Pure Daisy, Petal Apothecary, Elven Trades, Runic Altar, Terrestrial Agglomeration Plate: ID
- Mana Infusion: catalyst type, group (e.g. stone type conversion cycle), ID
- Orechid, Orechid Ignem, Marimorphosis: weight (not biome-specific for Marimorphosis), ID

### Steps to reproduce

_No response_

### Other information

As far as I can tell, EMI does provide a way to specify a sorting order. It may just not be immediately obvious in its API, because it's "hidden away" in the corresponding constructor variant with the most parameters.

TheRealWormbo:
- Applies the same sorting order as the JEI plugin. (fixes #4326)
- More mana infusion recipes were put into groups, where it made sense.


williewillus:
In the future, I'd split those two changes into separate PR's, but thanks.

PR #4341, opened 2023-04-24
Ring of loki improvement

LewisSaber:
Item and block interactions in minecraft have 2 sides, item side and block side, before ring of loki was only able to trigger action of item, now it will first try to trigger block activation, this can be useful when you want to configure a lot of ae2 components, or want to connect a lot of wires/pipes


LewisSaber:
I apologize, I wanted to PR to GTNH fork, sorry

Issue #3439, opened 2020-09-02
Trinket Case icon doesn't change when right clicking


SaikoChan1:
# Version Information
Forge version:  32.0.108
Botania version:  Botania-1.16.1-403

Steps to reproduce:
1. First step: right click
2. Second step: the box not openened
3. etc...:

What I expected to happen:
the box icon would open when I right click

What happened instead:
it didn't open :(

williewillus:
seems like this texture was never actually used

PR #4351, opened 2023-04-30
Enable opened icon for trinket case

TheRealWormbo:
When a trinket case is opened, its icon changes to the opened version. This change should also be visible to other players, when the case is in the main hand or off-hand. (implements #3439 )

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/client/gui/box/BaubleBoxContainer.java
+++ b/Xplat/src/main/java/vazkii/botania/client/gui/box/BaubleBoxContainer.java
@@ -124,7 +124,9 @@ public class BaubleBoxContainer extends AbstractContainerMenu {
 
 	@Override
 	public void removed(@NotNull Player player) {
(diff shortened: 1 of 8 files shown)


Xplat/src/main/java/vazkii/botania/client/gui/box/BaubleBoxContainer.java
```diff
@@ -121,4 +122,9 @@ public ItemStack quickMoveStack(Player player, int slotIndex) {
 		return itemstack;
 	}
 
+	@Override
+	public void removed(@NotNull Player player) {
+		ItemNBTHelper.setBoolean(box, BaubleBoxItem.TAG_OPEN, false);
```
williewillus: from briefly skimming vanilla, I think you need to check for `!player.level.isClient` here as this method is called on both sides.

williewillus (changes requested):
one thing

Issue #4277, opened 2023-01-29
add a toggle for the autocrafting of the manufactory halo


LeoBeliik:
title

TheRealWormbo:
The Assembly Halo's interactions currently are:
- right-click crafting table segment: Open crafting table interface
- right-click other segments: Store last crafting recipe or craft that item
- shift + (right or left) click a segment: clear that segment, if it's not the crafting segment
- shift + right-click crafting table segment: same as without shift
- shift + left-click crafting table segment: no effect

The Manufacturing Halo doesn't appear to modify these click interactions. Seems like it could repurpose one of the shift click actions on the crafting table segment to toggle auto-crafting. For consistency with other toggled items that should probably be shift + right-click, even though shift + left-click is not used.

LeoBeliik:
I have it (in my personal tweaks project) set it to be toggled via right click in the inventory, like the stone of temp, so I'll suggest that approach plus the sneak right click whilst just aiming to the crafting table segment for consistency as you said

TheRealWormbo:
The toggle is implemented as suggested (either right-click on the item in the inventory, or shift + right-click on the crafting table segment), with an indicator in the item tooltip and changing color of the rendered segment backgrounds (Assembly Halo blue = disabled, Manufactory Halo cyan = enabled). The only thing still missing is a proper visual indicator in the item's icon. I had an idea for a "programmer art" placeholder texture for the disabled halo, but didn't want to potentially break resource packs, so it's not included yet.

TheRealWormbo:
A better item texture for the active/inactive states from someone actually versed with texture creation would be nice.

PR #4352, opened 2023-05-01
Add toggle option to Manufactory Halo

TheRealWormbo:
Auto-crafting can be toggled similar to the Stone of Temperance's effect, either by right-clicking the Manufactory Halo in an inventory screen, or by sneak right-clicking on the crafting table segment while holding it, as suggested in issue #4277.
The auto-crafting status is visualized via tooltip (like for the Stone of Temperance) and via the color of the crafting segments when holding it (with auto-crafting disabled, the segment background color switches to that of the Assembly Halo). It'd probably be nice to also modify the item's icon, but there's currently no separate texture and I think reusing the Assembly Halo's icon might be a bit confusing.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/ManufactoryHaloItem.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/ManufactoryHaloItem.java
@@ -8,16 +8,32 @@
  */
 package vazkii.botania.common.item;
 
+import net.minecraft.network.chat.Component;
 import net.minecraft.resources.ResourceLocation;
+import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.InteractionHand;
+import net.minecraft.world.InteractionResultHolder;
 import net.minecraft.world.entity.Entity;
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.SlotAccess;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.inventory.ClickAction;
+import net.minecraft.world.inventory.Slot;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.item.TooltipFlag;
 import net.minecraft.world.level.Level;
 
+import org.jetbrains.annotations.NotNull;
+
 import vazkii.botania.client.lib.ResourcesLib;
+import vazkii.botania.common.handler.BotaniaSounds;
+import vazkii.botania.common.helper.ItemNBTHelper;
+
+import java.util.List;
 
 public class ManufactoryHaloItem extends AssemblyHaloItem {
+	public static final String TAG_ACTIVE = "active";
 
 	private static final ResourceLocation glowTexture = new ResourceLocation(ResourcesLib.MISC_GLOW_CYAN);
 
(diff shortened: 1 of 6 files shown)


Xplat/src/main/java/vazkii/botania/common/item/ManufactoryHaloItem.java
```diff
@@ -38,8 +54,51 @@ public void inventoryTick(ItemStack stack, Level world, Entity entity, int pos,
...
+
+	private void togglePassive(ItemStack stack, LivingEntity living, Level world) {
+		ItemNBTHelper.setBoolean(stack, TAG_PASSIVE, !isPassive(stack));
+		if (living instanceof Player player && world != null) {
+			world.playSound(player, player.getX(), player.getY(), player.getZ(), BotaniaSounds.manufactoryHaloConfigure, SoundSource.NEUTRAL, 1F, 1F);
+		}
+	}
+
+	private static boolean isPassive(ItemStack stack) {
+		return ItemNBTHelper.getBoolean(stack, TAG_PASSIVE, false);
```
williewillus: why is this stored as passive instead of active? Usually you'd store active instead so reading the code was confusing at first glance.

If it's a forward-compatibility concern that old worlds won't have the active tag, just make it default to `true` instead of `false`.

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -2895,7 +2895,7 @@
   "botania.entry.autocraftingHalo": "Manufactory Halo",
   "botania.tagline.autocraftingHalo": "An Assembly Halo that crafts for you",
   "botania.page.autocraftingHalo0": "The $(item)Manufactory Halo$(0) is an upgrade of the $(l:tools/crafting_halo)$(item)Assembly Halo$(0)$(/l): aside from doing all that the $(l:tools/crafting_halo)$(item)Assembly Halo$(0)$(/l) can do, the $(item)Manufactory Halo$(0) will (when $(o)not$() held) automatically craft all possible saved recipes.$(p)Automatic compression of ores while mining is just one of many applications this upgrade has.",
-  "botania.page.autocraftingHalo1": "Halo Now Launcher",
+  "botania.page.autocraftingHalo1": "Halo Now Launcher$(p)Automatic crafting can be toggled on/off by right-clicking the $(item)Manufactory Halo$(0) in your inventory, or by right-clicking its $(item)Crafting Table$(0) segment while sneaking.",
```
williewillus, line 2901: Does this all fit on the recipe page ingame?

williewillus (changes requested):
minor nits

Xplat/src/main/resources/assets/botania/lang/en_us.json
TheRealWormbo: It does at GUI scale 2 for me. That's the page with the crafting grid, which is top-aligned, leaving enough space at the bottom. Scale shouldn't affect it, right?
Screenshot at GUI scale 3 (most pages aren't actually readable at GUI scale 1):
Manufactory Halo entry in Lexica Botania

williewillus:
lgtm, rebase and fix the changelog entry and I'll apply

PR #4360, opened 2023-05-12
Make dispensers able to use pasture seeds

### LINKED ISSUE #4344: Dispenser don't use pasture seeds on dirt
### Mod Loader

Forge

### Minecraft Version

1.19.2

### Botania version

Botania-1.19.2-438-FORGE

### Modloader version

43.2.8

### Issue description

Using pasture seeds with a dispenser don't use them, instead drops em as item, this also counts for all the other botania seeds.

The expected behavior would be for the dispenser to use the item in dirt to generate the desired type of grass.

### Steps to reproduce

place a dispenser pointing to a dirt block or the top of said block
put a pasture seed in the dispenser
give the dispenser a redstone signal
observe

### Other information

_No response_

zacharybarbanell:
#4344


Xplat/src/main/java/vazkii/botania/common/block/BotaniaBlocks.java
```diff
@@ -940,6 +940,17 @@ protected Projectile getProjectile(@NotNull Level world, @NotNull Position pos,
...
+		DispenserBlock.registerBehavior(BotaniaItems.grassSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.podzolSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.mycelSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.drySeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.goldenSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.vividSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.scorchedSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.infusedSeeds, behavior);
+		DispenserBlock.registerBehavior(BotaniaItems.mutatedSeeds, behavior);
+
```
williewillus, line 958: how about something like:

```
Item[] seeds = { BotaniaItems.grassSeeds, ...};
for (Item seed : seeds) {
  DispenserBlock.registerBehavior(item, behavior);
}
```
to avoid repeating the method call over and over

Xplat/src/main/java/vazkii/botania/common/block/dispenser/GrassSeedsBehavior.java
```diff
@@ -0,0 +1,36 @@
...
+		ServerLevel world = source.getLevel();
+		BlockPos pos = source.getPos().relative(source.getBlockState().getValue(DispenserBlock.FACING));
+
+		setSuccess(((GrassSeedsItem) stack.getItem()).applySeeds(world, pos, stack));
+
+		if (isSuccess()) {
+			XplatAbstractions.INSTANCE.sendToNear(world, pos,
+					new BotaniaEffectPacket(EffectType.GRASS_SEED_PARTICLES,
+							(double) pos.getX(), (double) pos.getY(), (double) pos.getZ(),
+							GrassSeedsItem.getID(((GrassSeedsItem) stack.getItem()).getIslandType(null))));
```
williewillus: why null and not `stack`?

Xplat/src/main/java/vazkii/botania/common/item/GrassSeedsItem.java
```diff
@@ -35,17 +35,28 @@ public class GrassSeedsItem extends Item implements FloatingFlowerVariant {
...
+	private static final Map<IslandType, Integer> IDS = ImmutableMap.<IslandType, Integer>builder()
+			.put(IslandType.GRASS, 0)
+			.put(IslandType.PODZOL, 1)
+			.put(IslandType.MYCEL, 2)
+			.put(IslandType.DRY, 3)
+			.put(IslandType.GOLDEN, 4)
+			.put(IslandType.VIVID, 5)
+			.put(IslandType.SCORCHED, 6)
+			.put(IslandType.INFUSED, 7)
+			.put(IslandType.MUTATED, 8)
```
williewillus: We shouldn't have an ID system here, that's one more thing that could go wrong especially with addon supplied island types.
Instead, I'd do the following:

1. Keep the existing map here, but replace the floats with their packed integer hex-color equivalents
2. Grab the color from this map serverside and send it in the packet in the DispenserBehavior
3. The client just uses the color given in the packet.

Xplat/src/main/java/vazkii/botania/common/item/GrassSeedsItem.java
```diff
@@ -59,44 +70,54 @@ public GrassSeedsItem(IslandType type, Properties props) {
...
 		BlockPos pos = ctx.getClickedPos();
-		BlockState state = world.getBlockState(pos);
 		ItemStack stack = ctx.getItemInHand();
 
+		return applySeeds(world, pos, stack)
+				? InteractionResult.sidedSuccess(world.isClientSide())
+				: InteractionResult.PASS;
+	}
+
+	public boolean applySeeds(Level world, BlockPos pos, ItemStack stack) {
```
williewillus: An antipattern I've been trying to move away from is interaction methods that return bools. Instead, keep the existing InteractionResult return values. In `useOn`, just return it through, and in the `DispenserBehavior`, use `shouldConsume` to convert it to a bool near the return site. That way the result value stays as a meaningful enum as long as possible before turning into a meaningless bool.

Xplat/src/main/java/vazkii/botania/network/EffectType.java
```diff
@@ -15,7 +15,8 @@ public enum EffectType {
 	HALO_CRAFT(1), // Arg: Entity ID
 	AVATAR_TORNADO_JUMP(1), // Arg: Entity ID
 	AVATAR_TORNADO_BOOST(1), // Arg: Entity ID
-	THUNDERCALLER_EFFECT(-1) // Arg: All entity IDs hit by the bolt
+	THUNDERCALLER_EFFECT(-1), // Arg: All entity IDs hit by the bolt
+	GRASS_SEED_PARTICLES(1) // Arg: ID of the type of seeds
```
williewillus: As discussed above
```suggestion
	GRASS_SEED_PARTICLES(1) // Arg: Color of the particles
```

williewillus (changes requested):

Xplat/src/main/java/vazkii/botania/common/item/GrassSeedsItem.java
williewillus: I realize what I suggest still isn't open to addons, but it at least avoids yet another ID system

Xplat/src/main/java/vazkii/botania/common/block/dispenser/GrassSeedsBehavior.java
zacharybarbanell: Because the argument is unused, and I don't know why, semantically, it ought to be `stack` - I'm not honestly sure what `getIslandType` is supposed to do with its argument

williewillus: Please pass stack then, just because it's unused now doesn't mean you should pass something meaningless like null, because it can start to be used in the future.

PR #4361, opened 2023-05-12
Use a tag for items that can merge lenses, and add honey bottles to it

### LINKED ISSUE #4272: Make the composite lens crafting recipe use a new tag instead of hardcoded for slimeballs
### Mod Loader

### Minecraft Version

1.19.2

### Botania version

Release 1.19.2-436

### Modloader version

doesn't matter

### Modpack info

_No response_

### The latest.log file

no

### Issue description

Regarding this line: https://github.com/VazkiiMods/Botania/blob/1.19.x/Xplat/src/main/java/vazkii/botania/common/crafting/recipe/CompositeLensRecipe.java#L46

It might be nice for modpack makers to be able to change the item there instead of having it be hardcoded to slimeballs.

### Steps to reproduce

_No response_

### Other information

_No response_

zacharybarbanell:
#4272 

In addition to providing the requested feature, this also adds honey bottles to the new tag (in addition to slime bottles).


Xplat/src/main/java/vazkii/botania/common/crafting/recipe/CompositeLensRecipe.java
```diff
@@ -43,7 +44,7 @@ public boolean matches(@NotNull CraftingContainer inv, @NotNull Level world) {
 					} else {
 						foundLens = true;
 					}
-				} else if (stack.is(Items.SLIME_BALL) && !foundSlimeball) {
+				} else if (stack.is(BotaniaTags.Items.LENS_GLUE) && !foundSlimeball) {
 					foundSlimeball = true;
```
artemisSystem: change this variable name to `foundLensGlue`?

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -188,6 +188,7 @@
   "tag.botania.mini_flowers": "Petite Flowers",
   "tag.botania.mana_using_items": "Mana Using Items",
   "tag.botania.lens": "Mana Lenses",
+  "tag.botania.lens_glue": "Lens Glues",
```
artemisSystem, line 194: I feel like either, the tag should be named `lens_glues`, or the translation should be `"Lens Glue"`. Not sure which i prefer. I think the convention for tags is to be plural, though.

artemisSystem (changes requested):
Looks good, but see the nits commented inline.

Does the lexica botania mention slimeballs specifically where it talks about lens merging? If so, that needs to be updated.

Also, did you run spotless?

artemisSystem:
Oh, and can you update the changelog?

Something like "* Add: Mana lenses can now be merged using honey bottles in addition to slime balls" is probably fine.

Xplat/src/main/resources/assets/botania/lang/en_us.json
zacharybarbanell: I was going off "lens", which has the internal name be singular and the display name be plural, but I could change the internal name to be plural as well, if desired.

TheRealWormbo:
Would it make sense to include platform-specific slimeball item tags (`#forge:slimeballs`, Fabric's `#c:slimeballs`) on their respective platforms by default?

artemisSystem:
yes, good idea. they'd need to be marked optional

zacharybarbanell:
what do you mean by "marked optional"?

TheRealWormbo:
> what do you mean by "marked optional"?

Something similar to the inclusion of the Quark frog in the [rare cocoon entities](https://github.com/VazkiiMods/Botania/blob/1.19.x/Xplat/src/generated/resources/data/botania/tags/entity_types/cocoon/rare.json), I'd guess.

artemisSystem:
yeah, like that

zacharybarbanell:
Is the correct fabric tag `#c:slimeballs` or `#c:slime_balls`?  The wiki only mentions the latter, though I'm unsure if it's a reliable source.

artemisSystem:
unsure if this is the wiki you meant, but i'd say this is a good source. https://fabricmc.net/wiki/tutorial:tags. c:slime_balls it is

artemisSystem:
Mana Benses (in the changelog)

zacharybarbanell:
I'm not sure what the proper cross-platform way to get the list of items from the tag, for the JEI compat, is.

williewillus:
zacharybarbanell ***@***.***> writes:

> I'm not sure what the proper cross-platform way to get the list of items from the tag, for the JEI compat, is.

In `CompositeLensRecipeWrapper`, query the list of all glue items in the
constructor, like how the existing code does for `allLenses`.

zacharybarbanell:
Another thing I noticed while resolving this - the JEI and EMI versions of the recipe have different shapes in the crafting grid.  That seems out of the scope of this PR, but should those be unified?

williewillus:
zacharybarbanell ***@***.***> writes:

> Another thing I noticed while resolving this - the JEI and EMI versions of the recipe have different shapes in the crafting grid. That seems out of the scope of
> this PR, but should those be unified?
>

That can be done in a followup, yeah.

The changes lgtm, so rebase and I'll land it.

williewillus:
Manually applied

PR #4362, opened 2023-05-13
Use phantom ink instead of managlass to make abstruse platforms invisible

zacharybarbanell:
#4279


Xplat/src/main/java/vazkii/botania/common/block/block_entity/PlatformBlockEntity.java
```diff
@@ -107,9 +107,6 @@ public void writePacketNBT(CompoundTag cmp) {
 	@Override
 	public void readPacketNBT(CompoundTag cmp) {
 		BlockState state = NbtUtils.readBlockState(cmp.getCompound(TAG_CAMO));
-		if (state.isAir()) {
-			state = null;
-		}
```
williewillus, line 112: I'm very slightly concerned what happens if the camo state is set to a block that is later removed from the game (e.g. a modded block), because then the camoState would turn into air (whereas before this PR it would turn back into an uncamouflaged platform). Not sure if it's a problem.

TheRealWormbo:
I'm going to assume nothing is happening with this PR anymore. I'll take the changes and finalize them in a new PR targeting the 1.20 branch.

zacharybarbanell:
Sounds good to me.

PR #4363, opened 2023-05-13
add Fabric/.vscode/ to gitignore

zacharybarbanell:


williewillus:
Does this need to be done for forge too?

zacharybarbanell:
I don't believe so - I think the only forge command that interacts with vscode is `genVSCodeRuns`, which doesn't seem to make anything in a comparable folder.

PR #4366, opened 2023-05-18
Upload zh_cn files

Dawnwalker666:
The main work is to sort the entries, add missing entries and fix some mistranslations. 
In addition, it makes the number of zh_cn lines aligned with en_us, which is convenient for future Chinese translation. Currently in this file, the number of zh_cn lines = en_us + 2 
The reason for this is that Chinese files have two more lines of opening comments than English.
(you might not know that there are more than 3000 lines in total, it takes me more time to compare them one by one...)

***
I don't know how to translate line 209, there doesn't seem to be an entry or item that matches it.
The **Seed Reagents** on line 210, after I checked the changelog, may be a change from 1.19.2-436.
`
Add: Petal Apothecary recipes' reagent item (seeds) can now be customized in JSON. This has implications on how the Apothecary matches items, see commit 3149ffd02cce6e972fbe04ef949ebb2790924dc1 for more information.
`
It may be noted that seed will like catalysts that are put into the apothecary.But obviously I don't know if it's used here...Don't blame me please..

1578-1593, I guess it is to adjust the botania logo color, but I did not find where to adjust. I don't know what the final result will look like, so I'll have to try a direct translation.


Dawnwalker666:
I understand the question above. Special thanks to @TheRealWormbo in discord for answering my questions.

Xplat/src/main/resources/assets/botania/lang/zh_cn.json
```diff
@@ -432,6 +502,8 @@
 
   "advancement.botania:desuGun": "裏表ラバーズ",
   "advancement.botania:desuGun.desc": "ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN ASADA-SAN",
+  "advancement.botania:tinyPotatoBirthday": "祈福",
```
williewillus: Advancement名字都是V家的歌名，我印象中好像BLESSING通常不會翻成「祈福」，而是直接用英文寫BLESSING

例如：https://www.bilibili.com/video/BV1JE411X7sV/

williewillus (commented):
Looks good from a quick skim, just one note on the advancement name

Xplat/src/main/resources/assets/botania/lang/zh_cn.json
Dawnwalker666: ok，我没想到这里用了梗meme，还以为只是单纯对小土豆生日的祝福

Dawnwalker666 (commented):
OK

PR #4380, opened 2023-05-27
Add redstone root to Dandelifeon recipe

zacharybarbanell:
I didn't include any lexica changes - the Dandelifeon's page already mentions it reacts to redstone signals.  The redstone root description is still in the funcitonal flowers section, but I think it's probably fine.


TheRealWormbo:
Merged manually

PR #4382, opened 2023-05-28
Fix description of Efficiency Lens

TheRealWormbo:
The lens actually increases the time to mana loss a bit instead of decreasing it. Its main feature probably is the significantly reduced mana loss rate, though.


williewillus:
Is it the documentation or the code that is wrong? Can you dig the code history and see if it was always like this or if a bug happened? If a bug was introduced to the code, we should fix the code instead.

TheRealWormbo:
I found a commit from [January 24, 2015](https://github.com/VazkiiMods/Botania/commit/2d33b3d73c777735cd9131dfd462fffacb71ec01#diff-3fd4fde09ebba860d86139f221095eb24dffc2c80a5bf31c0e33f013ee349090) where the same values as now are used.

[edit]
Found the origin of today's values: [A commit from over 9 years ago.](https://github.com/VazkiiMods/Botania/commit/b422b7b82a14d55f4cfe2da8e758ec37a5389c3c#diff-74641b31bfa2430fa1f3acbf6cfcc1c75fc558de29e3623612ef4e16adf93e5bR132)

williewillus:
Ok for now in order to line up the doc with the behavior

PR #4386, opened 2023-05-29
New texture for pool minecart item

zacharybarbanell:
to match the vanilla one


artemisSystem:
I will redo this texture with an updated mana pool texture too, once i get to this item in the retexture

PR #4387, opened 2023-05-30
Minecart with Manastorm Charge

zacharybarbanell:
An implementation of minecarts with manastorm charges (similar to minecarts with TNT).  This includes an API component for entities which do something when hit by a mana burst.


zacharybarbanell:
Currently missing from this branch - the crafting recipe for the Minecart with Manastorm Charge exists in code but not in generated/data.  Data generation does not work at the moment, this will be fixed when possible.

artemisSystem:
I don't think this is something we want to have, closing

PR #4391, opened 2023-05-31
Update zh_cn.json

Dawnwalker666:
As of the Commits on May 30, 2023.
The commit is getting more and more, and it seems that it will prevent you from comparing. But I've searched for a long time and can't find how to squash commits on GitHub.
Or maybe these long commits are only visible to me locally. Otherwise I can only delete the local repository and fork it again.
When I see a new commit, I will try to update and fix the Chinese file (and then make the commit longer and longer). Maybe it's better to wait until the next botania update before merge.


williewillus:
you can rebase and squash locally using `git rebase -i <public ancestor>` and selecting "squash" on all but the first commit that comes up https://git-scm.com/docs/git-rebase#_interactive_mode

williewillus:
I can also squash on land, so don't worry about it too much.

Dawnwalker666:
Also I don't know why "All checks have failed" every time. Obviously I only uploaded language file.json. But it shouldn't affect the final review... right? Or is there a character in here that I need to check?

Dawnwalker666:
Awesome, I just found out that botania is going to be upgraded to 1.20. So I created a new fork and a new pull request to avoid committing to 1.20 using the 1.19 branch. I hope this is the right thing to do, although I will only change one file.

PR #4392, opened 2023-06-01
Only create Ender Air Bottles when aiming at nothing

### LINKED ISSUE #4273: Ender Air Bottling Interaction blocks bottle block interactions in the End

unilock:
Some mods add features that rely on using empty Glass Bottles on *blocks*, not just fluids.

This commit allows that to happen by only creating Ender Air Bottles when the player is aiming at *nothing* - not a block *or* a fluid.

PR #4327 implemented something similar, but would only check for fluids; Ender Air Bottles would still be created if looking at any block.

Fixes #4273.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/material/EnderAirItem.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/material/EnderAirItem.java
@@ -8,7 +8,6 @@
  */
 package vazkii.botania.common.item.material;
 
-import net.minecraft.core.BlockPos;
 import net.minecraft.core.particles.ParticleTypes;
 import net.minecraft.sounds.SoundEvents;
 import net.minecraft.sounds.SoundSource;
@@ -48,7 +47,7 @@ public class EnderAirItem extends Item {
 			return InteractionResultHolder.pass(stack);
 		}
 
-		if ((world.dimension() == Level.END && isClearFromDragonBreath(world, player.getBoundingBox().inflate(3.5)) && notAimingAtFluid(world, player))
+		if ((world.dimension() == Level.END && isClearFromDragonBreath(world, player.getBoundingBox().inflate(3.5)) && aimingAtNothing(world, player))
 				|| pickupFromEntity(world, player.getBoundingBox().inflate(1.0))) {
 
 			if (!world.isClientSide) {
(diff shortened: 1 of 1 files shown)


artemisSystem:
Thanks for contributing!

The more i've been thinking about this issue, the more i feel like we should just have a dedicated empty ender air bottle item, and not use vanilla's glass bottles. However, it would be a bit annoying to add an item just for this, so i'm actually unsure. @williewillus, any thoughts on this?

williewillus:
artemisSystem ***@***.***> writes:

> The more i've been thinking about this issue, the more i feel like we should just have a dedicated empty ender air bottle item, and not use vanilla's glass
> bottles. However, it would be a bit annoying to add an item just for this, so i'm actually unsure. @williewillus, any thoughts on this?
>

I'd be fine with the idea of using the Botania brew vials/flasks for
this. A new dedicated item seems a bit much.

artemisSystem:
I was also thinking about those. i'm assuming your thought then was also that vials and flasks could be filled with 4 and 6 ender air each (if using them once in the end turned them into an ender air item, that'd be quite expensive, no?) but then, how do you throw them? if you're collecting them in a dispenser, should it fill until it's full, then throw if it's already full? Would the remainder be annoying to work with wrt. autocrafting? much to consider

800020h:
mana glass is cheap enough for 1 vial to be 1 ender air

TheRealWormbo:
This PR will probably be made obsolete by a new way to obtain ender air in 1.21, involving managlass vials.

DaFuqs:
(you could also additionally merge this one for 1.20.1, since it was this PRs birthday recently :>)

TheRealWormbo:
We will not be changing this for 1.20 anymore. 1.21 Botania will feature a completely new implementation that uses a different base item. (see #5040)

PR #4425, opened 2023-07-16
Update Dandelifeon to work twice as fast on enchanted soil

NEstoll:


Alwinfy:
Not so simple; this would desynchronize adjacent Dandelifeons clocked at different rates. You'd need to track the source rate of the parent Dandelifeon, and probably disallow cross-rate Dandelifeons from interacting.

NEstoll:
It wouldn't desync (every other tick would still be synced), but there would be ticks where only overgrowthed Dandelifeons would be active. However, there isn't a way to avoid that (other than just wasting the overgrowth seed and not having it affect the Dandelifeon). This seemed like the most intuitive way to double the speed, and even allows for some cool setups where the difference in rate can be used to make otherwise infeasible patterns.

PR #4431, opened 2023-07-30
[1.20.x] Make `EntityHelper#syncItem` work without setting item to empty

frank89722:
Hello there, 

The current behavior of `EntityHelper#syncItem` could create mod conflict with the mod that hooked on `ItemEntity#setItem`.
This can be done by setting it to a copy of the stack as well.

Similar behavior can be found in `ItemEntity#merge(ItemStack, ItemStack, int)`.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/helper/EntityHelper.java
+++ b/Xplat/src/main/java/vazkii/botania/common/helper/EntityHelper.java
@@ -16,13 +16,10 @@ public class EntityHelper {
 	 * Forces an item entity to resync its item to the client.
 	 *
 	 * Entity data only resyncs when the old/new values are not Object.equals().
-	 * Since stacks do not implement equals, and we're not changing the actual stack object,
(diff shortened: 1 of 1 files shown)


williewillus:
This is fine, but I'm wondering why setting to an empty item causes issues here?
Generally, my understanding from vanilla's convention is that things that happen within the same tick (like this sort of "swap something out and back in") should be undetectable. Doing something like additional auto stacking should be done on a timer like how vanilla's stacking works.

Overall this is ok but it causes an extra stack allocation every call, which is slower than it should be on Forge (because of capability initialization)

frank89722:
> This is fine, but I'm wondering why setting to an empty item causes issues here?

This could affect mods that do specific things when the entity got set to an empty item. 

But since you mentioned it could cause worse performance in Forge and I think I didn’t consider thoroughly enough before sending this PR so I'll close it for now.

PR #4437, opened 2023-08-17
Fix creative tab title

extclp:


williewillus:
Fix in what way? I can't tell what is supposed to happen just from the code change

williewillus:
Hmm, we should probably remove the English text from the texture and find a way to render the native text in white, but this will work for now. Thanks.

williewillus:
Does forge need this fix too?

williewillus:
See my previous comment, does this affect forge too?

PR #4443, opened 2023-08-26
Fix assignment wtih Charms

Shibva:
Corrects a slight error wtih charms being put in the necklace slot when it shouldn't (on the basis that curio's have their own slot for charms)

makes it close to "parallel" features present in Curios

- Creates a Charm slot located in the main hand section
- Created a tag under hand/charm
- Migrate Diva Charm and Goddess Charm from necklace tag to charm tag

### DIFF
--- a/Fabric/src/generated/resources/data/trinkets/tags/items/chest/necklace.json
+++ b/Fabric/src/generated/resources/data/trinkets/tags/items/chest/necklace.json
@@ -3,8 +3,6 @@
   "values": [
     "botania:blood_pendant",
     "botania:cloud_pendant",
-    "botania:diva_charm",
-    "botania:goddess_charm",
     "botania:ice_pendant",
     "botania:lava_pendant",
     "botania:super_cloud_pendant",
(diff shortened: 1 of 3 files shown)


williewillus:
You need to do the change in the datagen instead of manually editing files in the `generated` folder

williewillus:
Also, please target 1.20, I'll pick back to 1.19.2

williewillus (changes requested):
as above

Shibva:
Noted

I'll do that when I have the chance

Shibva:
> You need to do the change in the datagen instead of manually editing files in the `generated` folder

Can you clarify what you mean by datagen? I'm not familular with this workspace

williewillus:
Change the code somewhere here https://github.com/VazkiiMods/Botania/blob/1.20.x/Fabric/src/main/java/vazkii/botania/fabric/data/FabricItemTagProvider.java#L63

then run the `:Fabric:runFabricDatagen` Gradle task to regenerate the json files

Shibva:
closing to make a proper pull merge

Issue #3815, opened 2021-09-08
[Immersive Engineering] Agricarnation does not accelerate growth of industrial hemp


dalapo:
Forge version:  36.2.2
Botania version:  1.16.5-419
IE version: 1.16.5-5.0.3-138

What I expected to happen: Agricarnation accelerates industrial hemp as it does Vanilla crops.

What happened instead: Agricarnation has no effect on industrial hemp.

Looking at the code, it seems to be because industrial hemp extends BushBlock, which the agricarnation explicitly blacklists. Unsure whether or not this is intended behaviour; however, I would expect it to work as it is possible to grow hemp with bone meal.

williewillus:
yeah it's a bit hard to judge with these. BushBlock is a really broad class that everyone vanilla or modded uses for lots of different things, so just letting the agri tick all of them isn't a good idea. Not sure of a good solution.

PR #4446, opened 2023-08-27
Config for agricarnation white list

rqy2002:
Add config white list for agricarnation, to solve issues like "why agricarnations don't work on something" (such as #3815).

(although we can't judge which block is actually "a crop", players or modpack authors can control it by such a config)


rqy2002:
(should we test isValidBonemealTarget then?)

williewillus:
why not just a tag?

rqy2002:
You are right. I'm just not familier to modding 😆

TheRealWormbo:
Is this PR still being worked on?

PR #4447, opened 2023-08-28
1.20 adding charms in charm slots (Trinkets "fix")

Shibva:
Add hand/charm and moved 	``divaCharm`` and ``goddessCharm`` to section (hopefuly I did this right)

Creates a new charm slot in the mainland section 

@williewillus NOTE: needs to be run like you said; I'm still fairly new to all of these things so I'm not sure how to go about running it. 

I also want to give the trinket case the needed trinket compatibility since it is missing as Curios has it but it is outside of my ability to implement such

Hopefully this will generate the other files as needed

### DIFF
--- a/Fabric/src/generated/resources/.cache/5e449fe289648869bd1f4d4d99715c8b4e0c057d
+++ b/Fabric/src/generated/resources/.cache/5e449fe289648869bd1f4d4d99715c8b4e0c057d
@@ -11,7 +11,8 @@ c3d9bfa84d88ef5deea5ad1f9b37fafb31297dad data/c/tags/items/wooden_chests.json
 94f96c69b63e660269e97982b78d7e6f54374c08 data/modern_industrialization/tags/items/replicator_blacklist.json
 bb7e6831696137af69ce475bbf164dd0cd8babf3 data/trinkets/tags/items/all.json
 eea867f5a1dcf993821e804195af53f5789ce613 data/trinkets/tags/items/chest/cape.json
-97a18698d3d4e04638d4787a638e33b77b450f25 data/trinkets/tags/items/chest/necklace.json
+20e51fa1c09698356074d3b37a61e3f8cb5eb8ff data/trinkets/tags/items/chest/necklace.json
+378361e248d598a38b7dfcc70a91e4a73898e129 data/trinkets/tags/items/hand/charm.json
 256e381daa579a2e6c23cc0897ac161120b2d936 data/trinkets/tags/items/hand/ring.json
 b1f83cb846b4c3d1578aba6ed515b5ad08aaa126 data/trinkets/tags/items/head/face.json
 bf5357b621fdcc974b60e76c82ec467fc0023268 data/trinkets/tags/items/head/hat.json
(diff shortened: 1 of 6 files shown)


Shibva:
also needs backport to 1.19.2 as a reminder :D

williewillus:
You need to run `./gradlew :Fabric:runFabricDatagen` (`./gradlew.bat :Fabric:runFabricDatagen` on windows) and commit the changes it generates.

williewillus:
Also, you're missing this change from the other PR. Is that intentional? https://github.com/VazkiiMods/Botania/pull/4443/files#diff-478bd022a79d7cb1ddd31a5f0a03f5d5ca5446fafcfe9f4abf6fc5b7e06f20f8

That part should be done manually because it's not in the `generated` folder

Shibva:
Please review this before approving it; I'll not lie when I say I only have a fundamental understanding of how the datapack aspect of this works. I will admit I do not know how to code with Java but I feel that my understanding of Minecraft's datapack system that is integrated with mods is solid enough for me to contribute in those areas. Thanks for the instructions on the datagen task.

I would also work in Trinket Case compact with trinkets but that is outside of my range of possibility

williewillus:
The code looks good to me. Did you run the game and see if it worked ingame? If so I can land this.

Shibva:
game boots but its not creatin the charm slot for some reason

Shibva:
Ok I see what needs to be done now

I'll need to add a custom slot in the same place entities are located;as in add a new file lel

I'll finish that up tomorrow when I have the chance

Shibva:
Thankfully, in the source, there's a test mod that shows me what to do

Shibva:
Ok, one look at another source later and a quick grab to make things quicker and its done.

ran it again and it now all works with the slot added and everything @williewillus

williewillus:
Hm. It doesn't seem right that we have jsons in trinkets' namespace. Let me check the docs about how this is supposed to work.

Shibva:
It should be the case; based on another mod I looked at with open source, it appears to be in the right spot

unilock:
It's important to mention that the "charm" slot *does not exist* in the Trinkets API by default, although the texture for it does.

As such, the tooltip for any item that can be put in that slot will be missing a translation key; thus the following entry should be added to the lang file(s):
```
"trinkets.slot.hand.charm": "Charm"
```

For what it's worth, it seems that the most popular mod that implements a "charm" slot, [Charm of Undying](https://github.com/illusivesoulworks/charmofundying), defines the slot as `"charm/charm"`, not `"hand/charm"`. ([source](https://github.com/illusivesoulworks/charmofundying/blob/0e1f38d751b2915acb557c45c74b1c52f7d24827/fabric/src/main/resources/data/trinkets/slots/charm/charm.json))

Shibva:
That is true; this is more or less a fallback in case non-such are present; the code should work, and I have tested it in the space.

Shibva:
I was told to make this to be applied for 1.20 and then retroactively for 1.19.2

TheRealWormbo:
This is worth revisiting after the 1.21 port, when hopefully the new "baubles wars" on NeoForge show hints of a winner.

Shibva:
Feel free to make the changes for it to work then as well, this was more of a nitpick fix for trinkets mod

Shibva:
Renamed this; this is probably why it has seen no movement lel

Issue #4415, opened 2023-07-02
Marimorphosis does not convert tuff


TheRealWormbo:
### Mod Loader

Both Fabric and Forge (I confirm that I have tested both loaders and will specify both loader versions below)

### Minecraft Version

1.19.2

### Botania version

1.19.2-440

### Modloader version

(any)

### Modpack info

(none)

### The latest.log file

(n/a)

### Issue description

The block tag `botania:marimorphosis_convertable` contains natural stone, natural deepslate, as well as the stone variants andesite, granite, and diorite. It seems only logical that tuff would be convertible as well, but it's not mentioned in the block tag definition.

### Other information

According to discussion on Discord, the conversion should include tuff as well.

TheRealWormbo:
@williewillus It looks like the changelog got its addition in the wrong section, and datagen might need to run as well.

williewillus:
Datagen is clean, there's a CI signal for it.

PR #4448, opened 2023-09-11
Allow Marimorphosis to convert tuff

PHIQW:
Fixes issue #4415
added tuff to the marimorphosis_convertable tag list
added this change to upcoming changelog

### DIFF
--- a/Xplat/src/generated/resources/data/botania/tags/blocks/marimorphosis_convertable.json
+++ b/Xplat/src/generated/resources/data/botania/tags/blocks/marimorphosis_convertable.json
@@ -5,6 +5,7 @@
     "minecraft:deepslate",
     "minecraft:granite",
     "minecraft:diorite",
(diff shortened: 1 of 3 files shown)


williewillus:
You should not be editing files in `generated` folder by hand. Edit the data generator here

https://github.com/VazkiiMods/Botania/blob/1.20.x/Xplat/src/main/java/vazkii/botania/data/BlockTagProvider.java#L193

then run `./gradlew :Fabric:runXplatDatagen`

williewillus (changes requested):
as above

PHIQW:
Thanks for letting me know! Changes to the data generator added.

After reading the tag provider, Tuff should also be considered for ENDER_AIR_CONVERTABLE since deepslate is similarly convertible. If it is, I'm not sure if it should be handled here or another pull request.

PR #4454, opened 2023-09-27
Tag support for Drum of the Gathering

### LINKED ISSUE #4453: Drum of the Gathering can milk baby cows, but not goats, and does not handle suspicious stew from brown mooshrooms
### Mod Loader

### Minecraft Version

1.17+

### Botania version

1.19.2-440

### Modloader version

(any)

### Modpack info

_No response_

### The latest.log file

(n/a)

### Issue description

Code analysis of the `DrumBlock` revealed some deficits in the handling of animals by the Drum of the Gathering:

- Cows do not need to be adults to be eligible for milking.
- Brown mooshrooms have no special handling, so milking them with a bowl always only produces mushroom stew. If under the effect of a flower previously fed to the mooshroom, it should produce suspicious stew instead.
- Goats are ignored for milking, as only entities extending `net.minecraft.world.entity.animal.Cow` can be milked. There are also modded animals (e.g. yaks) where both shearing and milking could make sense. (But while mooshrooms are both milkable and shearable, any code changes should still only consider them for milking to prevent reverting them to regular cows.)

### Steps to reproduce

_No response_

### Other information

Proposed changes:

- Keep special handling for filling bowls from mooshrooms and extend it to support suspicious stew.
- Generalize milking with an entity tag (e.g. `botania:gathering_drum_milkable`), which includes cows, goats, red mooshrooms and brown mooshrooms by default.
- Define a Shearable entity exclusion tag (e.g. `botania:gathering_drum_no_shearing`), which prevents shearing red and brown mooshrooms by default.

TheRealWormbo:
The drum now uses entity tags to determine which mobs can fill milk buckets and which shearable mobs can NOT be sheared: `botania:drum/milkable` (defaults to cow, mooshroom, and goat) and `botania:drum/no_shearing` (defaults to only mooshroom)
Additionally, if a brown mooshroom was fed a flower, the Drum of the Gathering will fill the first bowl for that mooshroom with the appropriate type of suspicious stew. Any additional bowls will be filled with mushroom stew as usual.
And finally, baby animals are excluded from filling buckets or bowls. (Whether they can be sheared is already determined by the mob's shearing logic, which usually excludes babies as well.)

(closes #4453)

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/mana/DrumBlock.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/mana/DrumBlock.java
@@ -11,15 +11,16 @@ package vazkii.botania.common.block.mana;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.particles.ParticleTypes;
 import net.minecraft.sounds.SoundSource;
+import net.minecraft.world.effect.MobEffect;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.Mob;
 import net.minecraft.world.entity.Shearable;
-import net.minecraft.world.entity.animal.Cow;
 import net.minecraft.world.entity.animal.MushroomCow;
 import net.minecraft.world.entity.item.ItemEntity;
 import net.minecraft.world.item.Item;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.item.Items;
+import net.minecraft.world.item.SuspiciousStewItem;
 import net.minecraft.world.level.BlockGetter;
 import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.Block;
@@ -37,12 +38,17 @@ import vazkii.botania.common.block.flower.functional.BergamuteBlockEntity;
 import vazkii.botania.common.handler.BotaniaSounds;
 import vazkii.botania.common.item.BotaniaItems;
 import vazkii.botania.common.item.HornItem;
+import vazkii.botania.common.lib.BotaniaTags;
+import vazkii.botania.mixin.MushroomCowAccessor;
 
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
 
 public class DrumBlock extends BotaniaWaterloggedBlock {
+
+	public static final int MAX_NUM_SHEARED = 4;
+	public static final int GATHER_RANGE = 10;
 
 	public enum Variant {
 		WILD,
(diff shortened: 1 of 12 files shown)


Xplat/src/main/java/vazkii/botania/common/block/mana/DrumBlock.java
```diff
@@ -64,26 +70,90 @@ public VoxelShape getShape(BlockState state, BlockGetter world, BlockPos pos, Co
...
+		Level world = mushroomCow.level();
+		List<ItemEntity> items = world.getEntitiesOfClass(ItemEntity.class, mushroomCow.getBoundingBox(),
+				itemEntity -> itemEntity.getItem().is(Items.BOWL) && !itemEntity.getItem().isEmpty());
+		for (ItemEntity item : items) {
+			ItemStack itemstack = item.getItem();
+			ItemStack stewItem = new ItemStack(Items.SUSPICIOUS_STEW);
+			SuspiciousStewItem.saveMobEffect(stewItem, effect, effectDuration);
+			spawnItem(mushroomCow, stewItem);
+
+			itemstack.shrink(1);
```
williewillus: Use EntityHelper.shrinkItem which handles proper resyncing instead of calling bare shrink()

technically the callsite above should be using this helper too, but that loop always consumes the entire stack until it's gone so it's not strictly necessary

Xplat/src/main/java/vazkii/botania/common/block/mana/DrumBlock.java
```diff
@@ -64,26 +70,90 @@ public VoxelShape getShape(BlockState state, BlockGetter world, BlockPos pos, Co
...
+	private static void fillBowlSuspiciously(MushroomCow mushroomCow) {
+		MushroomCowAccessor mushroomCowAccessor = (MushroomCowAccessor) mushroomCow;
+		MobEffect effect = mushroomCowAccessor.getEffect();
+		int effectDuration = mushroomCowAccessor.getEffectDuration();
+
+		Level world = mushroomCow.level();
+		List<ItemEntity> items = world.getEntitiesOfClass(ItemEntity.class, mushroomCow.getBoundingBox(),
+				itemEntity -> itemEntity.getItem().is(Items.BOWL) && !itemEntity.getItem().isEmpty());
+		for (ItemEntity item : items) {
+			ItemStack itemstack = item.getItem();
```
williewillus: let's call this something with `bowl` in it so it's clear which is which, it's easy to get mixed up when everything is `item` or `stack` or some combo

williewillus (commented):
couple small things

Xplat/src/main/java/vazkii/botania/common/block/mana/DrumBlock.java
TheRealWormbo: The other one probably wouldn't need to shrink the stack at all if it uses up the entire thing anyway.

PR #4459, opened 2023-10-07
Spark performance improvements

### LINKED ISSUE #4389: Poor spark performance
### Mod Loader

Fabric (actually Quilt but template didnt let me select me + see additional information)

### Minecraft Version

1.19.2

### Botania version

1.19.2-439-FABRIC

### Modloader version

Quilt Loader 0.19.0-beta.13; QFAPI v4.0.0-beta.30 (FAPI 0.76.0 compatible)

### Modpack info

https://github.com/HexxyCraft/modpack

### The latest.log file

https://mclo.gs/HkNGE7M

### Issue description

Sparks cache transfers, to which they try to transfer every tick (except when empty). However, they also check on the validity of those every tick and thus needlessly iterate through them, even if they wouldn't be able to transfer there because they're empty. 

Both recessive and dominant mana sparks search for nearby sparks (and iterate them) each tick, even if they wouldn't need to because they're empty (or full respectively). Dominant sparks then notify one of the other sparks to transfer to them (each tick, this only makes them ramp up in speed initially, after that its instant, so why is this even needed?), recessive sparks always try initiate a transfer to all other valid sparks in reach.
Because the transfer data is cached this searching gets needless once all sparks have been added as transfers, unless any of the sparks changes, or gets added

### Steps to reproduce

Setup a large spark-based mana storage, preferably with lots of connected recessive and dominant sparks
For the profiles I used mana pools, 6 recessive pools, and 6 dominant pools, and 242 normal ones, all within range of each other
With all of them empty: https://spark.lucko.me/CPNYkwU8Pw
With the recessive pools being filled by gaia mana spreaders: https://spark.lucko.me/Fwhsy9qU54

### Other information

Note that while I only tested on the version provided, the code I refer to is still live in the `xplat` source set on `1.19.x`. It should hold regardless of the plattform.

Such a high performance hit is unexpected when nothing visible is happening (everything is empty or filled up.

There also is https://github.com/VazkiiMods/Botania/issues/2428 which has been closed by stalebot.

anonymous123-code:
Fixes #4389 

Sparks already had a lot of the infrastructure needed, but I basically changed spark from recalculating everything every tick to being notified when stuff changes. Note that this approach is more prone to bugs, especially when more funcitonality gets added. The issue also goes a lot more into detail. I also tested using Fabric's block cache API but it didn't turn out to be worth it.

Setup profiled:
(Screenshot TBD)
10 Minutes superflat, no mob spawning, no daylight cycle (In dev)

Runtime of: `vazkii.botania.common.entity.ManaSparkEntity.tick()`
With mana running:
|                       | Fabric | Forge |
| --------------- | ------ | -------|
| Before (ms)    | 37,269 | 92,968 |
| After (ms)       | 19,736 | 46,217 |
| After/Before | 52%      | 49%      |

When empty:
|                       | Fabric | Forge |
| --------------- | ------ | -------|
| Before (ms)    |  24,147 | 50,048 |
| After (ms)       | 3,821 | 9,726 |
| After/Before | 15%          |  19%         |

I already tested a lot of stuff before rebasing (But that one was like 1.19.2), so I'll need to retest.
Tested:
- Fabric
  - [ ] Singleplayer
  - [ ] Server
- Fabric
  - [ ] Singleplayer
  - [ ] Server

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -62,7 +62,7 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 	private final ArrayList<ManaSpark> transfersTowardsSelfToRegister = new ArrayList<>();
 
 	private boolean shouldFilterTransfers = true;
-	private boolean wasFull = true;
+	private boolean receiverWasFull = true;
 	private boolean firstTick = true;
 
 	public ManaSparkEntity(EntityType<ManaSparkEntity> type, Level world) {
@@ -93,6 +93,7 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 			return;
 		}
 
+		// When loaded, initialize transfers
 		if (firstTick) {
 			updateTransfers();
 		}
@@ -171,27 +172,25 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 
 			}
 			case DOMINANT -> {
-				if (wasFull && !receiver.isFull()) {
+				if (receiverWasFull && !receiver.isFull()) {
 					updateTransfers();
 				}
-				if (transfersTowardsSelfToRegister.size() > 0) {
+				if (!transfersTowardsSelfToRegister.isEmpty()) {
 					transfersTowardsSelfToRegister.remove(0).registerTransfer(this);
 				}
 			}
-			case RECESSIVE -> {
-				// updateTransfers();
-			}
+			// Recessive does not need to be handled because recessive sparks get notified in all relevant cases
 			default -> {
-				if (wasFull && !receiver.isFull()) {
+				if (receiverWasFull && !receiver.isFull()) {
 					notifyOthers(getNetwork());
 				}
 			}
 		}
 
 		if (receiver != null) {
-			wasFull = receiver.isFull();
+			receiverWasFull = receiver.isFull();
 		} else {
-			wasFull = true;
+			receiverWasFull = true;
 		}
 
 		if (!transfers.isEmpty()) {
@@ -301,6 +300,7 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 							spawnAtLocation(SparkAugmentItem.getByType(upgrade), 0F);
 							setUpgrade(SparkUpgradeType.NONE);
 
+							// Recalculate transfers, recessive and dominant will register the proper transfers
 							transfers.clear();
 							notifyOthers(getNetwork());
 						} else {
@@ -422,6 +422,7 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 
 	@Override
 	public void setNetwork(DyeColor color) {
+		// The previous network needs to filter this spark out
 		var previousNetwork = getNetwork();
 		super.setNetwork(color);
 		updateTransfers();
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -59,6 +59,8 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 
 	private final Set<ManaSpark> transfers = Collections.newSetFromMap(new WeakHashMap<>());
 
+	private final ArrayList<ManaSpark> transfersTowardsSelfToRegister = new ArrayList<>();
+
 	private boolean shouldFilterTransfers = true;
 	private boolean wasFull = true;
 	private boolean firstTick = true;
@@ -172,6 +174,9 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 				if (wasFull && !receiver.isFull()) {
 					updateTransfers();
 				}
+				if (transfersTowardsSelfToRegister.size() > 0) {
+					transfersTowardsSelfToRegister.remove(0).registerTransfer(this);
+				}
 			}
 			case RECESSIVE -> {
 				// updateTransfers();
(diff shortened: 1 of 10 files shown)


williewillus:
Very nice, could you go through and add inline comments explaining what's being done at a high level? I think that would help review.

Like for example, I see the state/BE params being removed from findManaReceiver, which I'm assuming is to boost perf, but would be nice to have that called out instead of me assuming.

anonymous123-code:
Actually I think that one was a leftover from me implementing caching. I can change it, although that pattern might be common enough that its better to just keep it...
I'll comment the other stuff tho

anonymous123-code:
Stuff should be better now, I left the findManaReceiver refactorings in as that method already existed before the PR.

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
```diff
@@ -163,29 +172,25 @@ public void tick() {
...
-					SparkUpgradeType otherUpgrade = s.getUpgrade();
-					return s == this || otherUpgrade != SparkUpgradeType.NONE || !(s.getAttachedManaReceiver() instanceof ManaPool);
-				});
-				if (!validSparks.isEmpty()) {
-					validSparks.get(level().random.nextInt(validSparks.size())).registerTransfer(this);
+				if (receiverWasFull && !receiver.isFull()) {
+					updateTransfers();
+				}
+				if (!transfersTowardsSelfToRegister.isEmpty()) {
+					transfersTowardsSelfToRegister.remove(0).registerTransfer(this);
```
williewillus: If there's a lot of elements in the list, `remove(0)` is going to be expensive by forcing an arraylist recopy. Maybe you want a Deque instead?

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
```diff
@@ -212,8 +223,35 @@ public void tick() {
...
+			}
+			case DOMINANT -> {
+				List<ManaSpark> validSparks = SparkHelper.getSparksAround(level(), getX(), getY() + (getBbHeight() / 2), getZ(), getNetwork());
+				for (var spark : validSparks) {
+					SparkUpgradeType otherUpgrade = spark.getUpgrade();
+					if (spark != this && otherUpgrade == SparkUpgradeType.NONE && spark.getAttachedManaReceiver() instanceof ManaPool) {
+						transfersTowardsSelfToRegister.add(spark);
+					}
+				}
+				Collections.shuffle(transfersTowardsSelfToRegister);
```
williewillus, line 254: reasoning for the shuffles that were added here?

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
```diff
@@ -59,7 +59,11 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 
 	private final Set<ManaSpark> transfers = Collections.newSetFromMap(new WeakHashMap<>());
```
williewillus: not strictly related to the pr, but I think renaming this to `outgoingTransfers` might make the code clearer

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
```diff
@@ -59,7 +59,11 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
...
 
-	private int removeTransferants = 2;
+	private final ArrayList<ManaSpark> transfersTowardsSelfToRegister = new ArrayList<>();
```
williewillus, line 62: Any risk of sparks hanging around in this list and leaking memory? I'd either use weak references or filter out `remove`d sparks every tick.

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
```diff
@@ -365,12 +416,24 @@ public SparkUpgradeType getUpgrade() {
...
+		super.setNetwork(color);
+		updateTransfers();
+		notifyOthers(color);
+		notifyOthers(previousNetwork);
 	}
 
 	@Override
 	public boolean areIncomingTransfersDone() {
 		if (getAttachedManaReceiver() instanceof ManaPool) {
-			return removeTransferants > 0;
```
williewillus, line 373: tbh I'm not really sure what this old code did lol, do you know and can you explain why it should be removed?

williewillus (commented):
Looks mostly good, some more questions and we should land this for the next beta

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
anonymous123-code: Otherwise mana pools would be loaded in a relatively consistent order, making only some fill up. This makes it so if they're reloaded, that order is shuffled at least sometimes. A similar thing was in here before,  I'm only guessing the reason why it has been implemented: https://github.com/VazkiiMods/Botania/pull/4459/files#diff-6ab7cbbb33c3d206fc8ca670958f924a1353804a7fb8f234f3308622d23723f7L172 
I can also add the transfers directly, but that would be slightly breaking with older behavior (Previosly, and in this implementation dominant sparks would only add one transfer each tick, taking some time to initialize). This would make `transfersTowardsSelfToRegister` unnecessary.

Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
anonymous123-code: RemoveTransferants was the old reset logic when stuff like the upgrade or the network changed. This is now handled differently. And the only reason why it might be removed is if it is full, which is checked seperately, so I changed it to be a constant `false`

williewillus:
let's get it tested in the next beta

PR #4463, opened 2023-10-14
Document fractional radius and sphere mode for Worldshaper's Sextant

### LINKED ISSUE #4398: Sphere mode for Worldshaper's Sextant is not documented
### Mod Loader

### Minecraft Version

1.19.2

### Botania version

1.19.2-440

### Modloader version

(n/a)

### Modpack info

(n/a)

### The latest.log file

(n/a)

### Issue description

The sphere mode functionality and mode switching functionality of the Sextant is not explained in the corresponding Lexica Botania entry. Similarly the new fractional radius control should probably be explained, as it can be used to tweak the appearance of the created circle shape.

### Steps to reproduce

_No response_

### Other information

The current entry for the Sextant has two relatively full pages and the crafting recipe. Considering the possible amount of additional information (sphere mode, how to toggle, and maybe even a hint on how to control the radius selection), more than one additional page might be needed.

TheRealWormbo:
Inserts additional lexicon page mentioning circle shape adjustment option via fractional radius and the way to switch to sphere mode. Also fixes recipe caption showing "(circle mode)" and thus overflowing the page width. (closes #4398)


williewillus:
Wanting to slowly stop using random lang keys like `foo<random number>`, can you name it something like `sextant_modes`?

TheRealWormbo:
Done.

Issue #4388, opened 2023-05-30
Poor Elven Gateway and Mana Enchanter Performance


anonymous123-code:
### Mod Loader

Fabric (actually Quilt but template didnt let me select me + see additional information)

### Minecraft Version

1.19.2

### Botania version

1.19.2-439-FABRIC

### Modloader version

Quilt Loader 0.19.0-beta.13; QFAPI v4.0.0-beta.30 (FAPI 0.76.0 compatible)

### Modpack info

https://github.com/HexxyCraft/modpack

### The latest.log file

https://mclo.gs/HkNGE7M

### Issue description

Both the mana enchanter and the elven gateway (especially the latter) have a lot higher performance cost than expected by the user, which stems from them checking their structure every tick.

The gateway additionally searches for pylons, making it hit especially hard. However these locations are not required to know unless items are traded or during startup. This means that those locations could be cached and only be updated when mana gets consumed. In Addition to that, the structure checking by the portal is completely unnecessary because destruction could be implemented similarly to a netherportal, where blockupdates propagate it. This would reduce it's cost immensely.
Spark profile for the gateway: https://spark.lucko.me/B5IYSSKTgQ?hl=1990

As for the enchanter, I dont have a lot of ideas other than to onlyy check when enchanting or to listen for block destructions(which minecraft might have a system for?)

### Steps to reproduce

Build and initialize a mana enchanter or elven gateway and profile. Only really visible on a large scale such in large servers / in large farms.

### Other information

Note that, while I only tested on the version provided, the code I refer to is still live in the xplat source set on `1.19.x`. It should hold regardless of the plattform.
The main issue is that no player expects these blocks to have such a high cost when unused.

williewillus:
block updates only work for a portal because all the obsidian blocks touch the portal blocks, since these have an airgap we have to check every tick

TheRealWormbo:
I think there might be a couple of ways to improve performance of the Alfheim portal.

1. The block entity doesn't remember which way the portal is facing, so it needs to make a rotation-agnostic multiblock check each tick. Remembering the rotation could help skipping a needless verification attempt if the portal knows its multiblock matches the 90 degree rotation. (The same optimization probably also applies to the mana enchanter.)
[...]

PR #4465, opened 2023-10-20
Optimized mana enchanter multiblock checking

TheRealWormbo:
As per #4388, the enchanter no longer assumes its rotation could change from one tick to another and performs its structure validation at a much reduced rate while it is idle.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/ManaEnchanterBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/ManaEnchanterBlockEntity.java
@@ -77,6 +77,7 @@ public class ManaEnchanterBlockEntity extends BotaniaBlockEntity implements Mana
 	private static final String TAG_ITEM = "item";
 	private static final String TAG_ENCHANTS = "enchantsToApply";
 	private static final int CRAFT_EFFECT_EVENT = 0;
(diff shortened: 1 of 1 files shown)


Xplat/src/main/java/vazkii/botania/common/block/block_entity/ManaEnchanterBlockEntity.java
```diff
@@ -136,6 +137,8 @@ public class ManaEnchanterBlockEntity extends BotaniaBlockEntity implements Mana
 
 	public int stage3EndTicks = 0;
 
+	private transient int idleTicks = 0;
```
williewillus: remove `transient`

Xplat/src/main/java/vazkii/botania/common/block/block_entity/ManaEnchanterBlockEntity.java
```diff
@@ -465,6 +473,14 @@ public static Direction.Axis canEnchanterExist(Level world, BlockPos pos) {
 		};
 	}
 
+	private static Rotation getAxisRotation(Direction.Axis axis) {
+		return switch (axis) {
+			case X -> Rotation.NONE;
+			case Z -> Rotation.CLOCKWISE_90;
+			default -> null;
```
williewillus: you could just throw here and elide the null check on the caller side, this is never going to be Y

PR #4494, opened 2023-11-27
Lexicon can be placed in bookshelf and on lectern

### LINKED ISSUE #4488: Lexicon cannot be placed in Chiselled Bookshelf / on Lectern
### Mod Loader

### Minecraft Version

1.20.1

### Botania version

1.20.1-441

### Modloader version

n/a

### Modpack info

_No response_

### The latest.log file

n/a

### Issue description

The lexicon isn't added to the bookshelf_books or lectern_books item tags, so they can't be placed in the respective blocks. It's a quick fix but I'm not sure where the tags should be added (like part of datagen or just a similar place to where the damage type tags are stored).

### Steps to reproduce

_No response_

### Other information

_No response_

TheRealWormbo:
Fixes #4488 by adding the lexicon item to the two relevant tags. ~~Also needed to hook into the Patchouli lectern event handler to make sure the advancement for reading the lexicon for the first time works via lectern reading as well.~~

### DIFF
--- a/Xplat/src/generated/resources/.cache/5e449fe289648869bd1f4d4d99715c8b4e0c057d
+++ b/Xplat/src/generated/resources/.cache/5e449fe289648869bd1f4d4d99715c8b4e0c057d
@@ -56,10 +56,12 @@ adfc7deaa7aa9d5425da095270745d58605b86ad data/botania/tags/items/rods.json
 70a32077a4c06e2fc3ed8783037d4b6c659216be data/botania/tags/items/terrasteel_ingots.json
 de5556cc02036175ac731ce76827076c6b834d25 data/botania/tags/items/terrasteel_nuggets.json
 43bfa13de3ebc0be372fc48b0e4be56a25cb1893 data/minecraft/tags/items/axes.json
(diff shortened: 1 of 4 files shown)


TheRealWormbo:
Apparently the GitHub build did not like the Patchouli mixin, so no advancement for lectern readers for now.

TheRealWormbo:
Added an event handler for right-clicks on lectern blocks with the Lexica Botania as the book. On Forge it is registered at "high" instead of default priority, so it won't get canceled by Patchouli's event handler. I didn't see a similar mechanism on Fabric, so there's a random chance for Patchouli's handler to run first and prevent Botania's handler for unlocking the advancement.

TheRealWormbo:
A better implementation is on the way: https://github.com/TheRealWormbo/Botania/commit/2c40fe893b562ee30cb74f5104d401046bd721f9
That one uses an [upcoming feature in Patchouli](https://github.com/VazkiiMods/Patchouli/pull/715) to properly trigger the advancement for whatever way the lexicon ends up being opened. I will adjust this PR accordingly once a new Patchouli version is available.

PR #4507, opened 2023-12-04
Rewrote Wand of the Forest generic block manipulation logic

### LINKED ISSUE #3223: Wand of Forest rotating Double Chests or Extended Pistons goes funky.
# Version Information
Forge version:  31.1.25
Botania version:  r11-378.63 but 1.15

# Further Information
Steps to reproduce:
1. Take a Wand of the Forest.
2. Use it to rotate half of a double chest.
3. Observe.
4. Also try this with an extended piston.

What I expected to happen:
The chest not to go all funky.

What happened instead:
Chest Funkiness
Piston Funkiness

TheRealWormbo:
(fixes #3223)

- excluded various blocks from wand interactions entirely via "botania:unwandable" tag
- manipulation logic now ensures that either the new state can "survive" (according to the block state's own logic) or the block state remains unchanged
- added support for RotatedPillarBlock (e.g. logs), which cycles through the possible axis values
- added support for the 16 rotations of standing/ceiling-hanging sign blocks, which are cycled through
- manipulating the side of stair, trapdoor, or single slab blocks toggles their vertical adjustment in the block space
- restricted side property toggling to block states that have all six boolean direction properties (primarily for toggling the face states of huge mushroom blocks)
- improved facing property rotation:
  - excluded block states that represent half of a bed, double chest, or extended piston from all rotation attempts
  - blocks that can be attached to walls, floors, and ceilings, and also rotated horizontally are rotated clockwise around the clicked side
  - if all six directions are supported, rotate clockwise around the clicked side
  - if rotating around the clicked side isn't possible in either of the above cases, flip over to the opposite facing
  - otherwise, if not all six possible directions are supported, iterate to the next valid facing direction
- if neither of the above apply, let the block state handle the rotation, trying clockwise 90 degrees, 180 degrees and counter-clockwise 90 degrees (in that order, whichever is found to be valid first)

### DIFF
--- a/Xplat/src/generated/resources/.cache/bfa01a6ca2555c100103725bf5c9e6da285f29c3
+++ b/Xplat/src/generated/resources/.cache/bfa01a6ca2555c100103725bf5c9e6da285f29c3
@@ -29,7 +29,7 @@ d7dcc983bfdaf90fe776e2be93411250c03b9e7e data/botania/tags/blocks/mystical_flowe
 c2c6198ca6a711bf2d89080a82e564bae5739d3d data/botania/tags/blocks/terra_plate_base.json
 10fd0dc41dc5715e3707917341749645a6c1693b data/botania/tags/blocks/terraformable.json
 2ead7838be1f063018e6e13e8300f3a183adf5f5 data/botania/tags/blocks/terrasteel_blocks.json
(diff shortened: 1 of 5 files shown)


williewillus:
didnt read that much but trust u

Issue #4509, opened 2023-12-09
Slime/honey block moved by Force Relay does not pull block behind it


TheRealWormbo:
### Mod Loader

Both Fabric and Forge (I confirm that I have tested both loaders and will specify both loader versions below)

### Minecraft Version

1.20.1

### Botania version

441 (but also replicated in a 1.18 version)

### Modloader version

(dev snapshot)

### Modpack info

(none)

### The latest.log file

(n/a)

### Issue description

A sticky block, such as slime or honey (but with the corresponding mods also other blocks that form rigid structures for vanilla piston movement, such as chains or double chests), that is moved directly via a Force Relay block does not connect to any blocks on the side it moves away from. That block also will be ignored by any other sticky blocks around it that move with the structure.

### Steps to reproduce

1. Build the following setup: 
![Force relay between two pistons with buttons for activation next to a honey block structure with lime and red glass blocks attached.](https://github.com/VazkiiMods/Botania/assets/7620864/0f57f36b-b826-4987-b513-57cc53bca901)
2. Link the force relay to the honey block between the lime glass blocks.
3. Push the blackstone button. Only the linked honey block and the lime glass blocks move. The other honey block and the red glass stay where they are.
4. Push the stone button. The honey block moves back, but only takes the line glass blocks on its side with it. The lime glass block opposite to the push direction is left behind.

### Other information

As far as I can tell, this is caused by the `PistonStructureResolver` explicitly excluding the disconnected block. This is because that block would be where the pushing piston is located. Similarly, the following structure is not pushable, if the Force Relay is linked to the honey block next to the lime glass block:
![Force relay between two pistons with buttons for activation next to a 3x2 honey block structure with one lime and one red glass block attached](https://github.com/VazkiiMods/Botania/assets/7620864/049f3b5e-0d89-433e-8611-ed55e8c18efc)
The reason is that the honey block next to the red glass is considered as the pushing piston, which is not moveable by the honey structure reaching around to push it from behind, similar to this piston setup:
![A "bracket" of honey blocks enclosing a piston next to a button on red glass](https://github.com/VazkiiMods/Botania/assets/7620864/a8b94c7f-0129-4fe7-b278-2fef70d59af2)

Issue #4411, opened 2023-06-26
interaction logic between force relay and slime block failed


9SENSEI:
### Mod Loader

Forge

### Minecraft Version

1.19.2

### Botania version

1.19.2-440

### Modloader version

Forge:43.2.3

### The latest.log file

(n/a)

### Issue description

behavior of slime block and force relay is not consistent with version 1.16.5

https://github.com/VazkiiMods/Botania/assets/56625750/f4fbcb9e-f5fd-4ee9-8e4d-1a620f5e2fca

TheRealWormbo:
Looks like a duplicate of #3917.

PR #4510, opened 2023-12-09
Improve Force Relay/Lens movement logic for block structures

### LINKED ISSUE #3917: Force Relay bug
### Mod loader

Fabric

### Minecraft version

1.18.1

### Botania version

427

### Modloader version

fabric 0.12.12

### Modpack info

botania + deps only

### The latest.log file

https://pastebin.com/dpWebqPS

### Issue description

Relays do be acting weird
https://discord.com/channels/154261184701267969/801591103375015957/934085441907458068

One thing the video doesn't show is that even after the relay stops working, you can keep pushing it normally till it reaches the pos it should be and then starts working again, just a curiosity

### Steps to reproduce

it quite obvious in the vid how to do it

### Other information

_No response_

TheRealWormbo:
(fixes #3917, #4411, and #4509)

- fix pushing the wrong way if Force Relay is part of a retracting sticky piston's block structure
- don't assume that the location where the force push seemingly originates from contains an unmovable block
- updated/expanded tests to cover movement behavior in (potentially nested) block structures or with sticky pistons


williewillus:
tbh this logic is too complicated for me to understand anymore, so just gonna trust you

Issue #4403, opened 2023-06-14
Corpoera sparks iterate storage views outside transaction 


MattiDragon:
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-440-fabric

### Modloader version

Not relevant 

### The latest.log file

Not available 

### Issue description

The current implementation of transfer api corpoera sparks iterates over the views of the found storage and interacts with each of them in their own transaction. This works in most cases but doesn't work with extended drawers and other mods that modify views at the end of transactions.

Because the slots of drawers in extended drawers must stay in correct order I sort the slots in the drawer storage in an outer close callback (runs after top level transaction completes). This causes issues with botania because that mutates the list that botania is iterating over, causing a CME. This could easily be fixed on your end by wrapping the iteration in an outer transaction and creating nested transaction for each view. You might not even need the nested transactions as it seems like you commit all or none anyways. 

### Steps to reproduce

To reproduce the CME:
1. Install botania (any version with transfer api support) and extended drawers (any version) on fabric
2. Place a quad drawer down
3. Attach a corpoera network to the drawer
4. Attempt to remove items from the network
--> crash

### Other information

Originally reported [here](https://github.com/MattiDragon/ExtendedDrawers/issues/18#issuecomment-1589896893)

I don't have access to my computer right now so I can't actually reproduce this, but I've read the code and believe that the bug is present

PR #4514, opened 2023-12-13
Fix Fabric transfer transaction usage

### LINKED ISSUE #4486: Corporea Crystal Cube causing ticking block entity crash
### Mod Loader

Fabric

### Minecraft Version

1.20.1

### Botania version

1.20.1-441-fabric

### Modloader version

Fabric Loader 0.14.24 + API 0.90.7

### Modpack info

_No response_

### The latest.log file

https://pastebin.com/UuBZSUBV

### Issue description

The title is pretty much it. attempting to measure the amount of any item in a corporea system with a crystal cube causes the game to crash. 

### Steps to reproduce

Place down a master corporea spark on a corporea brick
put a regular corporea spark on a chest
put any item in the chest
place down a corporea crystal cube
set the corporea crystal cube to read whatever item you put in the chest
client should crash basically immediately due to a ticking block entity

### Other information

I first ran into this when I ported my 1.19 world to 1.20.1. several troubleshooting steps later, I created a fresh install, with just Botania, Patchouli, and Trinkets. I made a new creative world and attempted the steps above, which still produced the crash described above. the linked log file is from said creative world. the fully modded client as well as the fresh install ran completely fine right up until I right clicked the corporea crystal cube to tell it what item to read.

TheRealWormbo:
- access simulated extract operations in transaction as well (fixes #4486)
- get rid of the inner transaction, as commit or rollback depends on the spark type and thus is identical for all accesses within the outer transaction (as recommended in #4403)
- ensure to commit the outer transaction after iterating over the slots if extracting via non-creative spark

### DIFF
--- a/Fabric/src/main/java/vazkii/botania/fabric/integration/corporea/FabricTransferCorporeaNode.java
+++ b/Fabric/src/main/java/vazkii/botania/fabric/integration/corporea/FabricTransferCorporeaNode.java
@@ -36,7 +36,7 @@ public class FabricTransferCorporeaNode extends AbstractCorporeaNode {
 	protected List<ItemStack> iterateOverSlots(CorporeaRequest request, boolean doit) {
 		ImmutableList.Builder<ItemStack> builder = ImmutableList.builder();
 
-		try (Transaction outer = Transaction.openOuter()) {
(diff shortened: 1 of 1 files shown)


williewillus:
thank

PR #4521, opened 2023-12-24
Pollidisiac feeds flowers to brown mooshrooms

TheRealWormbo:
Also includes smaller optimizations to the finding and feeding logic:
* don't look for anything if out of mana
* exclude items from the search that can't be interacted with
* don't look for animals when no items are around
* exclude animals from the search that would not be fed anyway because they are a baby, or already in the mood and not a brown mooshroom

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/flower/functional/PollidisiacBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/flower/functional/PollidisiacBlockEntity.java
@@ -9,9 +9,14 @@
 package vazkii.botania.common.block.flower.functional;
 
 import net.minecraft.core.BlockPos;
+import net.minecraft.sounds.SoundEvents;
+import net.minecraft.tags.ItemTags;
 import net.minecraft.world.entity.EntityEvent;
 import net.minecraft.world.entity.animal.Animal;
+import net.minecraft.world.entity.animal.MushroomCow;
 import net.minecraft.world.entity.item.ItemEntity;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.block.SuspiciousEffectHolder;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraft.world.phys.AABB;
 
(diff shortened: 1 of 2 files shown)


Xplat/src/main/java/vazkii/botania/common/block/flower/functional/PollidisiacBlockEntity.java
```diff
@@ -35,37 +43,75 @@ public PollidisiacBlockEntity(BlockPos pos, BlockState state) {
...
+			var bounds = new AABB(getEffectivePos()).inflate(RANGE);
+			List<Animal> animals = getLevel().getEntitiesOfClass(Animal.class, bounds, Predicate.not(Animal::isBaby));
+			Collections.shuffle(animals);
 
 			for (Animal animal : animals) {
-				if (getMana() < MANA_COST) {
-					break;
+				// Note: Empty item stacks are implicitly excluded in Animal::isFood and ItemStack::is(TagKey)
+				if (!animal.isInLove()) {
+					for (ItemEntity item : items) {
```
artemisSystem, line 60: I'm not quite sure this works as intended. Won't this consume multiple items on one animal? There's nothing that breaks out of the loop once a food item has been consumed from what i can see.

artemisSystem (changes requested):
Left a comment inline, as i don't think the code is quite sound.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/PollidisiacBlockEntity.java
TheRealWormbo: Good point, that item loop was definitely missing a `break`.

artemisSystem:
I didn't test this, but i assume you did, and the code looks fine now. Thanks!

TheRealWormbo:
> I didn't test this, but i assume you did, and the code looks fine now. Thanks!

I did, only missing the multiple item entities case earlier.

artemisSystem:
Wonderful!

PR #4524, opened 2023-12-29
Botania Shader Compatibility for Particles

blagdorfinguy:
When a shader pack uses GLSL deferred rendering, particles with transparency are not rendered correctly. This change (reflected in comments on the config file) will change the rendering method when shaders are set to false in the config, so particles render opaquely (albeit not quite as pretty as they render as is, with shaders they would not have rendered at all, so I feel this change is beneficial but not perfect)


williewillus:
I think overloading this on the existing `useShaders` option is confusing and makes the call site code confusing. Can you make a new config instead like `opaqueParticles`?

williewillus:
Also, you didn't implement this on Fabric

blagdorfinguy:
Unfamiliar with Fabric, but WIP

blagdorfinguy:
Finished fabric and adjusted config. closing request and resubmitting

Issue #4297, opened 2023-02-17
Particals are invisible 


produxe:
### Mod Loader

Fabric

### Minecraft Version

1.18.1

### Botania version

1.18.1-429

### Modloader version

Fabric 0.13.3  Fabric Api 0.46.6

### Modpack info

https://www.curseforge.com/minecraft/modpacks/infinity-fabric

### The latest.log file

dont know how

### Issue description

Cant see any particals from mana and cant see the center of the pylons. Ive checked the config and it seems to be enabled

### Steps to reproduce

Download infinity mod pack to server and see

Pultex:
shaders of any sort do this I had to turn them off

produxe:
Yes that helped partly, i can see sparkes and pylons now but i still cant see the visual transfer from then mana spreaders

blagdorfinguy:
Bump.

Present in 1.20.1, Botania version 1.20.1-441

Disabling shaders in config helps with pylons (see #3951) but does not change the rendering of particles with alpha channels.
Running Oculus 1.6.13 and Embeddium 0.2.13

PR #4525, opened 2023-12-29
Forge + Fabric Opaque Particles for Shader Compat

blagdorfinguy:
When a shader pack uses GLSL deferred rendering, particles with transparency are not rendered correctly. 

This addition to the config file will change the rendering method when opaqueParticles is set to true in the config, so particles render opaquely (albeit not quite as pretty as they render as is, with shaders they would not have rendered at all, so I feel this change is beneficial but not perfect)


Fabric/src/main/java/vazkii/botania/fabric/FiberBotaniaConfig.java
```diff
@@ -252,6 +257,12 @@ public boolean splashesEnabled() {
 		public boolean useShaders() {
 			return this.useShaders.getValue();
 		}
+
+		@Override
+		public boolean opaqueParticles() {
+			return this.useShaders.getValue();
```
williewillus: you're reading the wrong value

Xplat/src/main/java/vazkii/botania/client/fx/FXWisp.java
```diff
@@ -104,14 +106,28 @@ public void setGravity(float value) {
...
 		AbstractTexture tex = textureManager.getTexture(TextureAtlas.LOCATION_PARTICLES);
 		ClientXplatAbstractions.INSTANCE.setFilterSave(tex, true, false);
 		bufferBuilder.begin(VertexFormat.Mode.QUADS, DefaultVertexFormat.PARTICLE);
+
+		//---
+//		RenderSystem.disableBlend();
+//		RenderSystem.depthMask(true);
+//		RenderSystem.setShader(GameRenderer::getParticleShader);
+//		RenderSystem.setShaderTexture(0, TextureAtlas.LOCATION_PARTICLES);
+//		$$0.begin(VertexFormat.Mode.QUADS, DefaultVertexFormat.PARTICLE);
```
williewillus: remove this. if it's vanilla code you referenced, just put a `// [VanillaCopy] <where in vanilla it came from>` comment on top of the code that was copied.

blagdorfinguy:
Should be resolved

williewillus:
did you push your changes? The comments I left still seem relevant on the latest version of your PR

blagdorfinguy:
Sorry, baby's first git bash :)

blagdorfinguy:
Supplemental info- intended to fix #4297 , btw- my comment there is similarly relevant; disabling botania's shaders decrease aesthetic detail, but increase compatibility with certain shaders. This fork is intended to extend that concept to include OpenGL deferred shaders as well

image
For review - if someone is taking a look and doesn't want to build just to see the effects of this PR this screenshot shows a full range of effects with shaders on/off and the new setting on/off. When the setting is off, there is no change to the "look" of the mod, and the new setting is disabled by default, with a relevant comment in the config to indicate visuals are impacted if shaders are not installed or used- 'for compatibility issues only'

_Note: This screenshot was made while the config settings for shader compat and particle compat were combined, but in the current build they are separate options, as noted above_

blagdorfinguy:
image
For review - if someone is taking a look and doesn't want to build just to see the effects of this PR this screenshot shows a full range of effects with shaders on/off and the new setting on/off. When the setting is off, there is no change to the "look" of the mod, and the new setting is disabled by default, with a relevant comment in the config to indicate visuals are impacted if shaders are not installed or used- 'for compatibility issues only'

TheRealWormbo:
There will probably be a bunch of shader-/rendering-related changes in the 1.21 port. This PR seems worth revisiting as part of that.

PR #4529, opened 2023-12-31
Manaweave boots and Snowflake pendant allow walking on powder snow

### LINKED ISSUE #4512: Manaweave boots and Snowflake pendant don't allow Powder Snow walking
### Mod Loader

### Minecraft Version

1.20.1

### Botania version

1.20.1-441

### Modloader version

n/a

### Modpack info

_No response_

### The latest.log file

n/a

### Issue description

With the somewhat new freeze immunity granted by the Snowflake Pendant and Manaweave Boots, it would make sense for wearers to be able to walk on Powder Snow without sinking, however currently you just sink as normal.

Technically the ring of odin also confers freeze immunity but it doesn't really make sense to give it powder snow walkability too.

### Steps to reproduce

_No response_

### Other information

_No response_

TheRealWormbo:
(fixes #4512)


artemisSystem:
Code looks good! I assume this is tested and works, thanks!

PR #4532, opened 2024-01-02
Nerf placing down cellular blocks via Create means

TheRealWormbo:
Block entity NBT is ignored when a cellular block is placed by a non-OP player or by the various automated ways the Create mod has to offer. That means cellular blocks placed in those ways always start at age zero.


TheRealWormbo:
Tested against both the Forge and Fabric versions of the latest Create release for 1.20.1 (0.5.1), attempting (and failing) to maintain the cellular block age while moving it via a contraption and reverting it to its block form in the original location again.

artemisSystem:
Huh, neat, thanks

PR #4542, opened 2024-01-12
Adjust Crystal Cube item HUD overlay

### LINKED ISSUE #4538: Corporea Crystal Cube HUD overlay renders item in wrong position
### Mod Loader

### Minecraft Version

1.20.1

### Botania version

444

### Modloader version

(any)

### Modpack info

(none)

### The latest.log file

(n/a)

### Issue description

The HUD overlay of the Corporea Crystal Cube renders its item icon way below where it should be:

Looking at a Corporea Crystal Cube set to Yellow Mystical Flower such that the HUD overlay is displayed

### Steps to reproduce

Place a Corporea Crystal Cube, then right-click it with any item other than a Wand of the Forest.
The block itself looks fine, but when pointing the crosshair at it, the HUD overlay shows the item hanging off the bottom of the overlay background instead of vertically centered inside the dark box.

### Other information

_No response_

TheRealWormbo:
Fixes #4538 by moving the rendered item up an appropriate amount.


TheRealWormbo:
Looking fine now on both Fabric and Forge:
Unlocked cube HUD overlay
Locked cube HUD overlay

PR #4547, opened 2024-01-13
Fix mana bottle effects

TheRealWormbo:
Fixes <wrong issue linked here>:
- placing water no longer deletes blocks; replaceable blocks are replaced (potentially dropping an item), waterloggable blocks are waterlogged, or the water is potentially placed one block higher
- dropping the entire inventory as the last use of the bottle no longer drops a single-use mana bottle in addition to also giving the player an empty bottle


TheRealWormbo:
(referenced wrong bug)

Issue #4491, opened 2023-11-27
Eye of Flügel and Gaia 2 Dice issues.


TurtleCage:
### Mod Loader

Forge

### Minecraft Version

1.20.1

### Botania version

 Botania-1.20.1-441-FORGE

### Modloader version

Forge 47.2.0

### Modpack info

All The Mods 9

### The latest.log file

https://gist.github.com/TurtleCage/aecced642654d50d32047523b418f89b.js

### Issue description

After defeating Gaia 2 and rolling the dice for Ring of Loki, it provided an Emerald Stoneplate ring from Iron's Spells
It still provided the corresponding achievement.

Eye of the Flügel's main function (warp) works fine but it does not allow for flight with the Tiara after the flight bar runs out.

Unclear if these are ATM or Botania bugs, will report there as well.

Thanks.

### Steps to reproduce

1 - Defeat Gaia 2, use Dice for Ring of Loki.

2 - Attempt Flight with Tiara with Eye of Flügel, mana present in 2 forms - Mirror and Tablet and wait for flight bar to deplete.

### Other information

No extra mods added or removed to the ATM9 pack.

TheRealWormbo:
For reference, the ATM issue can be found here: https://github.com/AllTheMods/ATM-9/issues/894

These are technically two independent bugs, likely with different causes. However, due to the nature of the mod pack (20 pages of included mods on its CurseForge page) it's next to impossible to figure out which of the mods might cause either of those issues. It might even be a customization by ATM itself.

The Dice of Fate check if the player already has certain advancements to determine whether the player has yet to obtain the corresponding item. Once all six relic advancements are obtained, it switches to picking random loot from six possible loot tables that could be customized in a data pack.

As for the Flügel Tiara and Eye of the Flügel, the tiara checks for the presence of the eye in the player's inventory. Also, it only works if the player actually has mana left. Check that the item type is actually `botania:fluegel_eye` (using advanced tooltips via F3+H), otherwise the tiara can't find it.

Danoned:
Tested in vanilla + botania only modpack and I can confirm it's a botania issue. It's a issue with 1.20.1 botania version

anoomolu:
Can confirm the second issue on fabric too, you don't need both mana containers, just one is enough to reproduce

TheRealWormbo:
[...]

PR #4552, opened 2024-01-14
Account for 1.20 changes in Inventory.contains(ItemStack)

### LINKED ISSUE #4530: Check uses of `Inventory::contains(ItemStack)` after logic change since Minecraft 1.19.2
### Mod Loader

### Minecraft Version

1.20.1

### Botania version

441, 442

### Modloader version

(any)

### Modpack info

(none)

### The latest.log file

(n/a)

### Issue description

At some point between 1.19.2 and 1.20.1 the implementation of `net.minecraft.world.entity.player.Inventory#contains(net.minecraft.world.item.ItemStack)` changed from using `ItemStack.isSame()` to `ItemStack.isSameItemSameTags()`. As a result, any check against items with NBT data may now fail if a plain stack of the item to check is passed to the `contains` method.

### Steps to reproduce

_No response_

### Other information

As far as I can tell this is the reason the Flügel Tiara no longer detects the presence of the Eye of the Flügel (see #4491), and it also seems to affect the Mana Flame block.

TheRealWormbo:
Added new helper method to replicate the old behavior and used it where necessary. (i.e. Flügel Tiara check for Eye, and GoG mana flame block; fixes #4530, and partially #4491)

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/decor/ManaFlameBlock.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/decor/ManaFlameBlock.java
@@ -30,6 +30,7 @@ import org.jetbrains.annotations.NotNull;
 import vazkii.botania.client.fx.WispParticleData;
 import vazkii.botania.common.block.BotaniaWaterloggedBlock;
 import vazkii.botania.common.block.block_entity.ManaFlameBlockEntity;
+import vazkii.botania.common.helper.InventoryHelper;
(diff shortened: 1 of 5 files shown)


ghost:
so...and how i fix the error?what im suposed to do with this files...dude, a just want to fly for long periods with my flugel tiara, and im having to turn myself into a programmer to solve it

TheRealWormbo:
You will have to wait for the next Botania release.

PR #4553, opened 2024-01-14
Prevent Animated Torch from emitting signal before its block entity is initialized

### LINKED ISSUE #4489: placing a floating torch in any direction that isn't south will not cause it to power existing redstone components
### Mod Loader

Fabric

### Minecraft Version

1.20.1

### Botania version

1.20.1-441

### Modloader version

Fabric: Loader 0.14.24 API 0.91.0

### Modpack info

_No response_

### The latest.log file

N/A

### Issue description

title

this isn't that important because the floating torch just toggling back to that side will power the component and you can just place the stuff after the torch

### Steps to reproduce

1. place a redstone component
2. place a floating torch facing it that isn't facing south
3. the redstone component will not be powered

### Other information

_No response_

TheRealWormbo:
Fixes #4489, where (on Fabric) the torch initially provides power to the south if placed any other direction. The torch now initially does not output a signal until its block entity is initialized, at which point neighbors are updated accordingly.
This also fixes the torch updating too few block when removed (e.g. if it was powering something through a solid block), and unnecessary updates to unaffected blocks when it starts or finishes rotating.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/AnimatedTorchBlock.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/AnimatedTorchBlock.java
@@ -10,6 +10,7 @@ package vazkii.botania.common.block;
 
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Direction;
+import net.minecraft.server.level.ServerLevel;
 import net.minecraft.world.InteractionHand;
 import net.minecraft.world.InteractionResult;
 import net.minecraft.world.entity.LivingEntity;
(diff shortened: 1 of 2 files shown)


TheRealWormbo:
Tested on both Fabric and Forge to ensure initial placement powers only the intended block, and that updates on rotation and removal only hit the intended blocks.

williewillus:
lgtm, but might be worth refactoring `int side` into just a `@Nullable Direction side` to avoid the untyped int and converting back and forth to Direction. Storing directions as ints is a 1.7-ism

PR #4584, opened 2024-03-22
Remove mod priority configuration option for Orechid outputs

TheRealWormbo:
~~If the configured list is non-empty, the first namespace ID on the list that is matched to blocks in the potential output tag is used to determine the output. If more than one such block is found, the output is randomized amon those blocks. If none of the namespace IDs matches any output block, the result is picked randomly as if no priority mods were configured.~~

Option is removed to prevent confusion, because pack authors should use data pack recipes to customize the Orechid.


TheRealWormbo:
For reference:
This functionality no longer worked since ed5b8602615361207a6f6ec4971d3f732a760979, which appears to have first been released with 1.16.4-415. The configuration option remained in the Forge version ever since, but was not ported to Fabric.

I became aware of it after someone [asked about it on Reddit](https://www.reddit.com/r/botania/comments/13quxwp/questions_about_the_toml_file_and_the_orechid/). While the Orechid is technically fully data-driven, someone (including us) might find this useful at some point, for example if we add back default weights for modded ore tags.

PR #4586, opened 2024-03-22
Updated pool minecart texture to match vanilla

anoomolu:


TheRealWormbo:
I'm not sure this particular texture will make it in, as @artemisSystem mentioned in #4386 that they wanted to redo it with the new mana pool texture.

anoomolu:
Oh okay, I must have missed that one 😭

PR #4619, opened 2024-05-01
Chicken egg timer reduction via Drum of the Gathering

TheRealWormbo:
- if more than 30 seconds remaining to next egg, the excess time is reduced by 50% (e.g. 1 minute remaining becomes 45 seconds remaining)
- if less than 10 seconds remaining, chicken gets startled and lays its egg prematurely
- nothing happens between 10 and 30 seconds remaining


artemisSystem:
Yeah sure

PR #4622, opened 2024-05-01
Update Fabric dependencies

TheRealWormbo:


TheRealWormbo:
Originally this PR also included adding MixinExtras, but it might be a better idea to delay using that until after we switch over to NeoForge, which supports it natively.

PR #4628, opened 2024-05-03
Use phantom ink instead of managlass to make abstruse platforms invisible

TheRealWormbo:
(implements #4279)


TheRealWormbo:
Supersedes PR #4362 for technical and organizational reasons.

PR #4635, opened 2024-05-06
Pasture seeds work more like random grass spread

### LINKED ISSUE #4251: Pasture seeds only consider current Y level
### Mod Loader

Fabric

### Minecraft Version

1.19.2

### Botania version

1.19.2-436

### Modloader version

Quilt 0.17.7-1.19.2

### Modpack info

_No response_

### The latest.log file

https://pastebin.com/xHqhMHzV

### Issue description

When using pasture seeds on grass, they only affect grass blocks on the current Y level, not ones below or above. So they're not that useful on sloped terrain. I feel they should consider nearby Y levels as well, just like normal slow grass growing does.

### Steps to reproduce

_No response_

### Other information

_No response_

TheRealWormbo:
Whether a block can be a source or target of spreading are determined the same way as random grass or mycelium spread. Spread is no longer restricted to one Y-level and considers water or covered blocks properly. (fixes #4251)
Additionally, the replaceable blocks are now defined by a block tag (`#botania:seed_replaceable`), which also includes mycelium by default now.


artemisSystem:
Seems okay, merging this for now. Should we also consider adding all the grasses to the replaceable tag?

Actually, i'm gonna rename it to `#botania:pasture_seed_replaceable`, for a bit more clarity.

artemisSystem:
Had to force push because gradle bitrot. yayyyy

PR #4639, opened 2024-05-10
Loonium with data-driven structure loot

TheRealWormbo:
Implementation for structure detection and structure-specific loot tables and mob spawning pools, customizable via data packs.

- structure-specific loot tables are identified by the structure ID (e.g. for `minecraft:pillager_outpost` the custom Loonium loot table is located at `data/botania/loot_tables/loonium/minecraft/pillager_outpost.json`), with a default loot table corresponding to dungeon loot (located at `data/botania/loot_tables/loonium/default.json`)
  - default loot tables are created with Datagen
  - loot tables usually just reference original chest or entity loot tables related to each structure, so Loonium loot can be modified separately from the loot found in those structures
- structure-specific Loonium configurations are identified the same way (e.g. for `minecraft:pillager_outpost` the custom configuration is located at `data/minecraft/config/loonium/pillager_outpost.json`) with a default configuration (located at `data/botania/config/loonium/default.json`) acting as fallback
  - default configurations are created via Datagen
  - configurations can reference a parent configuration and only override individual aspects of the referenced configuration
  - configurations consist of mana cost to spawn a mob, how many mobs of spawnable types are allowed around the flower, whether the Loonium must be placed in the overall bounding box or an individual piece of the structure, the weighted list of mobs to spawn, a list of potion effects to apply to spawned mobs, and a list of attribute modifiers to apply to the mobs
    - attribute and effect lists can be overridden for individual mobs
    - custom NBT data can be defined for individual mobs (used to make a Creeper spawn charged)
    - equipment loot tables can be specified for each spawnable mob (this will be a vanilla feature in later Minecraft versions, although the implementation here is slightly tweaked from what Trial Chambers use)
- expanded wand HUD to indicate whether a Loonium generates structure-specific loot
- added lexicon mention of the structure-specific behavior
- fixed the mob location selection logic to prevent mobs from immediately starting to suffocate
- added a new challenge advancement for killing every mob type in the structure configurations added with this change
- mobs spawned by the Loonium are put onto a unique team so they don't just start attacking each other (because there are zoglins in some of the mob spawn pools; Heisei Dream may need a future upgrade to support overriding teams allegiances)


TheRealWormbo:
Supersedes #4313, but is only loosely based on that PR.

artemisSystem:
Writing this comment as i'm going through the review:

> for `minecraft:pillager_outpost` the custom configuration is located at `data/minecraft/config/loonium/pillager_outpost.json`  

Why not `data/botania/config/loonium/minecraft/pillager_outpost.json`? The current location doesn't actually specify botania anywhere (only botania thing is the loonium folder), which maybe doesn't matter that much, but seems off to me at least. Changing it to my suggestion would also make it more consistent with the loot table locations.

Actually, what if we made the loot table specifiable in the config? Then, if you're just referencing something, you can point to it directly without creating a new loot table file, and if you want to make something custom, you can reference any custom loot table. Then the config becomes the only "magic" resource location, instead of having two. Thoughts?

> (this will be a vanilla feature in later Minecraft versions, although the implementation here is slightly tweaked from what Trial Chambers use)

What's different? Will/should our implementation change to match more closely with vanilla in 1.21? Or will we be able to reuse vanilla code?

> whether the Loonium must be placed in the overall bounding box or an individual piece of the structure

Realizing on a second read that this means *any* individual piece of a structure, right? Vs the cuboid bounding box that encompasses the whole structure. But the way i first interpreted this made me think: Is it possible to detect *which* structure piece you're in? I guess it would depend on the structure type, i don't know too much about structures, but there's at least jigsaw placement with pieces. Are there other, non-jigsaw ways to make a structure? Anyway, different configs for individual structure pieces would be cool to have, though maybe that could be an additional change later.

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
```diff
@@ -0,0 +1,255 @@
...
+
+		/*
+		Note: Be careful about adding individual items instead of loot table references.
+		The Loonium will randomly either generate a virtual chest full of loot from any referenced table, or put just
+		one of the defined item entries into the "chest". Either way it only picks a single stack from that "chest".
+		Individual item entries need to be weighted accordingly to not have them picked too often.
+		Also, archaeology loot tables need to be handled carefully, due to their limited loot pool options.
+		*/
+
+		tables.put(getLootId(BuiltinStructures.ANCIENT_CITY),
```
artemisSystem: This could use a helper function, probably. Nicer to author and edit. Maybe an interface like:
```
put(tables, BuiltinStructures.ANCIENT_CITY,
  LootTableReference.lootTableReference(BuiltInLootTables.ANCIENT_CITY).setWeight(4),
  LootTableReference.lootTableReference(BuiltInLootTables.ANCIENT_CITY_ICE_BOX).setWeight(1)
)
```
I could do this myself, sounds like a fun refactor :3

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
```diff
@@ -0,0 +1,255 @@
...
+				buildVillageLootTable(BuiltInLootTables.VILLAGE_TAIGA_HOUSE, TAIGA_VILLAGE_LOOT)
+		);
+		tables.put(getLootId(BuiltinStructures.WOODLAND_MANSION),
+				LootTable.lootTable().withPool(LootPool.lootPool()
+						.add(LootTableReference.lootTableReference(BuiltInLootTables.WOODLAND_MANSION).setWeight(99))
+						.add(LootItem.lootTableItem(Items.TOTEM_OF_UNDYING).setWeight(1))
+				)
+		);
+
+		var output = new ArrayList<CompletableFuture<?>>(tables.size());
```
artemisSystem: I think it's cleanest if this is it's own function. Or maybe this stays in the `run` function, and a separate function is made that handles creating and populating `tables`. Again, i can handle this

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
```diff
@@ -0,0 +1,255 @@
...
+	@NotNull
+	private static LootTable.Builder buildDelegateLootTable(ResourceLocation reference) {
+		return LootTable.lootTable().withPool(LootPool.lootPool()
+				.add(LootTableReference.lootTableReference(reference))
+		);
+	}
+
+	@NotNull
+	private static LootTable.Builder buildOceanRuinLootTable(ResourceLocation archaeology) {
+		// Note: since the Loonium does not supply a location, treasure maps will roll as empty maps
```
artemisSystem, line 216: Could we make it supply a location? would be cool

TheRealWormbo: We could supply the Loonium's (effective) location, or the mob's death location. However, generating a treasure map can come with a significant performance impact, and the usefulness of the generated would probably depend on how much you can manipulate the location.

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
```diff
@@ -179,7 +179,7 @@ public CompletableFuture<?> run(@NotNull CachedOutput cache) {
 		for (var e : tables.entrySet()) {
 			Path path = pathProvider.json(e.getKey());
 			LootTable.Builder builder = e.getValue();
-			LootTable lootTable = builder.setParamSet(LootContextParamSets.EMPTY).build();
+			LootTable lootTable = builder.setParamSet(LootContextParamSets.ALL_PARAMS).build();
```
artemisSystem: Shouldn't this ideally be changed to the context parameters we actually supply/use?

TheRealWormbo: IIRC there's validation happening when the loot table is written, loaded, or evaluated. And while you don't need to set any parameters, having them not defined in the first place can lead to warnings or errors if the type of the referenced loot requires them.
The Loonium loot is generated in an empty context (it could probably be a different context that provides the flower location and mob reference somehow, but that could influence the type of items you get, e.g. due to cooking food drops if the mob was on fire, or whether it was a player kill with or without looting), but there are references to e.g. entity, chest, or archaeology type loot tables, which may have different parameter sets.

Long story short: This commit changed it from the actually supplied set (empty) to the widest available set (all params) because something didn't work as expected.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -262,7 +266,7 @@ public void readFromPacketNBT(CompoundTag cmp) {
 		super.readFromPacketNBT(cmp);
 		if (cmp.contains(TAG_LOOT_TABLE)) {
 			var lootTableString = cmp.getString(TAG_LOOT_TABLE);
```
artemisSystem: This is just a String right? should just say `String` instead of `var`

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -282,4 +286,30 @@ public static void dropLooniumItems(LivingEntity living, Consumer<ItemStack> con
...
+			String attuneType;
+			if (Arrays.equals(flower.lootTables, DEFAULT_LOOT_TABLES)) {
+				attuneType = "generic_drops";
+			} else if (flower.lootTables.length == 1) {
+				attuneType = "structure_drops";
+			} else {
+				attuneType = "multiple_structure_drops";
+			}
+			String attuned = I18n.get("botaniamisc.loonium." + attuneType).formatted(flower.lootTables.length);
+			int filterWidth = mc.font.width(attuned);
```
artemisSystem: filter? is this copied from hopperhock?

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -132,6 +132,9 @@
   "botaniamisc.rannuncarpus.state_sensitive": "Match Exact State",
   "botaniamisc.rannuncarpus.state_insensitive": "Match Block Only",
   "botaniamisc.lokiRingLimitReached": "Selection limit reached",
+  "botaniamisc.loonium.generic_drops": "Not attuned",
+  "botaniamisc.loonium.structure_drops": "Attuned to structure",
```
artemisSystem: Do we want to say which structure it's attuned to if it's only one? The name could be specified in the config maybe. Another reason to require config i guess. It would result in twice the files together with my other idea (config references a loot table, and we have loot table functions that limit to one drop), but maybe not a huge issue? It feels cleaner and the most customizable

TheRealWormbo: That part changed in a later commit to "Generic loot", "Structure loot", or "Custom loot" (if the Loonium BE NBT is edited to have a custom loot table assigned), but either way, structures do not have display names. The only thing we could show is the structure resource location.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: I think it is, oops.

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
```diff
@@ -0,0 +1,227 @@
...
+	public static final Codec<LooniumStructureConfiguration> CODEC = RecordCodecBuilder.create(
+			instance -> instance.group(
+					ExtraCodecs.POSITIVE_INT.fieldOf("manaCost").forGetter(o -> o.manaCost),
+					StructureSpawnOverride.BoundingBoxType.CODEC.fieldOf("boundingBoxType").forGetter(o -> o.boundingBoxType),
+					WeightedRandomList.codec(MobSpawnData.CODEC).fieldOf("spawnedMobs").forGetter(o -> o.spawnedMobs),
+					Codec.list(MobAttributeModifier.CODEC).fieldOf("attributeModifiers").forGetter(o -> o.attributeModifiers),
+					Codec.list(MobEffectToApply.CODEC).fieldOf("effectsToApply").forGetter(o -> o.effectsToApply)
+			).apply(instance, LooniumStructureConfiguration::new)
+	);
+	public static final Codec<LooniumStructureConfiguration> OPTIONAL_CODEC = RecordCodecBuilder.create(
```
artemisSystem: I think i'd call this one `WITH_PARENT_CODEC` or something, to more clearly get across the difference

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
```diff
@@ -0,0 +1,227 @@
...
+
+import org.jetbrains.annotations.Nullable;
+
+import java.util.List;
+import java.util.Optional;
+
+public class LooniumStructureConfiguration {
+	public static final Codec<LooniumStructureConfiguration> CODEC = RecordCodecBuilder.create(
+			instance -> instance.group(
+					ExtraCodecs.POSITIVE_INT.fieldOf("manaCost").forGetter(o -> o.manaCost),
```
artemisSystem: I think i'd prefer this to be NON_NEGATIVE_INT, for more customizability

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
TheRealWormbo: That was the first shot at implementing the codec logic. Turns out I didn't need the second one.

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
TheRealWormbo: Should work without further modifications, as far as I can tell.

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
```diff
@@ -0,0 +1,227 @@
...
+	}
+
+	public static class MobEffectToApply {
+		public static final Codec<MobEffectToApply> CODEC = RecordCodecBuilder.create(
+				instance -> instance.group(
+						BuiltInRegistries.MOB_EFFECT.byNameCodec().fieldOf("effect").forGetter(o -> o.effect),
+						ExtraCodecs.validate(Codec.INT,
+								duration -> duration > 0
+										? DataResult.success(duration)
+										: DataResult.error(() -> "Invalid effect duration"))
```
artemisSystem: Why not `ExtraCodecs.POSITIVE_INT`?

TheRealWormbo: I can't remember if that works with the -1 (infinite) default value. Maybe it does.

Xplat/src/main/java/vazkii/botania/common/config/ConfigDataManager.java
```diff
@@ -0,0 +1,118 @@
...
+			ResourceLocation id, ResourceLocation parent, Set<ResourceLocation> visitedEntries) {
+		if (!visitedEntries.add(id)) {
+			BotaniaAPI.LOGGER.warn("Cyclic dependency between Loonium structure configurations detected: {}", visitedEntries);
+			return false;
+		}
+		if (parent == null) {
+			return true;
+		}
+		var parentConfig = map.get(parent);
+		return parentConfig != null && findTopmostParent(map, parent, parentConfig.parent, visitedEntries);
```
artemisSystem: From what i can see, specifying a missing ID as the parent, will lead to the `Ignoring Loonium structure configuration(s) without top-most parent` error message. Or is that caught earlier? If not, maybe we should have an error specifically for that also

Xplat/src/main/java/vazkii/botania/api/BotaniaAPI.java
```diff
@@ -33,6 +33,7 @@
 import vazkii.botania.api.corporea.CorporeaNodeDetector;
 import vazkii.botania.api.internal.DummyManaNetwork;
 import vazkii.botania.api.internal.ManaNetwork;
+import vazkii.botania.common.config.ConfigDataManager;
```
artemisSystem: (forgive me if this changes in a later commit)

I don't think you can import a non-api class in the api. Someone depending on just the API wouldn't be able to call this method anyway.

Also, i'm not entirely clear on the purpose of this method. From what i can tell, the api impl starts with an instance of the config manager, and another instance is registered as a listener from the class itself. Then, whenever data is reloaded, setConfigData is called with that instance, which has the correct data. But, the instance stays there and doesn't need to be set again on next reload, so this all just seems a bit weird.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -78,116 +92,119 @@ public void tickFlower() {
...
-			pos = pos.above();
+		var randomPick = lootTables.get(world.random.nextInt(lootTables.size()));
+		var pickedConfig = structureConfigs.getOrDefault(randomPick.key(),
+				structureConfigs.get(LooniumStructureConfiguration.DEFAULT_CONFIG_ID));
+		var pickedLootTable = randomPick.value();
 
-			double x = pos.getX() + Math.random();
-			double y = pos.getY() + Math.random();
-			double z = pos.getZ() + Math.random();
+		if (getMana() < pickedConfig.manaCost) {
```
artemisSystem, line 201: Maybe filter to only pick configs we can afford? Or maybe we explicitly don't want this, because i guess it's kinda cheesable? Idk, open to discussion

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -78,116 +92,119 @@ public void tickFlower() {
...
+		double x = pos.getX() + Math.random();
+		double y = pos.getY() + Math.random();
+		double z = pos.getZ() + Math.random();
 
-			// TODO: mob attribute modifications should be defined along with mob configuration
-			entity.getAttribute(Attributes.MAX_HEALTH).addPermanentModifier(new AttributeModifier("Loonium Modififer Health", 2, AttributeModifier.Operation.MULTIPLY_BASE));
-			entity.setHealth(entity.getMaxHealth());
-			entity.getAttribute(Attributes.ATTACK_DAMAGE).addPermanentModifier(new AttributeModifier("Loonium Modififer Damage", 1.5, AttributeModifier.Operation.MULTIPLY_BASE));
+		Entity entity = pickedMobType.type.create(world);
+		if (!(entity instanceof Mob mob)) {
```
artemisSystem, line 239: booring :p 
let us spawn TNT :3
or maybe this breaks the attribute and mob effect stuff. and loot tables i guess..

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -78,116 +92,119 @@ public void tickFlower() {
...
-						entity = new Spider(EntityType.SPIDER, world);
-					}
-					break;
+		mob.finalizeSpawn(world, world.getCurrentDifficultyAt(pos), MobSpawnType.SPAWNER, null, null);
+		// in case the mob spawned with a vehicle or passenger(s), ensure those don't drop unexpected loot
+		mob.getRootVehicle().getPassengersAndSelf().forEach(e -> {
+			if (e != mob && e instanceof Mob otherMob) {
+				// prevent armor/weapon drops on player kill, also no nautilus shells from drowned:
+				Arrays.stream(EquipmentSlot.values()).forEach(slot -> otherMob.setDropChance(slot, 0));
+				LooniumComponent otherLooniumComponent = XplatAbstractions.INSTANCE.looniumComponent(otherMob);
```
artemisSystem, line 323: This doesn't apply to the root mob anymore, does it?

Xplat/src/main/java/vazkii/botania/api/BotaniaAPI.java
TheRealWormbo: Ah, right. Probably need to extract an API interface for the ConfigDataManager to use in those methods.

The way data pack reloading (for vanilla, and also for this) works is as follows: The game starts without any loaded data, which is an empty accessible instance of the data manager. When data packs are reloaded, a new instance (not yet accessible outside the loading logic) is created, and when loading is done, that instance replaces the previous accessible one.

I'm not sure if there's any better place to put this, but BotaniaAPI seemed like a good place for the method that returns the currently valid config data instance. I'm also not quite sure if the setter should be publicly accessible, but I couldn't figure out a clean way to set the instance to return.

Xplat/src/main/java/vazkii/botania/common/config/ConfigDataManager.java
TheRealWormbo: The config loading logic should be validating that any specified parent ID actually exists.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: I thought about this kind of scenario for a potential future configurable Orechid before. If mana comes trickling in, you essentially only ever get the cheapest conversion among available input blocks.

Not sure what the best approach is. I guess all the configs are loaded at this point anyway, so filtering could be an option, but it seems of very niche use, especially if all structures involved would require the same mana amount.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: And the Loonium data component that needs to be attached. I'm not sure what kind of non-mob entity types even make sense in a configuration, since all entity spawn attempts consume the same amount of mana. You can technically still spawn boss mobs, if you want to totally destroy players' hopes.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: In case of mob stacks (e.g. chicken jockey, or spider jockey), this prevents secondary mobs from dropping anything. In a chicken jockey, the "root vehicle" is a chicken with a jockey flag set, while in a spider jockey the "root vehicle" is the spawned spider, but it has a skeleton as passenger. So no random feathers, bones, or armor.

In later versions you will see an expansion of that piece of code to cover various edge cases, such as not spawning pillagers as captains, or actually assigning loot drops to known mob types.

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
artemisSystem: The current one also doesn't work with -1, does it?

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
artemisSystem: I think it's best to just pick one, and then do nothing if we can't afford it, even if there are other configs we can afford. So keep current behavior as it's implemented here

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
artemisSystem: yeah probably not really viable

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -197,48 +214,102 @@ private ItemStack pickRandomLoot(ServerLevel world, RandomSource rand) {
...
-			if (lootTable != LootTable.EMPTY) {
-				lootTableCandidates.add(candidateId);
-			}
+		if (structureMap.isEmpty()) {
+			detectedStructures = Object2BooleanMaps.emptyMap();
+		} else if (structureMap.size() == 1) {
+			var key = structureMap.firstKey();
+			detectedStructures = Object2BooleanMaps.singleton(key, structureMap.getBoolean(key));
+		} else {
+			detectedStructures = new Object2BooleanArrayMap<>(structureMap);
```
artemisSystem: Is this check necessary? Can't you just use the constructor like this in the empty and singleton cases too?

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -132,9 +132,9 @@
   "botaniamisc.rannuncarpus.state_sensitive": "Match Exact State",
   "botaniamisc.rannuncarpus.state_insensitive": "Match Block Only",
   "botaniamisc.lokiRingLimitReached": "Selection limit reached",
-  "botaniamisc.loonium.generic_drops": "Not attuned",
-  "botaniamisc.loonium.structure_drops": "Attuned to structure",
-  "botaniamisc.loonium.multiple_structure_drops": "Attuned to %d structures",
+  "botaniamisc.loonium.generic_loot": "Generic loot",
+  "botaniamisc.loonium.structure_loot": "Structure loot",
+  "botaniamisc.loonium.custom_loot": "Customized loot",
```
artemisSystem: I think i don't actually like this. It's not really relevant to the player (in my eyes at least), whether the loot is from a structure or set via NBT. I think i liked "Attuned" better. Maybe it could be "Attuned to \<structure\>", (grabbing the structure name from the config (only if we make the config mandatory)) and you could choose your own string for that if you're setting the mobs with nbt.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -392,9 +392,10 @@ public void writeToPacketNBT(CompoundTag cmp) {
 
 	public static void dropLooniumItems(LivingEntity living, Consumer<ItemStack> consumer) {
 		var comp = XplatAbstractions.INSTANCE.looniumComponent(living);
-		if (comp != null && !comp.getDrop().isEmpty()) {
-			consumer.accept(comp.getDrop());
-		}
+		if (comp != null)
+			if (!comp.getDrop().isEmpty() || comp.isDropNothing()) {
```
artemisSystem: This `if` only fails if the component is set to drop an empty stack, without dropNothing being set, right? Does that ever happen? To me, it looks like the case when the set drop is empty works like if dropNothing is set anyway, so maybe we don't need dropNothing at all and can just set the component with an empty stack?

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
```diff
@@ -198,10 +216,16 @@ public void tickFlower() {
 		world.addFreshEntity(mob);
 		mob.spawnAnim();
 
-		addMana(-COST);
+		addMana(-DEFAULT_COST);
```
artemisSystem: Oh, this needs to use the cost from the config, doesn't it?

Xplat/src/main/java/vazkii/botania/api/configdata/LooniumStructureConfiguration.java
TheRealWormbo: Fire resistance is an infinite effect for most mobs, so it should be working in the current state.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: I guess it could be considered premature optimization. The same would be true for the parts where that map is serialized to/from NBT.

Xplat/src/main/resources/assets/botania/lang/en_us.json
TheRealWormbo: I don't think structures have a translated name by default. Displaying the structure resource location feels wrong.

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
artemisSystem: Using the mob's death location sounds fun, then you could bring them with you places to roll treasure maps for that area

Xplat/src/main/resources/assets/botania/lang/en_us.json
artemisSystem: Yeah, which is why the name would need to be specified in the config

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: The component should always be attached to mobs on Fabric, so we need to somehow specify whether the mob was spawned by a Loonium or not. (I believe on NeoForge the data attachment can be created later on-demand, but the CCA logic for Fabric runs at the end of the entity's constructor.)

Maybe a better way to structure the data would be to have an "is Loonium-spawned" field that specifies that the `toDrop` item stack should be used, even if it's empty.

Xplat/src/main/java/vazkii/botania/common/block/flower/functional/LooniumBlockEntity.java
TheRealWormbo: That's changed in a later commit. One thing that doesn't change later, though, is the mana capacity. That is currently fixed at the default cost, so technically no structure could use a higher mana cost.

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
TheRealWormbo: That sounds interesting. I'm just a little worried about the performance impact of frequently generating maps, because locating structures can take several seconds. Also, the loot table rolling needs to be moved from mob spawn to mob death.

Xplat/src/main/resources/assets/botania/lang/en_us.json
TheRealWormbo: Config is server side, the display name would need to be in a language file (i.e. resource pack) on the client.

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
artemisSystem: Then let's shelve that idea for now and maybe come back to it later

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
artemisSystem: Could maybe be solved with a special loot function that makes the table ignore errors/warnings for missing parameters in a smooth way, that can be applied if the source loot table expects some parameters. But specifying all_params is probably fine for now

artemisSystem:
Not sure if it's mentioned in-thread, but i had an idea to make a loot table entry type that takes another as input and makes it output a single item, so we don't need to hardcode that functionality. But in 1.20.1, loot tables are still manually serialized and i don't think there's a registry for loot table entry types. There is a registry for them in 1.21 though, so we can make that change then.

artemisSystem:
I think we're good to go...

artemisSystem:
Again, thanks a bunch for this, i think this turned out really cool

Xplat/src/main/java/vazkii/botania/data/LooniumStructureLootProvider.java
artemisSystem: Ended up deciding this isn't super necessary

PR #4656, opened 2024-05-26
Move Diva Charm effect outside of damage logic and support angry neutral mobs

### LINKED ISSUE #4561: Charm of the Diva does not make mob attack something else
### Mod Loader

Forge

### Minecraft Version

1.20.1

### Botania version

443

### Modloader version

(dev snapshot)

### Modpack info

(none)

### The latest.log file

(n/a)

### Issue description

On Forge only the Charm of the Diva fails to "brainwash" mobs. The method itself succeeds, causing the mob to heal and particles to be displayed, but the mob will continue to attack the player.

This works fine on Fabric when hitting hostile mobs that implement the Enemy interface. However, when hitting mobs that *don't* implement that interface, yet can still attack (e.g. polar bear, llama), they never seem to get "brainwashed" into attacking nearby hostile mobs.

### Steps to reproduce

1. Equip the Charm of the Diva.
2. Find or spawn two hostile mobs, e.g. husks.
3. Hit one of them until you hear the charm's sound effect and see particles. On Fabric the mob starts attacking another hostile mob, on Forge it continues attacking the player.

### Other information

Bug report #4064 covers mobs that use the new "brain"-based AI system. That currently only affects piglins, hoglins, zoglins and the warden.

The main difference between Forge and Fabric seems to be that for Fabric a mix-in is used that calls the Charm of the Diva's code after the damage is done, while the Forge hook uses a built-in event that happens before the damage is applied. I assume the "brainwashing" is simply overridden by the mob because it considers the player-dealt damage after having a new target set, therefore reverting to attacking the player.

TheRealWormbo:
Fixes #4561 by temporarily storing mob IDs on the charm and applying the brainwashing during the next worn tick.
Fabric damage hook for the charm was also moved to LivingEntity::actuallyHurt for better parity with Forge.

### DIFF
--- a/Fabric/src/main/java/vazkii/botania/fabric/mixin/LivingEntityFabricMixin.java
+++ b/Fabric/src/main/java/vazkii/botania/fabric/mixin/LivingEntityFabricMixin.java
@@ -13,6 +13,7 @@ import net.minecraft.world.damagesource.DamageSource;
 import net.minecraft.world.entity.Entity;
 import net.minecraft.world.entity.EntityType;
 import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.player.Player;
 import net.minecraft.world.item.ItemStack;
 import net.minecraft.world.level.Level;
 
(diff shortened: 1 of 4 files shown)


TheRealWormbo:
Neutral mob support works as follows:
- A mob can be charmed if it is not a tamable mob owned by the player, and it is an enemy mob, or a neutral mob that is angry at the player or targeting one of the player's tamed animals.
- A mob that is not tamed by the player can become the target of a charmed mob if it is an enemy mob, or it is a neutral mob that is angry at the player or targeting one of the player's tamed animals.
- Mobs in a "stack" (e.g. skeleton and spider in a spider jockey) won't target each other.

PR #4659, opened 2024-05-27
Made plant check of Agricarnation more flexible

### LINKED ISSUE #4644: Agricarnation has conflicting requirement and effect
### Mod Loader

### Minecraft Version

1.20.1

### Botania version

443

### Modloader version

(any)

### Modpack info

_No response_

### The latest.log file

(n/a)

### Issue description

Currently the Agricarnation flower explicitly looks for plants that are bonemeal targets, but then only random-ticks them, without even checking if the target even accepts random ticks. This leads to a couple inconsistencies:

- azalea bushes do not grow, even though they should probably be considered saplings
- cactus and sugarcane is ignored, which is intentional (see #856) but apparently not obvious
- on the other hand, very similar blocks like bamboo and twisting vines *are* boosted by the Agricarnation's effect
- some blocks like rooted dirt, glow lichen, or mangrove leaves cause the flower to consume mana without having any effect, because they can be bonemealed but not random-ticked
- hanging plants (glowberry vines, weeping vines) only have their top block randon-ticked, which does nothing for the rest of the plant, which grows downwards; in fact, after having grown to at least two blocks height, such vines cause the flower to consume mana without doing actually anything (glowberry vines don't even grow glowberries this way)

### Steps to reproduce

_No response_

### Other information

Agricarnation not growing sugarcane was declared intentional a long time ago, as even back then it checked that target blocks could accept bonemeal. However even back then it instead triggered block updates, which at the time where equivalent to random ticks for blocks that cared about those.

TheRealWormbo:
Resolves #4644 (and also closes #4446, which is superseded by this PR) by implementing the rule set laid out in that issue:

- when the location of a `GrowingPlantBodyBlock` is encountered, the corresponding head block location is used instead
- new block tags:
  - `#botania:agricarnation/growth_candidate` specifies blocks that should be considered candidates, in addition to any blocks that extend `BushBlock` or accept bonemeal
  - `#botania:agricarnation/apply_bonemeal` defines blocks that should be boosted by applying bonemeal instead of random ticks (only relevant if the block actually accepts bonemeal)
  - `#botania:agricarnation/growth_excluded` excludes blocks from any growth boost considerations
- azalea bushes are bonemealing candidates via the corresponding tag by default, to make them work similar to other saplings
- the bonemealing tag is included in the growth candidates tag by default
- vanilla mushrooms are growth-excluded via the corrsponding tag by default
- blocks extending `SpreadingSnowyDirtBlock` (i.e. grass or mycelium) or `NyliumBlock` are excluded via code, unless explicitly tagged as growth candidate
- the flower will boost growth for candidate blocks that can be random-ticked, or that are tagged for bonemeal application and also accept bonemeal
- bonemeal application needs to pass a built-in chance check and the block's own success check before applying the bonemeal effect; mana will still be consumed if either of those checks fail

### DIFF
--- a/Xplat/src/generated/resources/.cache/bfa01a6ca2555c100103725bf5c9e6da285f29c3
+++ b/Xplat/src/generated/resources/.cache/bfa01a6ca2555c100103725bf5c9e6da285f29c3
@@ -1,4 +1,7 @@
 // 1.20.1	Botania/Tags for minecraft:block
+8865bb2eca4192a496bda05497a584aee6a3a409 data/botania/tags/blocks/agricarnation/apply_bonemeal.json
+1f1a3978ea72d2be2545d91126d194b1b6639c49 data/botania/tags/blocks/agricarnation/growth_candidate.json
+caa65214e42ad067ce1935d9b7d48e0a9feb9e73 data/botania/tags/blocks/agricarnation/growth_excluded.json
 7c8a171ae6d6133a08aa4d939254a4290e6b44b6 data/botania/tags/blocks/corporea_spark_override.json
 e8a61a147b6a258b9c37788ac1bd4edf592c9dc9 data/botania/tags/blocks/double_mystical_flowers.json
 070322f325bf48a093e7a873066fd5992202fd74 data/botania/tags/blocks/dreamwood_logs.json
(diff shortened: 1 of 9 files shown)


artemisSystem:
Rebased on top of the build fix

PR #4660, opened 2024-05-28
Update Dandelifeon to respect enchanted soil

### LINKED ISSUE #4405: enchanted soil does not accelerate Dandelifeon
### Mod Loader

Forge

### Minecraft Version

1.19.2

### Botania version

1.19.2-440

### Modloader version

Forge:43.2.3

### Modpack info

_No response_

### The latest.log file

N/A

### Issue description

Enchanted soil does not accelerate dandelifeon in 1.19.2

### Steps to reproduce

_No response_

### Other information

_No response_

NEstoll:
Fixes #4405

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
@@ -67,8 +67,8 @@ public class CellularBlockEntity extends BotaniaBlockEntity {
 		generation = nextGeneration;
 	}
 
-	public boolean hasActiveParent(Level level) {
-		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && level.hasNeighborSignal(flowerCoords) && ((level.getGameTime() % DandelifeonBlockEntity.SPEED == 0) || parent.overgrowth);
+	public boolean hasActiveParent(DandelifeonBlockEntity dandie) {
+		return flowerCoords != null && dandie.getLevel().getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && dandie.getLevel().hasNeighborSignal(flowerCoords) && parent.overgrowthBoost == dandie.overgrowthBoost;
 	}
 
 	public int getGeneration() {
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
@@ -68,7 +68,7 @@ public class CellularBlockEntity extends BotaniaBlockEntity {
 	}
 
 	public boolean hasActiveParent(Level level) {
-		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && level.hasNeighborSignal(flowerCoords) && (!(level.getGameTime() % (DandelifeonBlockEntity.SPEED / 2) == 0) || parent.overgrowth);
+		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && level.hasNeighborSignal(flowerCoords) && ((level.getGameTime() % DandelifeonBlockEntity.SPEED == 0) || parent.overgrowth);
 	}
 
 	public int getGeneration() {
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
@@ -67,8 +67,8 @@ public class CellularBlockEntity extends BotaniaBlockEntity {
 		generation = nextGeneration;
 	}
 
-	public boolean hasPoweredParent(Level level) {
-		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity && level.hasNeighborSignal(flowerCoords);
+	public boolean hasActiveParent(Level level) {
+		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && level.hasNeighborSignal(flowerCoords) && (!(level.getGameTime() % (DandelifeonBlockEntity.SPEED / 2) == 0) || parent.overgrowth);
 	}
 
 	public int getGeneration() {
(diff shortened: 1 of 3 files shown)


TheRealWormbo:
I feel like the enchanted soil status should probably have an effect on the boundary cell check somehow, but I'm not sure if the adjacent flower also needs to be on enchanted soil, or if otherwise boundary cells from non-boosted flowers should just be ignored.

Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
```diff
@@ -67,8 +67,8 @@ public void update(Level level) {
 		generation = nextGeneration;
 	}
 
-	public boolean hasPoweredParent(Level level) {
-		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity && level.hasNeighborSignal(flowerCoords);
+	public boolean hasActiveParent(Level level) {
+		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && level.hasNeighborSignal(flowerCoords) && ((level.getGameTime() % DandelifeonBlockEntity.SPEED == 0) || parent.overgrowth);
```
TheRealWormbo: It's not great that the timing is being checked here. The query for whether a cellular block's parent flower is on enchanted soil should probably go in a separate method.

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/DandelifeonBlockEntity.java
```diff
@@ -58,9 +58,9 @@ public void tickFlower() {
 		super.tickFlower();
 
 		if (!getLevel().isClientSide) {
-			if (getLevel().getGameTime() % SPEED == 0 && getLevel().hasNeighborSignal(getBlockPos())) {
+			if ((getLevel().getGameTime() % SPEED == 0 || (getLevel().getGameTime() % (SPEED / 2) == 0 && overgrowth)) && getLevel().hasNeighborSignal(getBlockPos())) {
```
TheRealWormbo: There are now at least 5 different instances (counting the one in the cellular block) where game time is tested. As far as I can tell the four potentially relevant numbers modulo `SPEED` are `0`, `SPEED-1`, `SPEED/2`, and `SPEED/2-1`, with the latter two only mattering during overgrowth ticks. IMHO these numbers should be given reasonable names and used in a method that determines whether it's time to do something, based on whether it's a simulation tick, a claiming tick, or none of those.

Speaking of overgrowth ticks, I think this line might not not prevent the tick from happening twice for flowers on enchanted soil. It doesn't seem to do anything when it runs a second time, and I'm not sure why, but that should probably not happen. I would recommend running the regular tick when `overgrowthBoost` is false and the overgrowth tick when `overgrowthBoost` is true in addition to `overgrowth` being true. This way you can tell wither it's an overgrowth tick by simply checking the value of `overgrowthBoost`, e.g. during the boundary cell check, when you need to know if the other flower has to be on enchanted soil for its cells to count.

TheRealWormbo (changes requested):
See code comments. Functionality-wise it looks good to me, though.

Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
NEstoll: I'm not sure where else it would go, as putting in a different method would require duplicating the calls to verify that flowerCoords points to a real dandelifeon. I've removed the timing calls however.

Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
```diff
@@ -67,8 +67,8 @@ public void update(Level level) {
 		generation = nextGeneration;
 	}
 
-	public boolean hasActiveParent(Level level) {
-		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && level.hasNeighborSignal(flowerCoords) && ((level.getGameTime() % DandelifeonBlockEntity.SPEED == 0) || parent.overgrowth);
+	public boolean hasActiveParent(DandelifeonBlockEntity dandie) {
+		return flowerCoords != null && dandie.getLevel().getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && dandie.getLevel().hasNeighborSignal(flowerCoords) && parent.overgrowthBoost == dandie.overgrowthBoost;
```
TheRealWormbo: Unfortunately, the `overgrowthBoost` (and also `overgrowth`) variable cannot be used outside the flower's own tick logic, as it is reset at the end of `tick()`. You will have to call `isOnSpecialSoil()` for the other flower.

TheRealWormbo (changes requested):
See comment in code. Also, please rebase onto current `1.20.x` branch head to fix the dependency issues with VanillaGradle.

TheRealWormbo (approved):
Looks good to me.

Xplat/src/main/java/vazkii/botania/common/block/block_entity/CellularBlockEntity.java
```diff
@@ -67,8 +67,8 @@ public void update(Level level) {
 		generation = nextGeneration;
 	}
 
-	public boolean hasPoweredParent(Level level) {
-		return flowerCoords != null && level.getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity && level.hasNeighborSignal(flowerCoords);
+	public boolean hasActiveParent(DandelifeonBlockEntity dandie) {
+		return flowerCoords != null && dandie.getLevel().getBlockEntity(flowerCoords) instanceof DandelifeonBlockEntity parent && dandie.getLevel().hasNeighborSignal(flowerCoords) && (!dandie.overgrowthBoost || parent.isOnSpecialSoil());
```
artemisSystem, line 71: This or-check at the end is weird to me. It's true if the flower it's being transferred to is not boosted, or if the current flower is boosted. (Did i get that right?) So this returns true when both are boosted or neither are boosted, and false if the old flower is not boosted and the new one is. But if the old flower is boosted and the new one is not, this will still return true, right? Which we don't want, right?

It would also be nice if these variable names better reflected which flower it's coming from and which flower it's going to.

NEstoll: This method is always called by the new flower, so we only need to make sure that either this is a normal tick (we aren't in an overgrowth tick), or that the other flower is also enchanted, and therefor will tick as well. This means that adjacent flowers will only share cells on ticks where they both are ticking. If the old flower is boosted and the new one isn't, then the new one won't be ticked, so this will never be called, and thus the return value doesn't matter.

PR #4661, opened 2024-05-30
Smelty rod improvements

NEstoll:
Changes the Rod of the Molten Core to use punch rather than holding right click to smelt blocks. This both brings it in line with how the rod of the shifting crust works and improves its usability.


NEstoll:
Furthermore, I added the ability for the rod to smelt items in inventory using mana. Mana cost is currently the same per item as it normally is per block, but that is simple to change

TheRealWormbo:
I think the application of the block smelting effect should go back to right-click, but it can stay instant. That would also be more consistent with your proposed in-inventory smelting, which uses right-click. However, would it maybe make more sense (and be safer) that you right-click the rod item onto the stack and not vice versa?

There will also need to be some kind of replacement for the rod proficiency bonus, which before was a smelting speed increase. But we should discuss that part on Discord.

PR #4665, opened 2024-06-05
Add ramping heat to the Thermalily

NEstoll:
Currently, the Thermalily breaks a fundamental design principle of Botania, as automating it even partially is quite complex while only giving fairly paltry amounts of mana. This PR aims to fix that by decreasing the average cooldown of the Thermalily if lava is provided promptly after the cooldown ends. This stacks up to 10 times, eventually reducing the average cooldown down to 27 seconds. This means if the player is able to supply a large amount of lava consistently, the Thermalily will produce almost thrice its current mana/second.

Potential additional changes:
Allow the Thermalily to consume lava from caldrons, allowing for player-less automation
Increase the mana per tick rather than or in addition to the cooldown reduction to improve the mana per lava bucket
Implement a stronger penalty for "mistiming" a heated Thermalily (explodes, catches on fire, decays, etc.)


TheRealWormbo:
So, to summarize the intended changes in this PR:

- Thermalilies and Hydroangeas are both allowed to drink directly from cauldrons
- a streak-like mechanic ("heat") was implemented for the Thermalily, which provides bonus speed and efficiency for setting up automation:
  - heat increases by 1 (up to 25) if the next lava is consumed within 10 ticks of the cooldown ending, but immediately resets to 0 if lava is consumed outside that specific time window
  - cooldown duration is no longer chosen from a static weighted list, but calculated from a normal distribution, which gets more narrow and shifts towards lower average values at higher heat levels (increases average output rate)
  - active mana output per tick is increased at higher heat levels (increases output per lava bucket and average rate)

Static automation that always assumes the worst-case of 5 minutes cooldown is unaffected by this change, producing just under 1/3rd of a pool per hour. Manually feeding the flower may be slightly less efficient, since the average cooldown time at heat 0 is somewhat higher than before the change. Dynamic automation that tracks the actual cooldown time can be much more efficient at a bit over 2.5 pools per hour. That's in the general area of a Gourmaryllis alternating between cooked cod and salmon every 3 seconds (where the difficulty is obtaining that much fish for extended run times), or a single-flower 100-cycle 6-block Dandelifeon setup.

Xplat/src/main/java/vazkii/botania/common/block/flower/generating/ThermalilyBlockEntity.java
```diff
@@ -70,7 +58,7 @@ public void tickFlower() {
 			ticksSinceFueled++;
 		} else if (burnTime == startBurnTime) {
 			if (ticksSinceFueled <= FAST_PROVIDE_TICKS) {
-				heat = heat < 10 ? heat + 1 : heat;
+				heat = heat < MAX_HEAT ? heat + 1 : heat;
 			} else {
 				heat = 0;
```
TheRealWormbo, line 63: As far as I can tell, this is the only place where `heat` is reset. If lava is provided before the end of the cooldown, the cooldown restarts (as intended), but heat does not reset. I'm not sure if that's a bad thing, as lava is still wasted, so fine by me.

NEstoll: Ideally I would have made it so that providing lava before the end of the cooldown would cause the thermalily to "burn" into a dead bush, but that would require either changing the logic in FluidGeneratingFlower (doable, requires some refactoring and moving of code), or use the call to getCooldownTime(false), (janky and prone to errors if other mods/addons/code calls that).

artemisSystem: i think just resetting the heat is an ok penalty here

artemisSystem:
One can wonder whether this goes against the spirit of the thermalily, if it's considered a manual "quick, get me some mana" flower, kinda like the rosa arcana. But, since we already added the variable cooldown as an automation puzzle, i think this is a reasonable extension of that idea. I'll review the code itself more in-depth later.

artemisSystem:
We also need documentation. And an effect showing that the flower is "heated up" would be cool, if that's not already a thing

artemisSystem:
Slight (or, massive, actually) issue with being able to take lava from cauldrons: It effectively turns the thermalily into a passive flower

TheRealWormbo:
True, but giving lava at the wrong time (which the passive use variant would inevitably do) makes it inefficient. A cauldron refills from pointed dripstone roughly every 19.4 minutes. And if you put multiples near the flower, you risk it taking the lava during the cooldown, which restarts the cooldown.

IMHO it's the best way to automate the Thermalily, unless we can provide another way for Botania to make lava renewable without player interaction.

artemisSystem:
That amounts to ~83500 mana per hour, vs a hydroangea's 72000. You could just place a row of thermalilies and have them work better than hydroangeas, except they would never decay. Setup and flower cost is a bit more expensive than hydroangeas (earth + fire rune, and you need the lava and dripstone), but this PR would effectively turn the thermalily into a slightly more expensive, slightly better, and never-decaying version of the hydroangeas.

I'm starting to think the original random cooldown change was a bad idea. The Thermalily Was Not Meant To Be Automated, and you can't really make it automatable without completely breaking the mod at this point.

TheRealWormbo:
> I'm starting to think the original random cooldown change was a bad idea.

Not going to argue with that. That modification broke the effort-to-output "rule" of generating flowers by being way too slow for the amount of effort the automation required.

NEstoll:
> You could just place a row of thermalilies and have them work better than hydroangeas, except they would never decay

This was the original reason behind wanting to make the thermalily "burn" into a dead bush if you supply lava too early (as opposed to just losing heat if too late), as a fully passive setup will eventually cause the thermalily to burn.

PR #4669, opened 2024-06-06
Allow avatar to use rod of the depths

NEstoll:
Allow for living wood avatars to use the rod of the depths. Currently uses the MODEL_AVATAR_DIRT, because I lack the skills to make a cobble version


artemisSystem:
This was discussed a bit in the discord a little while ago, and i think it makes sense to not allow the avatar to use the cobble rod, since you can make cobble with a cobblestone generator

PR #4673, opened 2024-06-10
Fruit now automatically fills hunger

NEstoll:
The Fruit of Grisaia now automatically fills the player's hunger bar if it's held in the hot bar.


TheRealWormbo:
Not certain if this will actually be merged.

artemisSystem:
I don't think this is something that we want, sorry

PR #4676, opened 2024-06-12
Adds recipies to replace lost relics

NEstoll:
Mock relics can be crafted with gaia spirit ingots, and infused at the terrasteel plate to regain lost relics


artemisSystem:
I'm not a huge fan of this idea honestly. I also don't see the point of the 6 new classes. The items can just be `Item`, no?

NEstoll:
Being able to permanently lose the endgame boss relics with no way of getting them back seems like a design flaw to me, especially considering how easy it is to lose them. As for the 6 new classes, I was simply copying the relics because I don't actually know how to add items a different way.

artemisSystem:
I might agree that there should be a way to get back lost relics, but i don't think this is the way i think we should do it, sorry.

PR #4683, opened 2024-06-20
Damage type datagen fix

BrokenK3yboard:
Fix datagen for damage types, manually writing entries should no longer be necessary.


TheRealWormbo (commented):
I think the `FabricDynamicProvider` class could use a better name that describes what it's actually used for. Otherwise looking good to me.

BrokenK3yboard:
It's extended from `FabricDynamicRegistryProvider` and is meant to register multiple different types. Enchantments and Enchantment providers need to be registered there as well for them to be generated (At least in 1.21) among other things, so I don't have any clear-cut name in mind.

TheRealWormbo:
It's just that the class name itself says even less about what it's supposed to do than its parent class.
So as far as I understand, it fills dynamic registries with data relevant to Botania's datagen run. How about `BotaniaDynamicRegistryProvider`, then?

BrokenK3yboard:
Sounds fine to me.

PR #4684, opened 2024-06-20
xplat: Switch from VanillaGradle to minivan

quat1024:
[`minivan`](https://github.com/CrackedPolishedBlackstoneBricksMC/minivan) is my Gradle plugin that acts like VanillaGradle. It is written with pretty boring straightforward code, makes no attempt to download a copy of the Minecraft assets or anything else necessary to support a `runClient`, and doesn't use any IntelliJ "task activations", so it's a good bit lighter weight.

Interesting: When switching, I had to manually add a dependency on jetbrains annotations. Maybe vanillagradle did that?

Also adds the foojay-resolver-convention plugin, b/c apparently I don't have Java 17 installed in a place where Gradle can find it. This is a well-known Gradle 8 upgrade woe wrt. the toolchains feature.

### DIFF
--- a/settings.gradle
+++ b/settings.gradle
@@ -11,10 +11,23 @@ pluginManagement {
         maven {
             name = 'Sponge Snapshots'
             url = 'https://repo.spongepowered.org/repository/maven-public/'
+        }
+        maven {
+            name = 'Sleeping Town'
+            url = 'https://repo.sleeping.town/'
+            content {
+                includeGroup 'agency.highlysuspect'
+                includeGroup 'agency.highlysuspect.minivan'
+            }
+        }
         
         gradlePluginPortal()
-        }
     }
 }
+
+plugins {
+    id("org.gradle.toolchains.foojay-resolver-convention") version "0.8.0"
+}
+
 rootProject.name = 'Botania'
 include("Xplat", "Fabric", "Forge")
--- a/Xplat/build.gradle
+++ b/Xplat/build.gradle
@@ -1,6 +1,6 @@
 plugins {
     id('java')
-    id('org.spongepowered.gradle.vanilla') version '0.2.1-SNAPSHOT'
+    id('agency.highlysuspect.minivan') version '0.5-SNAPSHOT'
 }
 
 archivesBaseName = "${mod_name}-xplat"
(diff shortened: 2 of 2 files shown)


TheRealWormbo:
Is that just a thing with the IntelliJ decompiler, or does the generated code really have `@SideOnly(Side.CLIENT)` on too many classes? (I also noticed that before this update with access widening.)

quat1024:
Hm, interesting, might be a failure of the jar merging.

Which classes?

TheRealWormbo:
So far I've noticed it on any entity class I checked, including `ServerPlayer`. It might just be all classes, including nested ones. Unfortunately IntelliJ likes to navigate to Fabric versions of classes (presumably because those have configured sources), which is making this a bit harder to check.

quat1024:
Well that's embarrassing, I forgot to unbundle the server jar. Glad i didn't publish 0.5 yet >.>

Should be fixed in the latest minivan snapshot (commit hash [c5ae507](https://github.com/CrackedPolishedBlackstoneBricksMC/minivan/commit/c5ae5070377472b312cc4d287e0179b53cef7107)).

TheRealWormbo:
Yup, looks better.

settings.gradle
```diff
@@ -11,10 +11,23 @@ pluginManagement {
         maven {
             name = 'Sponge Snapshots'
             url = 'https://repo.spongepowered.org/repository/maven-public/'
+        }
+        maven {
+            name = 'Sleeping Town'
+            url = 'https://repo.sleeping.town/'
+            content {
+                includeGroup 'agency.highlysuspect'
+                includeGroup 'agency.highlysuspect.minivan'
```
artemisSystem, line 20: are both these needed?

quat1024: First one is for the plugin itself (group `agency.highlysuspect`, artifact `minivan`), second one is for [the plugin marker artifact](https://repo.sleeping.town/agency/highlysuspect/minivan/agency.highlysuspect.minivan.gradle.plugin/) (group `agency.highlysuspect.minivan`, artifact `agency.highlysuspect.minivan.gradle.plugin`). [Weird gradle thing](https://docs.gradle.org/current/userguide/plugins.html#sec:plugin_markers).

artemisSystem:
Tested a bit and it seems to work just fine. Thanks a lot!

PR #4688, opened 2024-06-23
Pollidisiac support for feeding baby animals

TheRealWormbo:
The flower can now feed baby animals in addition to (or instead of) adult animals.

- Wand of the Forest in function mode can toggle the flower between feeding adult/baby/all animals (default is adult-only, to not break existing contraptions)
- current mode is shown in wand HUD
- if allowed to feed babies, they are only fed every couple seconds, and only while they have more than 30 seconds left to grow up


artemisSystem:
Looks good! I added some documentation and made a small code change

PR #4705, opened 2024-07-05
British English lang file

anoomolu:
> basically got sick of typing gray to find Botania items when using Minecraft's en_gb lang

Just spelling changes really - gray to grey, -or and -ize suffixes to -our and -ise, chiseled to chiselled and so on.


TheRealWormbo:
That leaves AU, CA, and NZ variants as remaining "serious" English languages supported by vanilla Minecraft. A quick search suggests that Canadian mixes American and British spelling enough (e.g. -our, but -ize) to warrant its own stripped-down version, while this file might also pass for the other two.

PR #4715, opened 2024-07-12
Some tiny lexicon adjustments

anoomolu:
Fixes the advancement header fr this time (was missing from the big commit), manaweave armour to robes in the Imagination Forest advancement and adds a page in the Alchemy section for Dripleaf shrinking.


Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -5,7 +5,7 @@
   "_comment3": "Untranslated entries will automatically be inherited from this file.",
   "_comment4": "Once again, don't copy from English, and THANK YOU for your hard work!",
 
-  "itemGroup.botania.botania": "Botania",
+  "itemGroup.botania": "Botania",
```
TheRealWormbo: I'm pretty sure changing this property will break something. Can't remember what.

anoomolu: I’ve had this change in a resource pack for a while now and haven’t noticed anything wrong, hmmm

TheRealWormbo: I believe it's at least the creative tab name, and there were other changes to adjust similar usages of that property.

TheRealWormbo: I assume your resource pack version of en_us.json merely shadowed matching entries from the mod's own file, which is why it still worked. Anyway, I renamed the translation key uses in code now and reverted the key to what it used to be (and still is in some other language files) like you did in this PR.

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -489,7 +489,7 @@
   "advancement.botania:craftingHaloCraft": "Segments Blue",
   "advancement.botania:craftingHaloCraft.desc": "Craft an Assembly Halo to facilitate crafting",
   "advancement.botania:manaweaveArmorCraft": "Imagination Forest",
-  "advancement.botania:manaweaveArmorCraft.desc": "Create Manaweave Armor to power up your Rods",
+  "advancement.botania:manaweaveArmorCraft.desc": "Create Manaweave Robes to power up your Rods",
```
TheRealWormbo, line 492: This fix also needs to happen in `en_gb`.

anoomolu: Good catch, thank you ^^

PR #4716, opened 2024-07-12
Document Teru Teru Bozu + Blue Orchid interaction

anoomolu:
Feel free to close this if not desired - I asked a while back on Discord and there was conflicting info on whether it's meant to be an easter egg or documented.


TheRealWormbo:
Apparently this is supposed to be an easter egg, but I'm all for documenting it. Supposedly the flavor text on the recipe is a hint, but I think it's too subtle for a feature that synergizes nicely with Riptide.

artemisSystem:
Thanks!

PR #4748, opened 2024-09-05
Update ru_ru.json

krumplerban:
Huge update for the russian lang file


krumplerban (commented):
Translated and updated many lexicon entries

Issue #4589, opened 2024-03-25
IncompatibleClassChangeError with Dye Depot


TheRealWormbo:
### Mod Loader

Both Fabric and Forge (I confirm that I have tested both loaders and will specify both loader versions below)

### Minecraft Version

1.20.1

### Botania version

1.20.1-443

### Modloader version

Forge 47.1.3, Forge 47.2.21, Fabric Loader 0.15.7, Fabric API 0.92.0+1.20.1

### Modpack info

(n/a)

### The latest.log file

https://gist.github.com/TheRealWormbo/69313e37da2252337681ff0f96382003

### Issue description

The presence of Dye Depot causes Botania's floating flower registration logic to crash with an `IncompatibleClassChangeError`.

### Steps to reproduce

Start Minecraft with Botania (+ dependencies) and Dye Depot installed.

### Other information

User [MagicFlowersss](https://forum.violetmoon.org/u/MagicFlowersss) initially [reported the incompatibility](https://forum.violetmoon.org/d/208-my-modpack-with-botania-installed-crashes-during-startup-of-minecraft) for Forge and confirmed it to be independent of the loader version, later finding out it was caused by Dye Depot.

ereese99:
I too am getting this @Vazkii

TheRealWormbo:
> I too am getting this @Vazkii

For the record: Vazkii handed off Botania development years ago already.

I'm also inclined to track this issue for reference only, as it's really Dye Depot's design decision to implement a binary-incompatible change.

PR #4764, opened 2024-11-12
Specify supported dyes to allow co-existing with dye addon mods

PssbleTrngle:
This is a proposed fix for #4589

I extracted all required places which use `DyeColor.values()` and replaced it with a reference to a centralized location which manually specifies the vanilla dyes.
This way, dye colors injected by mods like dye  depot are simply ignored.

### DIFF
--- a/Fabric/src/main/java/vazkii/botania/fabric/FabricCommonInitializer.java
+++ b/Fabric/src/main/java/vazkii/botania/fabric/FabricCommonInitializer.java
@@ -86,6 +86,7 @@ import vazkii.botania.common.crafting.BotaniaRecipeTypes;
 import vazkii.botania.common.entity.BotaniaEntities;
 import vazkii.botania.common.entity.GaiaGuardianEntity;
 import vazkii.botania.common.handler.*;
+import vazkii.botania.common.helper.ColorHelper;
 import vazkii.botania.common.helper.PlayerHelper;
 import vazkii.botania.common.impl.BotaniaAPIImpl;
 import vazkii.botania.common.impl.DefaultHornHarvestable;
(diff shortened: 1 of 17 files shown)


CraftyZombie:
Is there no way to somehow implement this fix into Dye Depot itself?
The number of mods it crashes with for the same reason is endless.
The current implementation is obviously centered around Botania, but is there no way to make the game less strict about hardcoded dyes?

PssbleTrngle:
The only way to fix this in dye depot I could think of, which would have been the alternative, is mixing into Botania and doing the same thing there.
Doing what this PR does seems to my like the cleanest and most simple solution to the problem

PssbleTrngle:
The only way to make the game less strict about hardcoded dyes is by not assuming that they are hardcoded, which this PR does. Dye Depot crashed with some other mods, but almost none directly on start time and most so far could be solved by mixing into one or maybe two specific method, in Botania there are a lot more places like you can see

TheRealWormbo:
I don't see any mention of the Spectrolus in the changes. Did you test that as well?

PssbleTrngle:
No, I have not changed how the Spectrolus works, it still only requires vanilla wool. But I've checked how it's implemented and since it simply counts the index up to 15 and then resets to 0, it will not cause any problems with dye addons.

TheRealWormbo:
@PssbleTrngle: Could you please run `spotlessApply` on your changes to fix the build error? It's probably just an unused import.

PssbleTrngle:
Sorry for the late response, but it looks like you already took care of it, right?

TheRealWormbo:
Yes. I'm going to give it a proper test later today and then merge it.
Your changes also seem like a solid base to finally bring Botania's colored items into the new vanilla order in the Creative inventory, so there's even benefit beyond just compatibility.

PssbleTrngle:
Thank you very much and nice to know, did not even notice that

PR #4766, opened 2024-11-20
Fixed syntax of vertex, color, uv, and others for 1.21.1 port

kitminty:
Fixed the syntax of all instances of vertex to addVertex, color to setColor, uv to setUv, and most similar things with Tesselator


TheRealWormbo:
If you want to participate in the 1.21.1 porting, I highly recommend joining the #botania_engineering channel on the Violet Moon discord. That way you can coordinate your changes with the other people working on the port, and not have them sit around for a while before being noticed at all.

kitminty:
Thank you! I will definitely do that as soon as I can.

PR #4767, opened 2024-11-24
Update zh_cn.json.

Dawnwalker666:
The scope of the translation includes all commits about en_us.json from Sep 1, 2024 to Nov 18, 2024.


Dawnwalker666:
I don't know why there's a problem with the display checking, it seems to be stuck on ‘Build with Gradle’. But I just submitted a language file and rechecked it, so it shouldn't be my problem.

TheRealWormbo:
It might be worth starting a new branch or otherwise resetting your fork to the state of the main branch after this merge. You appear to be merging main branch changes into your branch via a merge commit, while Botania's main repository uses rebasing.

(Though I have no idea why your changes failed to build here.)

PR #4768, opened 2024-11-27
Custom TNT ethical fix

Suisuroru:
This bug will not happen on vanilla server, it will happened with some mod like AE2, their custom tnt can provide the ethical logic misjudgment. I have modified the logic for determining whether TNT is ethical to recognize other entities with TNT fields. I have checked the situation where the vanilla item does not have TNT, and those that have been judged as ethical have already been judged as TNT, so there should be no misjudgment

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/helper/EthicalTntHelper.java
+++ b/Xplat/src/main/java/vazkii/botania/common/helper/EthicalTntHelper.java
@@ -118,7 +118,7 @@ public class EthicalTntHelper {
 				final var blockEntity = entity.level().getBlockEntity(blockPos);
 				if (blockEntity instanceof PistonMovingBlockEntity movingBlockEntity
 						&& movingBlockEntity.getMovementDirection() == dir
-						&& movingBlockEntity.getMovedState().getBlock() instanceof TntBlock) {
+						&& ((movingBlockEntity.getMovedState().getBlock() instanceof TntBlock) || movingBlockEntity.getMovedState().getBlock().getName().toString().contains("tnt"))) {
 					// found a moving block that marks the destination of a TNT block moving away from the TNT entity
 					XplatAbstractions.INSTANCE.ethicalComponent(entity).markUnethical();
 					break;


TheRealWormbo:
I think I'd prefer a solution that doesn't use string matching, because this needs to test every single block that is being moved and thus is somewhat performance-critical. Maybe something like testing the moved state for a block tag?

Which block exactly gets around the check?

Suisuroru:
In AE2, the custom tnt is appliedenergistics2:tiny_tnt.
Tiny tnt can pass Entropinnyum's monitor and generate mana, but it can't be tracked in ethnic tnt judgement.So it will generate 6500 mana instead of 3 mana.

Suisuroru:
this bug may caused by the custom tnt like tiny tnt (from AE2) can't be recognized as TntBlock.
Caused by(maybe):
movingBlockEntity.getMovedState().getBlock() instanceof TntBlock) {Xplat/src/main/java/vazkii/botania/common/helper/EthicalTntHelper.java 121:

TheRealWormbo:
Can you please look at #4769 to check if that implementation would fix the issue?

Suisuroru:
PR#4768 works well, but I think a config is a better choice for server operators, if they find a tnt can't be reconized as TntBlock in server operating ,their can fix it themselves, and this also provides a more open space for adjustment in their personal way.
I also suggest that it is needed a interface to change the botania:rfgenerator 's mana conversion efficiency. Infact the rate 1:10 is not well for every Industry mod.Like mek ,there is too low efficiency to generate energy , it make use mana to generate FE is no meaning for normal player in game.







&nbsp;




------------------&nbsp;Original&nbsp;------------------
From:                                                                                                                        "VazkiiMods/Botania"                                                                                    ***@***.***&gt;;
Date:&nbsp;Thu, Nov 28, 2024 04:14 PM
***@***.***&gt;;
***@***.******@***.***&gt;;
Subject:&nbsp;Re: [VazkiiMods/Botania] Custom TNT ethical fix (PR #4768)





 
Can you please look at #4769 to check if that implementation would fix the issue?

Suisuroru:
I think it is solved in #4769

PR #4769, opened 2024-11-28
Block tag for unethical TNT check

TheRealWormbo:
Resolves #4768 (AE2 Tiny TNT duping test) but in a better way.


Suisuroru:
It works well, but I think a config is a better choice for server operators, if they find a tnt can't be reconized as TntBlock in server operating ,they can fix it themselves, and this also provides a more open space for adjustment in their personal way.
I also suggest that it is needed a interface to change the botania:rfgenerator 's mana conversion efficiency. In fact, the rate 1:10 is not well for every Industry mod.Like mek ,there is too low efficiency to generate energy , it make use mana to generate FE is no meaning for normal player in game.

TheRealWormbo:
Being a block tag means it is by definition configurable. You need to make a data pack, which you probably have anyway if you plug mods together manually.

PR #4770, opened 2024-11-30
[Issue-4641]Provide a way to get the player's head by uuid

Suisuroru:
This provide the way to get the player's head by using uuid.
It used mojang's API, please enter right uuid or user name to build the head.
Because of the uuid's head-building is based on mojang's API, if you give a fault uuid, it may cause lag now.(this is my test log,I write a unreal uuid to build the head,[latest.log](https://github.com/user-attachments/files/17966539/latest.log))

It fixed in the later commit by cache the uuid and do not load the uuid's skin for repeating times.(Suisuroru#1)


Suisuroru:
I've tried to fix the lagging issue in the case of entering the wrong uuid, now both correct and incorrect uuid will only cause one peak lag on initial load instead of repeated small lags

Issue #4793, opened 2025-02-08
10x higher player lag when mana band not in trinket slot


Luky05005:
### Mod Loader

Both Fabric and Forge (I confirm that I have tested both loaders and will specify both loader versions below)

### Minecraft Version

1.20.1

### Botania version

1.20.1-446-fabric 1.20.1-446-FORGE

### Modloader version

fabric loader: 0.16.10 api: 0.92.3+1.20.1

### The latest.log file

Fabric: https://mclo.gs/QgxsXuM  Forge: https://mclo.gs/ytl5Eua

### Issue description

Mana consuming items that take mana from a mana storage item in the players inventory (not the trinket ring slot) cause lag.
Happens on both fabric and forge

### Steps to reproduce

1. Have a mana storage item in inventory (band of mana, greater band of mana, mana table, mana mirror)
2. Equip a mana consuming accessory
3. Run profiler ([Observable](https://modrinth.com/mod/observable))
4. Player Causes about 10x higher lag than before

### Fabric
Mana Band in trinket slot
![Image](https://github.com/user-attachments/assets/e6b466cc-7ee6-4ef4-b014-8ce7699d4403)

![Image](https://github.com/user-attachments/assets/89f35d29-c1dd-4fb6-b9cb-387cc2cfb64a)

Mana band in inventory
![Image](https://github.com/user-attachments/assets/947ba0f8-c486-4f23-9127-b62df3e59d09)

![Image](https://github.com/user-attachments/assets/2ff94ba4-5309-4801-8557-499fd2282d65)

### Forge
Mana band in trinket ring slot

![Image](https://github.com/user-attachments/assets/8b84f93c-a0d2-40f8-b480-7f3886554d0a)

![Image](https://github.com/user-attachments/assets/1982160b-ea1c-4ea6-97b3-82b610730178)

Mana tablet in inventory

![Image](https://github.com/user-attachments/assets/5af92b36-00de-45ab-9bb3-7f9ff9aa3f8c)

![Image](https://github.com/user-attachments/assets/89bd74e1-ed69-4c1d-87cc-0b2d4cd18e83)

### Other information

I first noticed this on my server, because players were causing the most lag, but this also happens in singleplayer.
Also I have a spark profile with this issue present on a server [here](https://spark.lucko.me/glxVTPJy3F?hl=19184), however there are many more mods loaded, but i think you can still find something. Like in the spark profile it looks like it has something to do with inventory updates

[...]

PR #4794, opened 2025-02-10
Skip Advancement checks for ManaItem updates

jtuc:
Advancement checking is expensive, and mana items will often update every tick.

I don't think this will interfere with any actual advancements? The bauble advancement uses and ImpossibleTrigger and is manually granted in the BaubleItem onEquipped method. Worst case scenario, the advancement will just be delayed until a non-mana item triggers the check.

Should address #4793

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/mixin/ServerPlayerMixin.java
+++ b/Xplat/src/main/java/vazkii/botania/mixin/ServerPlayerMixin.java
@@ -10,6 +10,8 @@ package vazkii.botania.mixin;
 
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import net.minecraft.world.item.ItemStack;
 
 import org.spongepowered.asm.mixin.Mixin;
 import org.spongepowered.asm.mixin.injection.At;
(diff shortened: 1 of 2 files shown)


Xplat/src/main/java/vazkii/botania/mixin/ServerPlayerMixin.java
```diff
@@ -39,4 +41,14 @@ private void onTick(CallbackInfo ci) {
 		}
 	}
 
+	@Mixin(targets = "net/minecraft/server/level/ServerPlayer$2")
+	public static class ContainerListenerMixin {
+		@Inject(at = @At("HEAD"), method = "slotChanged", cancellable = true)
+		private void onSlotChanged(AbstractContainerMenu containerMenu, int slotIndex, ItemStack itemStack, CallbackInfo ci) {
+			// ManaItems update frequently - skip advancement processing for them
+			if (XplatAbstractions.INSTANCE.findManaItem(itemStack) != null) {
```
jtuc, line 59: Possibly worth hardcoding item classes instead for more performance? Not sure how annoying that list would be to maintain...

TheRealWormbo: Ideally this would only skip the check if the item type did not change, as the same system is also responsible for unlocking recipes.

jtuc: I don't think there's a good way to check the prior inventory - at the point this listener is invoked, the inventory has already been updated, so we can't tell if it was just an NBT change or a new item.

Since only the manaTablet triggers recipe unlocks, I think special-casing the first manaTablet per player resolves this?
If modpacks/addons add different botania advancements/recipes, they'll still be triggered the next time a player moves a non-manaItem in their inventory.

jtuc:
As discussed on #4793, it probably makes more sense for users to install a generic achievements optimizing mod instead of each mod having their own (potentially conflicting) mixins.

PR #4795, opened 2025-02-10
Floral fertilizer can produce shimmering mushrooms

TheRealWormbo:
It spawns flowers on blocks in the `minecraft:dirt` tag (except in the nether), and mushrooms on blocks in the `minecraft:mushroom_grow_block` tag. If the block matches both conditions, flowers are preferred over mushrooms at a ratio of 2:1.

This commit also removes the random render offset for the mushrooms to match their interaction shape.


TheRealWormbo:
Confirmed to be working properly. Since shimmering mushrooms only correspond to a single petal, a fertilizer application is now guaranteed (assuming there is enough space) to produce the equivalent of 5 to 7 petals, which still creates 3 or 4 flowers, if only flowers are possible in the area. This makes it "petal-positive" in all cases except with extreme space constraints.

PR #4796, opened 2025-02-10
Add item frame filter expansion

TheRealWormbo:
Filter items in glow item frames are expanded into a list of items if they are a bundle, shulker box, or similar item with non-empty content.

- Hopperhock, Rannuncarpus, and Interceptor filter for the exact container item *with the same contents* if it's in a regular item frame, and for the contents items if it's in a glow item frame
- Corporea Funnel treats item counts of expanded filter lists as random weights, similar to having multiple frames on the funnel block. Rotation of the item still defines the number of items to request.


TheRealWormbo:
Manaseer Monocle received an upgrade to show the contents of supported item containers in item frames, and to render a background behind its HUD overlay.

TheRealWormbo:
Lexicon entries for Hopperhock, Corporea Funnel, and Corporea Interceptor have been updated. Interceptor now briefly mentions the Hopperhock for filtering, like Rannuncarpus already did. Hopperhock got a new paragraph (I checked that it still fits the page without text resizing), and Corporea Funnel got a whole new page to also cover the stack sizes = random weights approach.
I updated en_us and de_de, and I believe en_gb doesn't need updating for these changes. Changelog mentions multi-item filtering and the new monocle item frame interaction.

artemisSystem:
very nice, thanks!

PR #4803, opened 2025-02-23
Spongepowered as mirror for Forge maven

TheRealWormbo:


TheRealWormbo:
For testing during the next planned or unplanned downtime of the Forge maven repository.

Abbie5:
Might NeoForged's maven (`https://maven.neoforged.net/releases/`) be a better option?

TheRealWormbo:
Maybe? But we use Spongepowered as source for Mixin already.
This entire PR might get obsolete anyway if I get around to switching to the NeoForged legacy toolchain for the Forge and maybe even vanilla parts of our build. In that case we'd definitely be using the NeoForged maven repo as well.

TheRealWormbo:
With the switch to ModDev Legacy Forge this has become redundant.

PR #4822, opened 2025-04-01
compress PNG textures

AnAwesomGuy:
used oxipng with zopfli to losslessly compress PNGs
`oxipng -Zsro max --zi 50 --nc *` run in the `Xplat/src/main/resources/assets/botania/textures` folder
`-Z` enables Zopfli, `-s` strips unnecessary metadata, `--nc` keeps the color palettes (textures were NOT changed to indexed and kept as RGB)


williewillus:
I'm pretty sure we already optimized the pngs in the mod before, so repeating the process is just reapplying a lossy function and losing more quality for little gain.

AnAwesomGuy:
this is fully lossless, zopfli is lossless, no quality is lost at all

TheRealWormbo:
I can confirm we have a not insignificant number of textures that zip quite well. Maybe the last optimization pass was before the texture update?

Also tested the output now. The compiled jars are 400kB smaller, and testing some of the files I can't find any pixel differences.

clarfonthey:
Definitely agree that losslessly compressing the PNGs is an easy win, since as mentioned it doesn't actually affect anything but making the file size smaller.

PR #4835, opened 2025-04-30
tr_tr.json Turkish Localization

RuyaSavascisi:


TheRealWormbo:
Please only target the 1.20 branch unless it's specifically help with porting to the new Minecraft version.

PR #4837, opened 2025-05-04
Fix NPE attenuating sounds if temporary field is unset from elsewhere

LostLuma:
Fix NPE in sound engine mixin if the `tmpSound` field is unset while sounds are attenuated by storing it locally.  
This would probably be better off using [MixinExtras' Share](https://github.com/LlamaLad7/MixinExtras/wiki/Share), but I was unsure whether I should add a new dependency.

A user reported this issue to a different mod I maintain, however since it keeps all its sound engine operations on the main thread it shouldn't be the cause. See the issue here: https://github.com/juliand665/Dynamic-FPS/issues/257


TheRealWormbo:
Hmm, yes, that code should definitely get a rewrite in the 1.21 port, where we already added MixinExtras.

PR #4868, opened 2025-07-16
Faster load

boubou19:
Was measuring the preinit of the pack, saw botania spending 1.8s in preinit. Reducing the surface for forge reflection brought it down to 0.5s.


f-raZ0R:
Sir this is a wendy’s

PR #4873, opened 2025-07-22
[Example] Save information about player/fakeplayer inside of spreader/burst

SaloEater:
We, players, have an issue in CtE2 where they use Flan that allows us to claim land and provides some sort of land protection.
But Flan heavily relies on game events which Botania doesn't emit.

Before changing all parts of code, this is an example how event emitting could be implemented in Botania, so need to hear if you, guys, are fine with approach from this PR

It works flawlessly, here is the proof: 

https://github.com/user-attachments/assets/d12b6490-8543-455e-82cd-4e3ca55b179a


TheRealWormbo:
There are currently multiple issues with this PR:

- It doesn't build because the Fabric version does not fully implement the extended `XplatAbstractions` interface.
- We do not intend to introduce player ownership for mana spreaders, as they are supposed to work somewhat similarly to dispensers.

Good points in this PR to maybe improve upon:

- Mana gun bursts should definitely be owned by the player using the item, although not to the point that a fake player would be introduced to keep that information.
- Forge events for things like breaking blocks absolutely make sense to check.

SaloEater:
> * Forge events for things like breaking blocks absolutely make sense to check.

Unfortunately, block breaking event requires a not nullable player - https://mcstreetguy.github.io/ForgeJavaDocs/1.20.1-latest/net/minecraftforge/event/level/BlockEvent.BreakEvent.html

Maybe this will influence you to change your mind? :)

TheRealWormbo:
> Maybe this will influence you to change your mind? :)

Not really, and I don't understand why you keep insisting. It merely means the events would be restricted to player-owned bursts. It's really the same as explosions, e.g. from a TNT dispenser – they *can* be owned by a player (or mob), but don't have to be.

SaloEater:
>  to player-owned bursts. 

I can see you point of view, let me talk to land protection mod developer, maybe we should add a feature of adding tile entities to claim members too.

SaloEater:
@TheRealWormbo hi, may I ask your feedback on this new implementation?

TheRealWormbo:
You're adding all these events now. Is that really going to be that useful? It's all just block placing, removing, or replacing.

PR #4879, opened 2025-07-31
Unbreaking enchant fix on magical tools

LewisSaber:
This pr aims to fix one of botania issues
https://github.com/VazkiiMods/Botania/issues/2907
Basically when dmaage is done to tool via damageItem it ignores Unbreaking because unbreaking is calculated further in pipeline, and will only affect tool if you have no mana to repair it
I believe its oversight, because when you first dmg tool and then get mana source to repair it, it will consider unbreaking(because dmg to be repaired was done with consideratino of unbreaking)

To put it simple, currently it is like this:
Lets say it costs 10 mana to repair 1 durability
If we have to beak 1000 blocks via damageItem:
do you have enough mana? substract 10k mana, return

not enough mana? deal 1k dmg to tool, 
tool: oh I have to get 1k dmg? well I got Unbreaking 3, I should take 250 damage only!
After breaking 1k blocks, you get some mana source to inventory to repair tool
you have to repair 250 durability, that is 2500 mana to repair


LewisSaber:
oopsie

PR #4880, opened 2025-07-31
Unbreaking fix but for real

NEstoll:
Allows the unbreaking enchant to reduce mana costs. Currently uses the default unbreaking logic for all items, making it more efficient than normal on armor pieces, but I'd argue that's on the armor for using ToolCommons


TheRealWormbo:
I don't like the double Unbreaking effect if the player is out of mana.

What if you first check if the item has any Unbreaking (outside the loop, if amount is initially > 0), and if so, check if enough mana is available, and only then potentially replace mana consumption with the Unbreaking effect?

PR #4881, opened 2025-07-31
don't delete item in locked slots

### LINKED ISSUE #4824: Crafty Crate voids unused ingredients after applying pattern
### Mod Loader

### Minecraft Version

1.20.1

### Botania version

448

### Modloader version

(any)

### Modpack info

(n/a)

### The latest.log file

(n/a)

### Issue description

When locking slots of a Crafty Crate by applying a crafting pattern, the items in those slots stay there. Then, when the crate next crafts a result, those slots are emptied, but the items are not returned.

### Steps to reproduce

1. Place Crafty Crate with some means to manually input items, e.g. a hopper pointing into the crate.
2. Fill in 2-8 of an item that can be crafted in a 1x1, e.g. a flower.
3. Apply the 1x1 crafting pattern to the crate. It will automatically execute the crafting operation, since it's now completely full.
4. Inspect the items dropped by the box (just a single crafting operation's result) and the items remaining in the box (holding a wand; the crate is empty)

### Other information

Before applying the pattern, the Crafty Crate contains 5 cornflowers

After applying the pattern, the crate is empty, but only dropped a single piece of blue dye

NEstoll:
should fix #4824

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CraftyCrateBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/CraftyCrateBlockEntity.java
@@ -175,7 +175,7 @@ public class CraftyCrateBlockEntity extends OpenCrateBlockEntity implements Wand
 				ItemStack s = remainders.get(i);
 				ItemStack inSlot = handler.getItem(i);
 				if ((inSlot.isEmpty() && s.isEmpty())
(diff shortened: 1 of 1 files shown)


williewillus:
Hmm.....I'd rather the crate just eagerly eject items from locked slots when the pattern changes. Keeping it in this sort of hidden state is kinda weird.

TheRealWormbo:
Fixed via eaf5872 as per williewillus's suggestion

PR #4884, opened 2025-08-03
Update zh_cn.json

Dawnwalker666:
The scope of the translation includes all commits about en_us.json from Mar 30, 2025 to Jul 22, 2025.

In the past 5 months, the language files have been committed less, and I'm not a traditional Chinese user. So I plan to wait for more commits before updating the zh_tw.json file. This is just for me personally, not to stop others from fixing the files.


Xplat/src/main/resources/assets/botania/lang/zh_cn.json
```diff
@@ -1672,7 +1672,7 @@
   "item.botania.slime_bottle": "史莱姆罐子",
   "item.botania.star_sword": "召星者",
   "item.botania.exchange_rod": "移壳杆",
-  "item.botania.magnet_ring_greater": "巨型磁化环",
+  "item.botania.magnet_ring_greater": "高级磁化指环",
```
Dawnwalker666, line 1675: https://github.com/CFPAOrg/Minecraft-Mod-Language-Package/pull/5451

PR #4888, opened 2025-08-05
Reach ring

### LINKED ISSUE #20484: Re-equiping Ring of Far Reach increases Reach Length permanently, stacking with every time
### Your GTNH Discord Username

_No response_

### Your Pack Version

2.8.0 Beta-1

### Your Server

SP

### Java Version

Java 21

### Type of Server

None

### Your Expectation

Ring of Far Reach would set Reach Range to some higher value when you have it equipped.

### The Reality

If you remove and equip the Ring many times you can stack Reach Length to over a hundred blocks. Re-entering the world sets Reach to the expected Reach amount.
More specifically, it seems that having a Ring equipped gives some Reach, which stays as long as you have it equipped, and each action of equipping it gives some more Reach, untill re-entering the world.
I think that because entering the world with the Ring already equipped and equipping and de-equipping the Ring a single time - not having it equipped at the end - has the exact same Reach above normal. If you enter the world without equipped Ring and put it on, you'll have more reach that either of the 2 previous cases. If you keep re-equipping it, the Reach will increase even more.

### Your Proposal

Fix to expected behaviour

### Final Checklist

- [x] I have searched this issue tracker and there is nothing similar already. Posting on a closed issue saying the bug still exists will prompt us to investigate and reopen it once we confirm your report.
- [x] I can reproduce this problem consistently by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.
- [ ] I have asked other people and they confirm they also have this problem by follow the exact steps I described above, or this does not need reproducing, e.g. recipe loophole.

DarkShadow44:
Fixes https://github.com/GTNewHorizons/GT-New-Horizons-Modpack/issues/20484

Built on top of https://github.com/GTNewHorizons/Botania/pull/67, hopefully this time fixed for good..

To summarize, the ring has a few issue:
Primarily, it didn't get applied after a relog, that's why it got handled in `onWorkTick` as well.
Problem is, after unequip we might get another tick, which might reapply the effect erroneously.
The current code deal with that by expecting that tick and handling it.
Additional issue is the shared state of client/server in SP (`toRemoveItems`)
Since expecting (and needing) the ticks is a pain to deal with, I replaced it with a short delay - those weird ticks after unequip are now ignored and don't execute any logic. Should make it easier to understand and hopefully fix this issue.

@Shadow1590 Since you are the original author, mind taking a look as well?


TheRealWormbo:
Wrong place. Please point your GTNH PRs at their repository.

DarkShadow44:
> Wrong place. Please point your GTNH PRs at their repository.

Sorry, I accidentally filed the PR against the parent repo. :confounded:

PR #4914, opened 2025-08-24
1.21.1 runes rebalance

TheRealWormbo:
The goal of this change is to open up most runes without having to not just visit the nether, but actually obtain nether wart (or blaze rods) from a nether fortress or housing units bastion remnant. Instead of randomly fortress-gating some of the tier 2 and 3 runes though the fire rune, this change makes a trip to the nether mandatory only for the tier 3 runes. Interestingly, this only seems to make two new items nether-gated, compared to 1.20.1: the damage lens and the entropic lens.

However, if those were the only changes, terrasteel and most of the Alfheim content would be available without going to the nether first. Only the Gaia fight arena setup would require going to a nether fortress as part of obtaining a beacon.

Overview of changes:
- Tier 1 (elemental) runes: recipes were generally simplified to reasonably obtainable overworld ingredients
  - Rune of Water: replaced fishing rod with lily pad (unstackable ingredient is awkward)
  - Rune of Fire: replaced nether wart with a choice between coal and charcoal (technically removes nether-gating because netherrack for nether brick is available in the overworld at ruined portals)
  - Rune of Earth: replaced coal block with mud block (since coal is used in the fire rune recipe now)
- Tier 2 (season) runes: reduced repeated ingredients to single instances and each of them required three different non-rune ingredients now
  - Rune of Spring: added dandelion as third ingredient (maybe the wheat should be replaced with a crop that is actually available in spring, like carrot)
  - Rune of Autumn: added beetroot as third ingredient (it's an autumn crop and has no uses otherwise)
  - Rune of Winter: replaced cake with sweet berries (a cold climate crop, also cake is much more complicated to make than any other season rune ingredient)
- Tier 3 (sin) runes: added Mana Quartz item as additional ingredient (nearly no change in minimum mana requirement, but gates these runes behind a visit to the nether and increases complexity of automated material production)
- Terrestrial Agglomeration Plate: replaced Block of Manasteel in recipe with Block of Manaquartz (gates the recipe behind a nether visit, though less difficult than the nether wart gating through the old fire rune recipe)


TheRealWormbo:
Note: While the Runes of Envy, Lust, and Wrath were not nether-gated before, it seems all items that use them, except the two mentioned mana lenses, depended on one of the other nether-gated runes (Fire, Spring, or Autumn) instead. As a result the nether gating for such items doesn't change.

On the other hand, many interesting items become available before having made a nether portal. Some examples:
- various rods (fire, cobble)
- some baubles (e.g. Tectonic Girdle, Benevolent Goddess's Charm, Cirrus Amulet, Pyroclast Pendant)
- potency lens
- various special flowers (e.g. Thermalily, Gourmaryllis, Exoflame, Marimorphosis, Agricarnation, Jaded Amaranthus)

TheRealWormbo:
Probably slightly too invasive for now. Will rework and open a new PR.

PR #4915, opened 2025-08-24
skies rod uses acceleration rather than set speed

NEstoll:
works more like rockets. Also increases speed, since the proficiency bonus is currently impossible while gliding


TheRealWormbo:
Proficiency bonus is technically possible with mods like Elytra Slot already. And considering we are likely to turn the tiara's weird gliding into elytra-like gliding in the 1.21.1 release, it might be worth considering to not add that multiplier.

NEstoll:
Even with proficiency bonus being possible, it still seems better to have the boosted version be better than rockets, rather than requiring the full mana weave set to just match them.

TheRealWormbo:
Merged manually.

PR #4916, opened 2025-08-25
1.21.1 mana infusion rebalance

TheRealWormbo:
The goal of this PR is to cover a few more controversial changes to mana infusion with the aim to get rid of many alchemy catalyst cycles and all conjuration catalyst recipes. These feel cheap and uninspired in terms of game design, and are partially outdated due to vanilla changes. Instead, the focus is moved towards a more natural availability of the resources these recipes intended to provide.

Overview of changes:

- All conjuration catalyst recipes have been removed. The block may get renamed to "Deconstruction Catalyst" in the future and is now responsible for the deconstruction recipes that formerly were covered by the alchemy catalyst.
- Various conversion cycles and some individual recipes have been removed:
  - Froglight cycle – vanilla farms usually have no issues getting all three frog types anyway
  - Stone bricks -> chiseled stone bricks – we have the stonecutter, and it's not a difficult crafting recipe either
  - Fish cycle – you get all of them from fishing (or from looting guardian farms)
  - Shrub cycle – resources not needed in Botania, fern can be duplicated manually with bone meal and shears, dead bushes generate from decaying Hydroangeas
  - Flower cycle – just makes exploration for decorative items unnecessary
  - Berry cycle – sweet berries and glow berries require a but of exploration, but should generally not be that hard to find, apples can be farmed already (some way to speed that up would be nice)
  - "1.17 stones" cycle – no need to have that separated from the other stone types
- Adjustments to existing cycles:
  - Log cycle – The 8 existing log types have been rearranged in the cycle to alter between tree types with "simple" and "complex" tree shapes. This allows any log type that would be a bit tricky to farm to be created by just one conversion from a log type with a reasonably easy to handle tree shape.
  - Sapling cycle – Rearranged to have the same order as log cycle, and increased to the same cost as the crop cycle. This cycle exists to make the tree types available, even though it undercuts exploration. Similar to the crop cycle it will stay around for now, but both might become GoG-only eventually.
  - Stone cycle – The cycle now properly starts with andesite (for seamless entry via the stone -> andesite conversion), and tuff and calcite have been appended to the end. Deepslate is not part of the cycle, making it only endgame-renewable via an endstone purification loop with the Pure Daisy. Then again, it's abundantly available in regular worlds.
- Added a nether "stones" cycle – Starting with a basalt generator, this cycle converts basalt into netherrack, netherrack into blackstone, and that back into basalt. Alternatively, blackstone can also be bartered for from piglins, making netherrack available via two conversion steps.
- Added alchemy recipe for wart block -> nether wart (1:1) – This makes brewing available without having to visit a nether fortress or housing units bastion remnant. Players now also have an option to farm huge crimson fungi instead of nether wart directly.
- Reworked deconstruction recipes using the conjuration catalyst:
  - added lossless deconstruction for snow block, bamboo block, and melon
  - existing deconstruction recipes (except for clay block) are now lossy, only returning 3 items
  - additional lossy 2x2 deconstruction recipes for: nether bricks, dripstone block, honeycomb block, magma block, amethyst block, red nether bricks (yields only 2 nether brick)
  - special deconstruction recipes for: snow layer -> 1 snowball, small/medium/large amethyst bud/amethyst cluster -> 1..4 amethyst shards
- Other changes to compensate for removed recipes:
  - Horn items and drum blocks have a chance (currently 5%) to break their blocks as if the proper tool was used, which is shears for horn/drum of the wild/canopy, and shovel for horn of the covering. (As a bonus horns and drums track usage statistics now.)
  - Floral Fertilizer can be applied to various "mundane" plants to make them spread, a bit like bone meal does on Bedrock edition. This currently covers all small flowers (except wither rose and torchflower), lily pads, spore blossoms, as well as brown and red mushrooms. (Mushrooms slowly spread on their own and can be farmed via their huge forms, but… options.)
  - Dispensers can apply floral fertilizer to spreadable plants. (It cannot be applied to make more mystical flowers or shimmering mushrooms this way.)


TheRealWormbo:
There are still some open points to figure out, like:

- What about the Conjuration Catalyst block itself and its documentation? Will it be renamed to "Deconstruction Catalyst" and have its lexicon entry completely replaced?
- The crop and sapling cycles still undercut exploration. In case of the crop cycle it's for somewhat important resources, as you need various crops for things eventually. (Especially with the rune rework.) In case of the sapling cycle it's more of a "nice to have", as you are not required to use any particular log/leaves/sapling type for any kind of automation.
- Fern and dead bush can no longer be properly automated, even though they are not required for anything in Botania.
- Converting blaze rod to nether wart, converting dirt to coarse dirt, and "unpoisoning" potatoes are still there, but are removal candidates. (Poisonous potatoes might become tiny… and probably grumpy, or "angery"?)

(GoG item availability is out of scope for now.)

TheRealWormbo:
Too many changes, too invasive. Will rework and open a new PR.

PR #4917, opened 2025-08-26
Updated PT_BR localization

PrincessStellar:


TheRealWormbo:
This PR makes a lot of changes without explaining why. Especially the way it "untranslates" some Botania terms like "Dreamwood" seems like a step backwards. I'm going to reject this for now.

PR #4918, opened 2025-08-26
Updated PT_BR localization

PrincessStellar:


Coffee-0xFF:
This patch causes some rendering issues in the mod’s Patchouli book (no space for the text). It also has some inconsistencies with block names compared to vanilla. [See screenshots](https://imgur.com/a/8h7OwPb)

I haven’t fully reviewed it carefully.

I also don’t see the point of redoing almost the entire pt_br.json if only one line was changed in en_us.json since #4874.

TheRealWormbo:
I'm going to reject this PR because it seems to copy 1.21.1 changes to the 1.20.1 version, which is wrong. Also, as pointed out already, there are a lot of changes, so I'd at least expect a description of why so much was changed.

PR #4926, opened 2025-09-06
update pt_br.json - 1.20.1

Coffee-0xFF:


TheRealWormbo:
~~This is not a change that will happen in 1.20.1 anymore.~~

TheRealWormbo:
Nevermind, I should wake up before doing these things.

PR #4936, opened 2025-10-01
Use fake player instead of null player

baerchen201:
This allows whitelisting block placement events in chunk claiming mods, as some of them deny all block changes if no player is provided.


TheRealWormbo:
This is likely not going to be the implementation we will use. Also, how do these claiming mods work with dispensers that place blocks (powder snow, water, lava, etc.) or farmer villagers planting crops?

TheRealWormbo:
Going to actually close this, as we will not be changing that kind of behavior in 1.20 anymore.

Please make a bug report for the claiming mod to have more reasonable default settings, and maybe open an enhancement issue for Botania. This is not the only Botania feature recent versions of claiming mods interfere with.

PR #4943, opened 2025-10-19
Improve mana burst handling of potentially unloaded chunks

### LINKED ISSUE #4928: Catalyst overlay in fabulous pool renders in weird way
### Mod Loader

Fabric

### Minecraft Version

1.21.1

### Botania version

dev

### Modloader version

any

### Modpack info

_No response_

### The latest.log file

n/a

### Issue description

Something seems wrong with the way catalyst overlays in mana pools render. They seem to make that area of the pool transparent in a weird way.

Image

Image

### Steps to reproduce

_No response_

### Other information

_No response_

TheRealWormbo:
- bursts track their source dimension so they don't attempt to look at a completely unrelated position in a different dimension when looking for their source spreader
- redirective lens can only operate in its source dimension, and only if its source spreader is still loaded
- warp lens adds a "post-teleport" chunk loading ticket when the burst is teleported via a force relay, similar to entities traveling through an end gateway (this does not enable block entity ticking on its own)
- bore-warp lens can only teleport items in its source dimension; it adds a "post-teleport" chunk loading ticket when teleporting items to the source spreader

(fixes #4928)

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/api/internal/ManaBurst.java
+++ b/Xplat/src/main/java/vazkii/botania/api/internal/ManaBurst.java
@@ -9,11 +9,14 @@
 package vazkii.botania.api.internal;
 
 import net.minecraft.core.BlockPos;
+import net.minecraft.core.GlobalPos;
 import net.minecraft.world.entity.projectile.ThrowableProjectile;
 import net.minecraft.world.item.ItemStack;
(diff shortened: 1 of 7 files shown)


TheRealWormbo:
Spectranthemum item teleportation also gets a ticket, while we're at it.

PR #4945, opened 2025-10-23
Fix a typo in the rainbow rod entry

NathanSnail:
Rainbow rod is missing an "a" or "the" in the guide, "the" is used for most of the other items so I used that.


TheRealWormbo:
Covered in the most recent commit.

Issue #4949, opened 2025-11-05
Broken gif on website


Goldenlion5648:
### Mod Loader

Forge

### Minecraft Version

-

### Botania version

-

### Modloader version

-

### The latest.log file

-

### Issue description

The tree farm gif on the website doesn't show since gfycat shutdown https://botaniamod.net/index.html

<img width="595" height="345" alt="Image" src="https://github.com/user-attachments/assets/3822604d-9e43-4390-9d8b-04792fd648e6" />

### Steps to reproduce

1. Open https://botaniamod.net/index.html
2. See gif is empty

PR #4950, opened 2025-11-05
fix tree farm gif

Goldenlion5648:
I was having trouble building the site (seemed to be built using an older ruby/ jekyll version), but in theory this should work?


Goldenlion5648:
Fixes #4949

PR #4959, opened 2026-01-12
Add getProvidedBlock method to BlockProvider

ChiefArug:
This lets people determine what a block provider provides without iterating the registry.

My plan for this is compatibility with Tinkers Construct exchanging ability, which replaces mined blocks with the block in the offhand, and adding the ability to provide blocks from things such as the black hole talisman. To replace the block Tinkers needs to know what block its placing.

This is not a breaking change to the API as the method has a default implementation, so the two published addons that implement this interface won't break.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/api/item/BlockProvider.java
+++ b/Xplat/src/main/java/vazkii/botania/api/item/BlockProvider.java
@@ -40,7 +40,10 @@ public interface BlockProvider {
 	int getBlockCount(Player player, ItemStack requestor, Block block);
 
 	/**
-	 * @return the block that this provides. null if the block is not easily determinable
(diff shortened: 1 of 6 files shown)


williewillus:
What if an item wants to provide multiple types of Blocks?

ChiefArug:
It can either choose one to return, or simply return `null` to indicate that what it provides isn't as simple as a single block type (which is what the hand of ender does in this PR).

The intended usecase of this is for things that can consume any block and so need to be told what block to consume, so changing the return type to a list doesn't make sense for that usecase (the additional entries in the list will just end up being ignored).

ChiefArug:
Updated the javadoc to better clarify what to do if it can return multiple blocks.

PR #4963, opened 2026-02-13
Rewrite Dominant Spark transfer

### LINKED ISSUE #4833: Spark Network seizes up when charging items that drain mana constantly.
### Mod Loader

Forge

### Minecraft Version

1.20.1

### Botania version

1.20.1-448

### Modloader version

Forge-47.3.0

### Modpack info

Custom

### The latest.log file

https://gist.github.com/RiArgaet/111f6f6611e219ac199d61018a5c679c

### Issue description

When using a spark network with a mana battery connected to a dominant spark, it will select 1 pool or more depending on connected bellows then *only* take from them, ignoring all other pools, after the main poll will empty.

### Steps to reproduce

1. Join Multiplayer world.
2. Use the simplest mana battery setup with 6 or more pools.
3. Wait for the pools to fill up slightly, around halfway.
4. Throw in an item that takes mana constantly such as a shatterer.
5. Watch the spark mana flow.

### Other information

The issue is remedied slightly by the fact you can simply take then replace the dominant augment for it to choose a next set of pools, but when charging something like the Terra Shatterer, it is a massive nuisance.

We are using "Botania Editor" In this modpack, however I do not believe it is the cause due to the fact it does not tamper with spark code, do test with it on if unable to replicate though.

NEstoll:
Fixes #4833

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -52,6 +52,7 @@ import vazkii.botania.network.clientbound.BotaniaEffectPacket;
 import vazkii.botania.xplat.XplatAbstractions;
 
 import java.util.*;
+import java.util.stream.Collectors;
 
 public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 	private static final int TRANSFER_RATE = 1000;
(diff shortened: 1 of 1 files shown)


williewillus:
given how sensitive this system has been, it would be nice to have some gametests to assert the correctness of things here

NEstoll:
Do you have a definition for what correct would be? Because this PR does change the behavior somewhat

PR #4964, opened 2026-02-15
Restrict dominant sparks to pulling from a single spark per tick

cjlarose:
The changes in https://github.com/VazkiiMods/Botania/pull/4963 generally fixed behaviors associated with establishing and re-establishing mana transfers to dominant sparks, but I think introduced an unintended consequence: it made it possible for dominant sparks to receive `TRANSFER_RATE` mana every tick from each unaugmented spark in range.

Prior to the change, a dominant spark could receive at most `TRANSFER_RATE` mana from a single pool per tick. This change restores that behavior.

On my own build, this issue was immediately apparent just because of the number of particles moving around. My dominant pool (on the left, surrounded by glass) had only a small mana drain from a Tainted Blood Pendant I was wearing, but tons of particles were moving towards the dominant pool. I suspect a bunch of particles were coming from 0-mana transfers to the dominant pool. In other words, the first couple of sparks in a tick fill up the dominant pool, but the remaining sparks in `incomingTransfers` still end up being processed and transfer `0` mana but still spawn particles.

Normal behavior (prior to #4963 and with the proposed fix):

Current behavior (introduced by #4963)

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -60,7 +60,8 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 
 	private final Set<ManaSpark> outgoingTransfers = Collections.newSetFromMap(new WeakHashMap<>());
 
-	private final ArrayList<ManaSpark> inboundTransfers = new ArrayList<>();
+	@Nullable
+	private ManaSpark inboundTransfer = null;
 
 	private boolean shouldFilterTransfers = true;
 	private boolean receiverWasFull = true;
@@ -215,34 +216,26 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 				receiver.receiveMana(-manaSpent);
 			}
 		}
-		if (!inboundTransfers.isEmpty()) {
-			int manaNeeded = Math.min(TRANSFER_RATE * inboundTransfers.size(), tile.getAvailableSpaceForMana());
-			int count = inboundTransfers.size();
-			int manaRecieved = 0;
+		int manaNeeded = Math.min(TRANSFER_RATE, tile.getAvailableSpaceForMana());
 
-			if (manaNeeded > 0) {
-				if (shouldFilterTransfers) {
-					filterTransfers();
-					shouldFilterTransfers = false;
-				}
+		if (manaNeeded > 0) {
+			if (shouldFilterTransfers) {
+				filterTransfers();
+				shouldFilterTransfers = false;
+			}
 
-				inboundTransfers.sort(Comparator.comparingInt(s -> s.getAttachedManaReceiver().getCurrentMana()));
-				for (ManaSpark spark : inboundTransfers) {
-					count--;
-					SparkAttachable attached = spark.getAttachedTile();
-					var attachedReceiver = spark.getAttachedManaReceiver();
-					if (attached == null || attachedReceiver == null) {
-						shouldFilterTransfers = true;
-						continue;
-					}
-
-					int gained = Math.min(attachedReceiver.getCurrentMana(), (manaNeeded - manaRecieved) / (count + 1));
+			if (inboundTransfer != null) {
+				SparkAttachable attached = inboundTransfer.getAttachedTile();
+				var attachedReceiver = inboundTransfer.getAttachedManaReceiver();
+				if (attached == null || attachedReceiver == null) {
+					shouldFilterTransfers = true;
+				} else {
+					int gained = Math.min(attachedReceiver.getCurrentMana(), manaNeeded);
 					attachedReceiver.receiveMana(-gained);
-					manaRecieved += gained;
+					receiver.receiveMana(gained);
 
-					particlesFrom(spark.entity());
+					particlesFrom(inboundTransfer.entity());
 				}
-				receiver.receiveMana(manaRecieved);
 			}
 		}
 
@@ -261,7 +254,7 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 
 	@Override
 	public void updateTransfers() {
-		inboundTransfers.clear();
+		inboundTransfer = null;
 		outgoingTransfers.clear();
 		switch (getUpgrade()) {
 			case RECESSIVE -> {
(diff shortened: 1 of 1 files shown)


TheRealWormbo:
Why would dominant sparks only pull from a single spark now? That heavily restricts mana throughput.

cjlarose:
> Why would dominant sparks only pull from a single spark now? That heavily restricts mana throughput.

Perhaps I misunderstand the intended behavior, but as I understand it, prior to https://github.com/VazkiiMods/Botania/pull/4963, it was certainly possible for a dominant spark to receive mana from more than one spark, just never more than one spark in a single tick.

I _think_ the proposed change restores that behavior. Mana throughput should be identical to the behavior prior to https://github.com/VazkiiMods/Botania/pull/4963

With the proposed change, a dominant spark can still pull from multiple sparks, it just picks one at random every tick.

cjlarose:
Re-reading the code prior to https://github.com/VazkiiMods/Botania/pull/4963, I think I might be wrong. There's definitely something unusual going on now with the number of particles I'm seeing, but the old code does actually seem like it permits multiple transfers from unaugmented sparks to a dominant spark in a single tick.

cjlarose:
After some more digging, I think I now understand what happened:

**The behavior after #4459**

PR #4459 introduced a queue-based mechanism (`transfersTowardsSelfToRegister`) for dominant sparks. Roughly,

1. When a dominant spark's pool went from full to not-full, it would populate a queue with all nearby unaugmented sparks
2. Each tick, the dominant spark would pop one spark from the queue and call `registerTransfer(this)` on it
3. That unaugmented spark would add the dominant spark to its `outgoingTransfers` set and transfer mana when it ticked
4. The dominant pool would fill (or nearly fill) from that transfer
5. When other unaugmented sparks ticked and checked their `outgoingTransfers`, `filterTransfers()` would remove the dominant spark because `isFull()` returned true
6. Result: Even with a queue of several unaugmented sparks, only one would successfully transfer per tick

For a dominant-sparked pool with a slow drain, this meant you'd typically see one particle stream per tick, carrying just enough mana to keep up with the drain.

**What #4963 changed**

PR #4963 changed the approach by having the dominant spark pull mana directly during its own tick, rather than waiting for unaugmented sparks to push:

1. The dominant spark finds all nearby unaugmented sparks and adds them to `inboundTransfers`
2. During the dominant spark's tick, it calculates how much mana it needs and evenly divides that amount among all unaugmented sparks
3. All transfers happen during the same tick

**The visual impact**

This is where I got confused. With the slow drain from my Tainted Blood Pendant:

- **Old behavior**: 1 unaugmented spark transfers ~2 mana per tick → 1 particle stream
- **New behavior**: 12 unaugmented sparks each transfer ~0.16 mana per tick (integer division means most transfer 0, but all attempt) → 12 particle streams

The total throughput is identical (both limited by the available space in the pool), but the number of concurrent transfers increased dramatically. I became so familiar with the single-particle-stream visual that seeing 12+ streams made me think something was wrong.

I don't necessarily think there's a bug. Visually, things look much busier now for a dominant-sparked pool with a slow drain that has many unaugmented sparked pools nearby. The new behavior enables the intended maximum throughput of 1,000 mana per tick per unaugmented spark when the dominant pool has sufficient capacity to receive it.

TheRealWormbo:
Hmm, good point about the much more dense particle streams. @NEstoll can you maybe have a look at that, since it was your PR?

NEstoll:
IMHO this is the more correct behavior (and is the way that 1 recessive spark to many unaugmented sparks works, particles and all). If there are too many particles, I would think the correct fix would be to just globally reduce the particles that sparks make when transferring mana, not anything specific to dominant sparks

cjlarose:
I'm gonna experiment with skipping particle spawning for transfers that move 0 mana and see if that helps at least

cjlarose:
Opened https://github.com/VazkiiMods/Botania/pull/4965 just to skip spawning particles for 0-mana transfers to dominant sparks, which helps a bit, but I think I'm on the same page as @NEstoll for a more complete solution

If you have 12 pools in your mana battery, all with unaugmented sparks and a single dominant pool with a slow 2 mana-per-tick drain), every tick, two random unaugmented sparks each transfer 1 mana to the dominant pool. In just a few ticks, particle streams are spawned from every nearby unaugmented spark to the dominant spark, giving the appearance of a lot of mana being transferred, but really only 2 mana per tick are moving.

The random selection of unaugmented sparks seems desirable and the choice to divide up the fill request for mana to all nearby unaugmented pools also seems desirable. I think the problem is related to how the visual weight of a particle stream isn't related to the amount of mana being transferred. 

One option might be to conditionally spawn particles based on a probability proportional to the amount of mana transferred. That is, if you're moving a lot of mana, we should almost certainly spawn a particle, but for small amounts, we should only rarely spawn one in. I can work on this if it seems like a good approach.

cjlarose:
Playing around with the probabilistic particle spawning approach for both outgoing and incoming transfers, things seem to looking better visually. When moving a small amount of mana, there are only a few occasional particles, and when moving a lot of mana, there are a bunch of particles.

```diff
diff --git a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
index d84e2b993..8422d97da 100644
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -210,7 +210,11 @@ public void tick() {
 					manaSpent += spend;
 					spark.checkReceiverFull();
 
-					particlesTowards(spark.entity());
+					// Spawn particles probabilistically based on transfer amount
+					double particleProbability = (double) spend / TRANSFER_RATE;
+					if (level().random.nextDouble() < particleProbability) {
+						particlesTowards(spark.entity());
+					}
 				}
 				receiver.receiveMana(-manaSpent);
 			}
@@ -241,7 +245,12 @@ public void tick() {
 						attachedReceiver.receiveMana(-gained);
 						manaRecieved += gained;
 
-						particlesFrom(spark.entity());
+						// Spawn particles probabilistically based on transfer amount
+						// Probability ranges from 0.0 (for small transfers) to 1.0 (for TRANSFER_RATE)
+						double particleProbability = (double) gained / TRANSFER_RATE;
+						if (level().random.nextDouble() < particleProbability) {
+							particlesFrom(spark.entity());
+						}
 					}
 				}
 				receiver.receiveMana(manaRecieved);
```

If that idea seems like a good approach, I can open a PR though I'll probably wait for https://github.com/VazkiiMods/Botania/pull/4965 to move through. That one I think should be valuable regardless of whether we decide to move forward with an approach that conditionally spawns particles based on the amount of transferred mana because it removes unnecessary calls to `attachedReceiver.receiveMana`

cjlarose:
Moving discussion about the visual busyness of low-throughput mana transfers to https://github.com/VazkiiMods/Botania/issues/4967

PR #4965, opened 2026-02-15
Skip zero-amount mana transfers between sparks

cjlarose:
When a dominant spark needs a small amount of mana (e.g., 2 mana) but has many inbound sparks available (e.g., 12), integer division in the fair share calculation causes most sparks to compute gained=0. This resulted in unnecessary receiveMana() calls and particle effects for transfers that moved no actual mana.

The proposed change performs transfers and spawns particles when gained > 0, reducing visual clutter and eliminating redundant operations.

Related

* https://github.com/VazkiiMods/Botania/pull/4963
* #4964

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -237,10 +237,12 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 					}
 
 					int gained = Math.min(attachedReceiver.getCurrentMana(), (manaNeeded - manaRecieved) / (count + 1));
-					attachedReceiver.receiveMana(-gained);
-					manaRecieved += gained;
+					if (gained > 0) {
+						attachedReceiver.receiveMana(-gained);
+						manaRecieved += gained;
 
-					particlesFrom(spark.entity());
+						particlesFrom(spark.entity());
+					}
 				}
 				receiver.receiveMana(manaRecieved);
 			}
--- a/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java
@@ -206,11 +206,14 @@ public class ManaSparkEntity extends SparkBaseEntity implements ManaSpark {
 					}
 
 					int spend = Math.min(attached.getAvailableSpaceForMana(), (manaTotal - manaSpent) / (count + 1));
-					attachedReceiver.receiveMana(spend);
-					manaSpent += spend;
-					spark.checkReceiverFull();
 
-					particlesTowards(spark.entity());
+					if (spend > 0) {
+						attachedReceiver.receiveMana(spend);
+						manaSpent += spend;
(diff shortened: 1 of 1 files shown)


cjlarose:
FWIW, the same issue doesn't appear to apply to outgoing transfers. In other words, `spend` here is never 0 in my testing

https://github.com/VazkiiMods/Botania/blob/b9d92b022cb90edf38cbea705b435fc9bd878260/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java#L208

TheRealWormbo:
Presumably that's because it's going to stop sending when it's out of mana, and if multiple source spawns are pushing towards the same target spark, they will be acting in tick order, one after another.
I'll have to try this one before I merge, just to make sure. Unfortunately I don't know when exactly that will be, but likely some time during the week.

NEstoll:
> FWIW, the same issue doesn't appear to apply to outgoing transfers. In other words, `spend` here is never 0 in my testing
> 
> https://github.com/VazkiiMods/Botania/blob/b9d92b022cb90edf38cbea705b435fc9bd878260/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java#L208

It does happen, you just need a source that generates a small amount of mana rapidly, of which there aren't many in base Botania. Letting a mana mirror accept mana from items and using an aura band shows it pretty well

cjlarose:
> > FWIW, the same issue doesn't appear to apply to outgoing transfers. In other words, `spend` here is never 0 in my testing
> > https://github.com/VazkiiMods/Botania/blob/b9d92b022cb90edf38cbea705b435fc9bd878260/Xplat/src/main/java/vazkii/botania/common/entity/ManaSparkEntity.java#L208
> 
> It does happen, you just need a source that generates a small amount of mana rapidly, of which there aren't many in base Botania. Letting a mana mirror accept mana from items and using an aura band shows it pretty well

Oh perfect. I'll see if I can reproduce this and if so, I'll make the change to the outgoing transfers as well

cjlarose:
I was able to successfully reproduce the same 0-mana transfer behavior by letting a Band of Aura add a little bit of mana to a Band of Mana and then dropping that Band of Mana into a recessive-sparked pool in range of some unaugmented-sparked pools (I wasn't able to figure out how to get the Band of Aura to push into a pool via a Mana Mirror).

I confirmed that 0-mana transfer happening with some [debug log messages](https://github.com/VazkiiMods/Botania/commit/789c36afbb67161510ccffac22fcba5013457df5), made the change to skip those 0-mana transfers, and then verified from my logs that the change successfully removed the 0-mana transfers.

TheRealWormbo:
I have to ask, because I only just noticed: Why are you committing changes this simple with the help of an LLM?

Vazkii:
Hello. 

After some chatter, we've reached the conclusion we are not interested in accepting LLM backed pull requests. There are several reasons behind this - chief among which being the additional load generated upon the current team (which is already operating on fumes) to review an increasing amount of PRs that may be prone to hallucinations or potential long-term complications from lack of understanding of project structure.

We appreciate the interest in providing us help, we absolutely do, but ultimately, we have determined that the downsides aren't worth it. Thank you for the initiative nontheless. With that in mind, we'll be closing this and any other open PRs authored by LLMs.

PR #4976, opened 2026-04-22
Remove overgrowth seed and enchanted soil

TheRealWormbo:
(as suggested by Artemis)


artemisSystem:
LGTM! i assume it builds and spotlesses. Also remember to update the changelog (i'd amend the commit i think)

TheRealWormbo:
So far, none of the 1.21 changes have been added to the changelog, mainly because of the way that causes a pain in the behind. Apparently building this branch will also update the website, so the changelog needs to be synchronized between the two active branches. I'm only sync'ing over 1.20 changes so far, but I wouldn't want to sync back 1.21 changes to the 1.20 branch as well.

artemisSystem:
Oh i see. That's a bit annoying. We should probably make it only push to the website from the "main" branch. But then this is fine to just merge now

PR #4979, opened 2026-04-29
1.21.1 gog porting

TheRealWormbo:
Set up datagen for GoG, including a new way to disable certain Botania recipes that GoG wants to modify.

GoG datagen happens in Neoforge, since some of the generated files are Neoforge-specific.


TheRealWormbo:
See https://github.com/TheRealWormbo/forged-conditions for the sources of that JAR file.

artemisSystem:
Why is the gog datagen under neoforge? I thought the point was to datagen it with fabric.

TheRealWormbo:
After looking into it, I found that only Botania actually needed the conditional loading. GoG has some Neoforge-specific data files (global loot modifiers) that would be highly annoying to generate either in Fabric or with a separate run that would have to be combined with a Fabric-based datagen run.

Basically it was easier to have the Botania versions of the affected recipes just turn off and put the GoG versions into the GoG namespace, since GoG being loaded would then be the reason they exist at all. Up until 1.20.1 we had to work around mod load order issues because we didn't just turn off the Botania versions, but attempted to replace them while keeping the same recipe IDs.

garden_of_glass/Makefile
```diff
@@ -1,4 +1,4 @@
-VERSION=1.13
+VERSION=21.1.0
```
artemisSystem: What's this version number based on? MC version + .0 for if we want to make changes? I see the appeal, but we don't necessarily need a new gog version per new minecraft version. What if we just make it a number that increments, like the botania patch version? maybe start at 14 or something (as a continuation of 1.13)

NeoForge/src/main/java/vazkii/botania/neoforge/data/gog/GogRecipeProvider.java
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	@Override
+	protected void buildRecipes(RecipeOutput recipeOutput) {
+		// modifications of standard Botania recipes:
+		nineBlockStorageRecipes(recipeOutput,
+				RecipeCategory.BREWING, Items.BLAZE_POWDER,
+				RecipeCategory.BUILDING_BLOCKS, BotaniaBlocks.blazeBlock,
+				"gardenofglass:blaze_block", null, "gardenofglass:blaze_powder_from_blaze_block", null);
+		ShapelessRecipeBuilder.shapeless(RecipeCategory.MISC, BotaniaItems.fertilizer)
```
artemisSystem: Isn't it supposed to give 3 in gog?

NeoForge/src/main/java/vazkii/botania/neoforge/data/gog/GogBlockLootSubProvider.java
```diff
@@ -0,0 +1,50 @@
...
+import net.minecraft.world.level.storage.loot.predicates.MatchTool;
+import net.neoforged.neoforge.common.Tags;
+
+import vazkii.botania.common.loot.BotaniaLootTables;
+
+import java.util.function.BiConsumer;
+
+public class GogBlockLootSubProvider implements LootTableSubProvider {
+	public GogBlockLootSubProvider(HolderLookup.Provider provider) {
+		// not necessary
```
artemisSystem: I'd rather leave this constructor completely empty than put a comment here

NeoForge/src/main/java/vazkii/botania/neoforge/data/gog/GogBlockUseLootProvider.java
```diff
@@ -0,0 +1,26 @@
...
+import net.minecraft.world.level.storage.loot.entries.LootItem;
+
+import vazkii.botania.common.item.BotaniaItems;
+import vazkii.botania.common.loot.BotaniaLootTables;
+
+import java.util.function.BiConsumer;
+
+public class GogBlockUseLootProvider implements LootTableSubProvider {
+	public GogBlockUseLootProvider(HolderLookup.Provider provider) {
+		// not necessary
```
artemisSystem: Same as other comment

Xplat/src/main/java/vazkii/botania/common/lib/BotaniaTags.java
```diff
@@ -341,6 +342,9 @@ public static class Blocks {
 		 */
 		public static final TagKey<Block> UNSUPPORTED_PLATFORM_DISGUISE = tag("unsupported_platform_disguise");
 
+		public static final TagKey<Block> GOG_PEBBLE_SOURCES =
```
artemisSystem, line 345: Skeptical of this, since it's only present in gog, right? Maybe this should live in the GoG class.

Xplat/src/main/java/vazkii/botania/common/loot/BotaniaLootTables.java
```diff
@@ -132,10 +133,17 @@ public class BotaniaLootTables {
...
+	public static final ResourceKey<LootTable> GOG_PEBBLES_TABLE = registerGog("pebbles");
+
 	private static ResourceKey<LootTable> register(String path) {
 		return register(ResourceKey.create(Registries.LOOT_TABLE, botaniaRL(path)));
 	}
 
+	private static ResourceKey<LootTable> registerGog(String path) {
+		return register(ResourceKey.create(Registries.LOOT_TABLE, gogRL(path)));
+	}
+
```
artemisSystem, line 146: same with these ones

Fabric/build.gradle
```diff
@@ -99,4 +99,8 @@ dependencies {
     modRuntimeOnly "maven.modrinth:here-be-no-dragons:1.0.0"
 
     compileOnly "org.jetbrains:annotations:26.0.2"
+
+    if (file('../garden_of_glass/gardenofglass-21.1.0.jar').exists()) {
```
artemisSystem: This version number should be in gradle.properties along with all the other version numbers, and interpolated in wherever it's used

artemisSystem (changes requested):
Nice work! I had some comments looking through. Can you amend the relevant commits to fix the comments? Then we can apply the commits directly without squashing, since they are nicely delimited.

Xplat/src/main/java/vazkii/botania/common/lib/BotaniaTags.java
TheRealWormbo: It's used in the `SkyblockWorldEvents` class in Xplat. It made sense for me to not have random resource/tag keys defined elsewhere than where Botania defines most of the others.

Xplat/src/main/java/vazkii/botania/common/loot/BotaniaLootTables.java
TheRealWormbo: Both loot tables need to be known in multiple places. `extra_seeds` is used in the Fabric loot table injection logic, `pebbles` is used in Xplat `SkyblockWorldEvents`, and both are used in GoG datagen in Neoforge.

Xplat/src/main/java/vazkii/botania/common/lib/BotaniaTags.java
artemisSystem: Hm, fair enough. I thought there was a main GoG entrypoint class where it would make sense to put this, but gog doesnt have any code, so i must have imagined it. this is fine then

TheRealWormbo:
> Can you amend the relevant commits to fix the comments? Then we can apply the commits directly without squashing, since they are nicely delimited.

Done. I had it locally rebased onto the same commit as the overgrowth PR already, so I had to push all commits again, but the changes really are just the version numbers with central definition in Gradle, the fertilizer recipe output count, and the removed constructor comments.

artemisSystem:
Looks good, thanks! Github of course also includes irrelevant changes when comparing the force pushes, but the changes that look relevant look good. I'll merge them in when the CI is done.

PR #4983, opened 2026-04-30
Allow ManasteelArmorItem into Mana Enchanter fix

Drakkens:
Ever since the addition of the new DataComponents, the Item#isEnchantable method now checks by default for a MAX_DAMAGE, which our ManasteelArmorItems do not have, making them unable to be put into the ManaEnchanter.

I've just overridden the isEnchantable to allow it.

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/item/BotaniaItems.java
+++ b/Xplat/src/main/java/vazkii/botania/common/item/BotaniaItems.java
@@ -202,30 +202,30 @@ public final class BotaniaItems {
 	public static final Item missileRod = make(LibItemNames.MISSILE_ROD, new UnstableReservoirRodItem(unstackable().rarity(Rarity.RARE)));
 
 	// Equipment
-	public static final Item manasteelHelm = make(LibItemNames.MANASTEEL_HELM, new ManasteelHelmItem(unstackableCustomDamage()));
-	public static final Item manasteelChest = make(LibItemNames.MANASTEEL_CHEST, new ManasteelArmorItem(ArmorItem.Type.CHESTPLATE, unstackableCustomDamage()));
-	public static final Item manasteelLegs = make(LibItemNames.MANASTEEL_LEGS, new ManasteelArmorItem(ArmorItem.Type.LEGGINGS, unstackableCustomDamage()));
-	public static final Item manasteelBoots = make(LibItemNames.MANASTEEL_BOOTS, new ManasteelArmorItem(ArmorItem.Type.BOOTS, unstackableCustomDamage()));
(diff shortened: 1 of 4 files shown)


TheRealWormbo:
That seems like it's treating the symptoms of a different problem. As you pointed out, manasteel armor items don't have that component, which is probably wrong, since they should be able to take damage when the player has no mana.

Drakkens:
I guess I haven't tested that either, good point. Will look into it and get back to you.

Drakkens:
From what I've seen, durability was removed from the ArmorMaterial definition, and it was never added to the 1.21.1 version of the amours. 

I've just added it manually via constructors, which is what seems to be done in vanilla armors too.

On the topic of armors, I noticed Elementium was missing its armor value, so I fixed it and moved pixieChance to an override on #getDefaultAttributeModifiers (same approach as used for Knockback Resistance on Terrasteel).

Should we use botaniaRL() when creating the Modifier?

TheRealWormbo:
> From what I've seen, durability was removed from the ArmorMaterial definition, and it was never added to the 1.21.1 version of the amours.
> 
> I've just added it manually via constructors, which is what seems to be done in vanilla armors too.

Vanilla would be a good reference point, yes. Vanilla appears to define its item durabilities in the `Items` class definitions, no the individual constructors though. Maybe we should do that as well.

> On the topic of armors, I noticed Elementium was missing its armor value, so I fixed it and moved pixieChance to an override on #getDefaultAttributeModifiers (same approach as used for Knockback Resistance on Terrasteel).

I'm not sure if the terrasteel armor can be trusted as example, considering its comments. How is it done in vanilla, e.g. for netherite knockback resistance?

> Should we use botaniaRL() when creating the Modifier?

If it involves a namespaced ID for a Botania thing, `botaniaRL()` is usually the correct choice. Not sure why that one uses a default namespace ID instead. If it works with a Botania namespace, it should be changed.

Drakkens:
> I'm not sure if the terrasteel armor can be trusted as example, considering its comments. How is it done in vanilla, e.g. for netherite knockback resistance?

Fair. As far as I can see, Netherite has `knockbackResistance` which comes baked in from the `ArmorMaterial`, which takes a uniform value for all the pieces (in Netherite's case, a flat 10% per piece). Since Terrasteel's is defined per piece, we can not use this approach directly.

Back to Elementium, I fail to see/find a way to implement it differently. The main issue is, whenever the `Item.Properties#attributes()` is called, it overwrites the already existing `DataComponent.ATTRIBUTE_MODIFIERS` (same is true when `Item.Properties#component()` is called), and I am not able to find any other way to append to it instead.

> If it involves a namespaced ID for a Botania thing, `botaniaRL()` is usually the correct choice. Not sure why that one uses a default namespace ID instead. If it works with a Botania namespace, it should be changed.

Seems to work fine with `botaniaRL()`

TheRealWormbo:
Oh, okay, I see it now. Your implementation seems good then.

PR #4985, opened 2026-05-02
Adjust nether-gating for rune crafting

TheRealWormbo:
This PR makes adjustments to several rune recipes and the terra plate crafting recipe to make the distinction between nether-gated and pre-nether content more defined.

Previously the Rune of Fire was nether-gated by requiring nether wart, which in turn nether-gated the Terrestrial Agglomeration Plate (including all of the Alfheim content), half of the "season runes", most of the "sin runes", and various pre-Alfheim flowers and tools.

With this change, the nether-gating is more clearly defined: As far as involved runes are concerned, only terrasteel (and thus Alfheim) and "sin runes" (and anything crafted from them) require having been to the nether. This new gating also is somewhat less strict, since quartz can be found anywhere, while nether wart requires finding a nether fortress or certain types of bastion remnants. (Botania doesn't help with exploration for those structures.)


TheRealWormbo:
Changes in rune availability:
- Fire, Spring, and Autumn are now pre-nether
- Envy, Lust, and Wrath are now post-nether

Recipes that are post-nether now:
- Damage Lens (Wrath)
- Entropic Lens (Envy)

Recipes that are pre-nether now:
- Agricarnation (Spring)
- Exoflame (Fire)
- Fallen Kanade (Spring)
- Gourmaryllis (Fire)
- Hyacidus (Autumn)
- Jaded Amaranthus (Spring)
- Marimorphosis (Fire)
- Thermalily (Fire)
- Tigerseye (Autumn)
- cobble rod (Fire; not bothering to look up the proper names for these)
- cloud pendant (Autumn)
- fire rod (Fire)
- goddess charm (Spring)
- knockback belt (Fire)
- lava pendant (Fire)
- power lens (Fire)

Any other recipes involving those runes are either further gated (usually via terrasteel, Alfheim materials, or the previously-gated sin runes), or switched their nether-gating from a fire/spring/autumn rune to a previously ungated sin rune.

PR #4988, opened 2026-05-09
Update zh_cn.json and zh_tw.json

Dawnwalker666:
The scope of the translation includes all commits about en_us.json from Jul 22, 2025 to May 09, 2026.
zh_tw.json is from March 30, 2025 to May 09, 2026.


TheRealWormbo:
Will there be an update to zh_tw.json as well or can I merge this change already?

Dawnwalker666:
All right, since that's the case, I'll find some time today to update TW.

Dawnwalker666:
Done

PR #4990, opened 2026-05-09
1.21.1 mana infusion rebalance, 2nd attempt

TheRealWormbo:
The primary goal of this PR is to clean up mana infusion datagen and to close some gaps in Botania functionality regarding newer Minecraft versions.

Overview of changes:
- Mana infusion recipe IDs are now mostly based on the catalyst type, group name, and result item
  - Alchemy recipes moved to the `mana_infusion/alchemy/` path prefix
    - Alchemy cycles further moved to `mana_infusion/alchemy/<cycle name>/` path prefix, with the recipe names changed to match the standard `<output>_from_<input>` format vanilla uses in various places
    - Deconstruction recipe names changed to the less-standard (but sufficiently unique) format `<input>_deconstruct`
    - All other alchemy recipe names match the output item name
  - Conjuration recipes moved to the `mana_infusion/conjuration/` path prefix and changed name to just the duplicated item
  - Non-catalyst recipes stay in the `mana_infusion/` path prefix and generally are named after the output item, with mana powder being the exception
  - Mana Powder recipes have the name format `mana_infusion/mana_powder_from_<source name>`, with the two source names being `dye` and `dust`
- Deconstruction recipes were adjusted to better match the general block availability in current Minecraft versions
  - Whether a deconstruction is lossy or lossless now depends on whether there is some vanilla way to potentially deconstruct the block (e.g. glowstone blocks can return up to 4 glowstone dust items when mined, while there is no way to deconstruct a quartz block in vanilla)
    - Lossless deconstruction is considered a convenience and is generally cheaper
    - Lossy deconstruction is considered partial material recovery and is generally more expensive
    - Mana cost is also based on the number of items returned, following a square-root function with a base mana factor depending on the type of deconstruction (tuned to be the same as the existing wool deconstruction cost for lossy yielding 3 items, and almost the same as the previous cost for losslessly yielding 4 items)
  - New lossy deconstructions (recovered items):
    - amethyst block (3x amethyst shard)
    - bamboo block (8x bamboo)
    - dripstone block (3x pointed dripstone)
    - honeycomb block (3x honeycomb)
    - magma block (3x magma cream)
    - nether bricks (3x nether brick)
    - red nether bricks (2x nether brick)
  - New lossless deconstructions:
    - melon block (9x melon slice)
  - Deconstructions that are now lossy:
    - all quartz block variants (3x corresponding quartz variant item)
    - bricks (3x brick)
- Cycle changes:
  - New cycle: nether stems (crimson/warped stem) – priced like overworld logs cycle)
  - New cycle: nether fungi (crimson/warped fungus) – priced at 1% of the very expensive sapling cycle
  - Changed cycle: stones – now also contains tuff, which is converted from granite and converts into andesite
  - Removed cycle: 1.17 stones (deepslate, tuff, calcite) – calcite is a Pure Daisy recipe now, converting from dripstone, which is vanilla-farmable; deepslate is already farmable via end stone purification
- Other changes:
  - Mana Powder recipes now use tags for the input items (`c:dyes` and `botania:mana_powder_source_dusts`, respectively)
  - Chiseled Stone Bricks recipe is gone


Fabric/src/main/java/vazkii/botania/fabric/data/xplat/ManaInfusionProvider.java
```diff
@@ -51,189 +59,219 @@ public String getName() {
...
-		cycle(consumer, 40, "botania:log_cycle",
+		normal(consumer, BotaniaItems.manaSteel, ConventionalItemTags.IRON_INGOTS, COST_MANASTEEL);
+		normal(consumer, BotaniaBlocks.manasteelBlock, ConventionalItemTags.STORAGE_BLOCKS_IRON,
+				BLOCK_COST_MULTIPLIER * COST_MANASTEEL);
+		normal(consumer, BotaniaItems.manaPearl, Items.ENDER_PEARL, 6000);
+		normal(consumer, BotaniaItems.manaDiamond, ConventionalItemTags.DIAMOND_GEMS, COST_MANA_DIAMOND);
+		normal(consumer, BotaniaBlocks.manaDiamondBlock, ConventionalItemTags.STORAGE_BLOCKS_DIAMOND,
+				BLOCK_COST_MULTIPLIER * COST_MANA_DIAMOND);
+
+		normal(consumer, id("mana_powder_from_dust"), "botania:mana_powder", BotaniaItems.manaPowder,
```
artemisSystem: why use the "botania:mana_powder" ID here instead of the item reference?

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -2983,51 +2984,46 @@
...
-  "botania.page.manaAlchemy5": "Breaking down $(item)Glowstone Blocks$(0)",
-  "botania.page.manaAlchemy31": "Quartz",
-  "botania.page.manaAlchemy6": "Breaking down $(item)Quartz Blocks$(0)",
-  "botania.page.manaAlchemy7": "Creating $(item)Chiseled Stone Bricks$(0)",
-  "botania.page.manaAlchemy8": "Compressing $(item)Snow$(0) into $(item)Ice$(0)",
-  "botania.page.manaAlchemy32": "Vines & Lily Pad",
-  "botania.page.manaAlchemy9": "Converting between $(item)Vines$(0) and $(item)Lilypads$(0)",
-  "botania.page.manaAlchemy33": "Fish",
-  "botania.page.manaAlchemy10": "Transmuting breeds of $(item)Fish$(0)",
+  "botania.page.manaAlchemyStems.title": "Nether Logs",
```
artemisSystem: I'd rather call these "Nether Stems", or maybe "Wooden Stems"

artemisSystem:
Looks good apart from a few small inline comments

Fabric/src/main/java/vazkii/botania/fabric/data/xplat/ManaInfusionProvider.java
TheRealWormbo: Using the item reference now.

artemisSystem: Oh, i didn't see that that was for the group, i thought you were specifying the item ID as a string (i didn't ask you to remove it, i asked why :p). But this works too.

PR #4991, opened 2026-05-09
Fixed casting to double from rnd in AlfheimPortalBlockEntity blockParticle

AidaNightcore:
Fixed/added a double type casting to double dh = (double) rnd/3 - 1 on line 245 in blockParticle, as rnd is int on its own and the division affects the arithmetic math in some cases. Same change, if accepted, should be put in the 1.21.1 porting too.
This small change is for a uni project of finding bugs of my teacher's liking, not to be nitpicky.


artemisSystem:
Why does this PR also move around two other methods, and add a line break in that record definition?

AidaNightcore:
I think I accidentally pressed save which reformats the code before adding in git, nothing malicious, just desperate for extra project points.  I can try doing another PR?

artemisSystem:
Also, this fix doesn't make sense, sorry. It seems we specifically want the int here (-1, 0 or 1). Div and mod are used separately to get an x and y position in a 3x3 square from a 0-8 int.

AidaNightcore:
Thank you for the explanation!

artemisSystem:
No problem, good luck with your project!

PR #4994, opened 2026-05-13
Gaia fight update

TheRealWormbo:
Adjustments and upgrades to various aspects of the gaia fight:

- visuals for the guardian, its pixies, and the arena
- details of the fight choreography
- internal improvements


TheRealWormbo:
Changes so far:
- gaia guardian invulnerability distortions are no longer disabled by the presence of gaia pixies
- gaia guardian teleports to the beacon when the mob spawning phase begins
- pixies (both gaia and elementium variants) no longer get incredibly slow when they are further away from their target
- tweaks to the gaia distortion shader: it now applies luminance-based grayscale conversion and also fades over to the static noise instead of just adding that to the grayscale texture
- pixie type is synchronized as part of the spawn packet, not as a separately synchronized value
- gaia guardian now also renders any applicable outer layers of the player model, according to which parts the player wants to show (No capes!)
- gaia head no longer crashes when placed and renders (mostly) correctly (the wobble still is a bit hard to see for inventory rendering)

Known issues:
- gaia boss bar doesn't survive a reload

TheRealWormbo:
Gaia fight logic is now more data-driven through JSON configs that are created via datagen.
The actual fight itself should not really have changed, as the initial datagen for the "normal" and "hard" configs replicate the previously hard-coded logic.

TheRealWormbo:
I'll call this one done for now. Relevant technical changes are in, which is a good basis for possible gameplay changes.

Summary of gameplay-relevant changes:
- fixed gaia head placement crash
- removed in-fighting among gaia-spawned mobs
- restricted mobs from targeting things outside the arena area
- the area for pulling in players is a sphere instead of a cube now
- pixie speed is independent of distance to target

Visual changes:
- boss bar works after reload
- boss bar shader is actually applied now
- improved boss bar static noise shader
- decoupled static noise shader parameters used for the gaia boss from those used for the gaia head and gaia pixies (still same parameter names, just two instances)
- switched static noise shader grayscale conversion from average to luminosity method, and somewhat decreased the amount of darkening applied
- gaia boss also renders the player's outer texture layers, matching the player's choice of which layers are enabled (except cape – No capes!)

Internal changes:
- many of the gaia fight parameters are data-driven now, including datagen
- moved Loonium structure configurations to `config/loonium` folder, next to the two gaia fight configurations under `config/gaia`
- Loonium now supports the new `count` parameter in mob spawning configurations, which is an `IntProvider` (this is not yet used in any Loonium structure configuration), and mob spawn configurations can also allow equipment drops
- moved gaia fight network initialization to synchronized gaia guardian entity values rather than a special packet type
- moved pixie type synchronization from a synchronized value to the initial packet data value
- added a way to use `GuiGraphics.blit` without having it set a default shader (cast the `GuiGraphics` instance to `GuiGraphicsAddon` to call `botania_blitWithoutShader`)
- renamed a bunch of gaia-related things from "doppleganger" (except the entity ID itself and things that rely on it) to what they should actually be called

TheRealWormbo:
Rebased onto recent commits in `1.21.1-porting` due to some merge conflicts and combined a couple of commits. No relevant changes beyond that.

artemisSystem:
I'm not super knowledgeable on network code yet, but had a look at that too and everything seems ok. Nice work!

PR #5001, opened 2026-05-22
1.21.1 Rosa Arcana rework

TheRealWormbo:
This updates the mechanics and visuals of the Rosa Arcana (and related Hopperhock visuals) as follows:

- instead of trying to first draining players, then consuming XP orbs, and finally disenchanting items, the flower now primarily consumes XP orbs, then attempts to disenchant items, and only if neither of those work, it extracts XP orbs from nearby players (instead of draining them directly)
- players cannot pick up XP orbs while they are near the flower
- the flower attracts XP orbs, similar to players, and the orbs even prefer moving towards a Rosa Arcana over moving towards a player
- for visuals, the Rosa Arcana and Hopperhock now use a vanilla-like pickup animation for XP orbs and items, respectively

### DIFF
--- a/Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
+++ b/Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
@@ -13,6 +13,7 @@ import it.unimi.dsi.fastutil.objects.Object2IntMap;
 import net.minecraft.core.BlockPos;
 import net.minecraft.core.Holder;
 import net.minecraft.core.component.DataComponents;
+import net.minecraft.server.level.ServerLevel;
 import net.minecraft.sounds.SoundEvents;
 import net.minecraft.sounds.SoundSource;
 import net.minecraft.tags.EnchantmentTags;
@@ -36,11 +37,13 @@ import vazkii.botania.common.helper.EntityHelper;
 import vazkii.botania.common.helper.MathHelper;
 import vazkii.botania.mixin.ExperienceOrbAccessor;
 
+import java.util.ArrayList;
 import java.util.List;
 
 public class RosaArcanaBlockEntity extends GeneratingFlowerBlockEntity {
 	private static final int MANA_PER_XP = 50;
 	private static final int RANGE = 1;
+	public static final int PLAYER_TAKE_XP_DELAY = 10;
 
 	public RosaArcanaBlockEntity(BlockPos pos, BlockState state) {
 		super(BotaniaBlockEntities.ROSA_ARCANA, pos, state);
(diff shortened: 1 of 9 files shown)


Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
```diff
@@ -50,28 +53,48 @@ public RosaArcanaBlockEntity(BlockPos pos, BlockState state) {
...
+		}
+		if (disenchantAnItem(serverLevel, effectBounds)) {
+			return;
+		}
+		drainPlayerXp(serverLevel, effectBounds);
+	}
+
+	private static void drainPlayerXp(ServerLevel serverLevel, AABB effectBounds) {
+		List<Player> players = new ArrayList<>();
+		serverLevel.players().stream()
```
artemisSystem, line 75: Why not keep the getEntitiesOfClass with the bounds? That seems faster than looping over every existing player (unless getEntitiesOfClass has to do that anyway?)

Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
```diff
@@ -50,28 +53,48 @@ public RosaArcanaBlockEntity(BlockPos pos, BlockState state) {
...
+		// possibly equal to zero even when the level is > 0.
+		// Instead, check the level and intra-level progress separately.
+		players.removeIf(player -> player.experienceProgress == 0 && player.experienceLevel == 0);
+		if (players.isEmpty()) {
 			return;
 		}
+		// pull XP from one random player
+		Player player = players.get(serverLevel.random.nextInt(players.size()));
+		// this effectively only happens every other tick, so average out the XP drain to 1 per tick
+		int drainAmount = player.experienceLevel > 0 && serverLevel.random.nextBoolean() ? 3 : 1;
```
artemisSystem, line 93: 1. Won't this (`player.experienceLevel > 0`) make it not drain the player's last level?
2. What's meant by it only happening every other tick?

artemisSystem:
Got some questions in the inline comments ^

Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
TheRealWormbo: Entity search is a bit of a weird one. Minecraft generally first assumes all entities are the size of ghasts when selecting the subchunks to search, then iterates over all entities in those subchunks, filtering out any that don't match the class, actually overlap with the specified area, or fail any additional predicate. In the best case it's only one subchunk, but it could be up to eight, and with Botania there are probably a bunch of item frames and/or sparks around, at the very least.
The server-wide player list is just that – a list of all player entities. This might include fake players, but I'm not sure if it would cause a relevant performance impact.

artemisSystem: Gotcha, that makes sense. Probably better looping over the players (your new change) in most cases then. Mind leaving a comment explaining this, so we don't forget?

Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
TheRealWormbo: The way this works is:
- When there's an orb, the flower won't drain the player, but instead eat the orb.
- When there's no orb, it drains the player. If the player has at least one level, it randomly makes orbs of sizes 1 or 3, i.e. average size 2. The orb will be eaten next tick, so it's one orb, averaging 2 XP, every other tick, averaging 1 XP per tick, as before.
TBH, I just didn't want to have to calculate the actual XP amount left if the player has no complete levels. That's why the final few XP points are effectively drained at a rate of 1 point every other tick.

artemisSystem: Oh i see, it will still clear it out, just slower on the last level. That's fine, it's almost like it's harder to pull out the last level.

Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/generating/RosaArcanaBlockEntity.java
TheRealWormbo: Done.

PR #5004, opened 2026-05-25
Update animated torch visuals to match redstone torch

TheRealWormbo:
- looks like an actual floating redstone torch now
- visually turns off while rotating (torch off texture, no redstone particles, light level zero)
- bonus "magic floaty" particles
- now uses wood sounds for placing and breaking
- no more entity collision (still counts as spawn-proof due to being defined as a redstone signal source)


TheRealWormbo:
Screenshots for reference:

PR #5012, opened 2026-06-08
Fixed Integer overflow for TerraShatterer

DefaultEugene:
Hi there!
I found an issue with the TerraShatterer. Since its mana cap is `Integer.MAX_VALUE`, it is the only item vulnerable to the Integer Overflow problem

If the pickaxe has `(MAX_MANA - 5)` mana and we try to add 10 mana, the mana value will wrap around and reset to 0. I've fixed it using a safe comparison before adding

I also want to draw attention to the `getMana` method: it returns `stack mana * stack count`, which can also cause an overflow. However, that bug is not as critical as this one (since it requires a `stack size > 1` for an unstackable item) and would require a much more complex fix using long arithmetics


TheRealWormbo:
Did you actually find an overflow scenario? The only way mana is added to the Terra Shatterer is via a mana pool, and that should be limiting the amount of mana to exactly reach the maximum.

[edit] (Though I agree the limitation implemented in `addMana` is ineffective.)

PR #5036, opened 2026-08-01
Fix sampler definition in film_grain_particle.json

ErythroCraft:
Deleting this line is fixing the following issue.

Shader botania:film_grain_particle could not find sampler named Sampler2 in the specified shader program.

Including Sampler2 in this particle shader causes rendering errors and log spam (e.g., “Could not find sampler named Sampler2”), especially when users run performance mods or shaderpacks like Iris, Oculus, or Complementary.

### DIFF
--- a/Xplat/src/main/resources/assets/botania/shaders/core/film_grain_particle.json
+++ b/Xplat/src/main/resources/assets/botania/shaders/core/film_grain_particle.json
@@ -14,8 +14,7 @@
     "UV2"
   ],
   "samplers": [
-    { "name": "Sampler0" },
-    { "name": "Sampler2" }
+    { "name": "Sampler0" }
   ],
   "uniforms": [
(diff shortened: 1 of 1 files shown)


TheRealWormbo:
Seems reasonable.

What about the `Light0_Direction` and `Light1_Direction` uniforms the `botania:pylon` shader appears to have trouble with. Do they also cause log spam like that?

ErythroCraft:
That would require modifying the pylon.json file.
It's a different fix. I'll do it now.

ErythroCraft:
> Seems reasonable.
> 
> What about the `Light0_Direction` and `Light1_Direction` uniforms the `botania:pylon` shader appears to have trouble with. Do they also cause log spam like that?

https://github.com/VazkiiMods/Botania/pull/5037

TheRealWormbo:
Thanks. I squashed your two PRs into one commit, since they are related.

PR #5040, opened 2026-08-16
Replace Ender Air with "Ender Essence" (End-gating rework)

TheRealWormbo:
It's still technically the same thing, but handled somewhat differently:
- cannot be collected from the air in the End
- is collected with Alfglass Flasks instead of mundane glass bottles
- exists in "diluted" and "pure" forms that can be used interchangeable for some recipes, while others only work with the pure form
- only the pure form can be thrown for all the usual effects (irritating ghasts' eyes, converting stone to end stone)

There are two new ways to create ender essence clouds, from which the essence can be bottled:
- stabbing an Enderman with a Soulscribe
- capturing a teleporting Enderman with a Viculotus (this causes the Enderman to lose its teleport ability)

In either case a cloud of Ender Essence is spawned, but it will be diluted if this happens outside the End.

Along with this change, the recipes of various Corporea-related items have changed
- most prior uses of Ender Air now require Pure Ender Essence, which is only obtainable after going to the End
- Corporea Crystal Cube: now crafted from three Alfglass blocks and an Elementium ingot surrounding a Corporea Block
- Corporea Funnel: recipe layout like a dispenser, except Corporea Blocks instead of cobblestone and any quality Ender Essence instead of the bow
- Corporea Index: swapped out the Corporea Spark for a Corporea Block
- Corporea Interceptor: a row of Corporea Blocks at the top and bottom, middle row is Redstone Dust on the outsides and Pure Ender Essence in the center
- Corporea Retainer: Corporea Blocks in a "cauldron/composter shape", Redstone Torch at the top center, Pure Ender Essence in the middle
- Corporea Spark: any quality of Ender Essence can be used
- Extrapolated Bucket: any quality of Ender Essence is now required at the center of the "bucket shape" of Elementium ingots (comparable to its use in the Black Hole Talisman)
- Spectranthemum: swapped out the Pixie Dust for any quality of Ender Essence (better fit for the teleportation flavor)
- Life Imbuer: any quality of Ender Essence can be used in the crafting recipe (spawners aren't that great of a way to spawn mobs that their "playerless" use would need to be locked behind End access)

Effectively these changes make some Corporea features (sparks, decorative blocks, funnel, and crystal cube) and the Life Imbuer available before access to the end, but still after opening the Alfheim portal. The changes to the Extrapolated Bucket and Spectranthemum recipes don't represent a gating change, since they work with diluted essence. Farming significant amounts of Ender Essence has become a greater automation challenge, as simply spamming empty bottles in the End is no longer an option.


Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/functional/VinculotusBlockEntity.java
```diff
@@ -93,10 +102,44 @@ public static Vec3 onEndermanTeleport(EnderMan entity, double targetX, double ta
...
 			flower.addMana(-cost);
 
+			// mark it for allowing this teleportation attempt and capturing right after that
+			EnderEssenceCaptured.HOLDER.setFor(entity, false);
 			// Endermen are 0.6 blocks wide, so +/- 1.2 blocks offset should always fit into the 3x3 block target area
 			return new Vec3(x + (Math.random() * 2.4 - 1.2), y, z + (Math.random() * 2.4 - 1.2));
 		}
 
+		// remove any potential marker that may have been left over from a previous attempt to capture this EnderMan
+		EnderEssenceCaptured.HOLDER.removeFrom(entity);
```
artemisSystem, line 111: In what cases does this happen? Is the check necessary?

artemisSystem: i guess if the position the vinc finds is obstructed and the enderman won't teleport? Is there a way to instead only place the marker after a successful teleport?

Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/functional/VinculotusBlockEntity.java
```diff
@@ -121,11 +120,7 @@ public static Vec3 onEndermanTeleport(EnderMan entity, double targetX, double ta
 	 */
 	public static void postEndermanTeleport(LivingEntity entity) {
 		if (EnderEssenceCaptured.HOLDER.existsFor(entity)) {
-			EnderAirCloudEntity cloud = BotaniaEntities.ENDER_AIR_CLOUD.create(entity.level());
-			if (cloud != null) {
-				cloud.moveTo(entity.position(), entity.getYRot(), 0);
-				entity.level().addFreshEntity(cloud);
-			}
+			EnderAirCloudEntity.spawnForEnderman(entity);
```
artemisSystem: EnderAirCloudEntity -> EnderEssenceCloudEntity?

Xplat/src/main/java/vazkii/botania/common/entity/EnderAirCloudEntity.java
```diff
@@ -11,35 +11,105 @@
...
+		return new EnderAirCloudEntity(entityType, level, true);
+	}
+
+	public EnderAirCloudEntity(EntityType<?> entityType, Level level, boolean diluted) {
 		super(entityType, level);
+		this.diluted = diluted;
+	}
+
+	public static void spawnForEnderman(LivingEntity entity) {
+		EntityType<EnderAirCloudEntity> type = entity.level().dimension() == Level.END
```
artemisSystem: If we want to be really nerdy about it, maybe we could add a dimension property like vanilla has for water evaporating and lava flowing fast and such in the nether. It might be difficult in this version (inject into end.json anyone?) but later i think this is part of a way more extensible system, so it can be integrated into that down the line

Xplat/src/main/resources/assets/botania/lang/en_us.json
```diff
@@ -3574,11 +3574,14 @@
...
+  "botania.entry.enderEssence": "Ender Essence",
+  "botania.tagline.enderEssence": "Throwable flasks that create End Stone",
+  "botania.page.enderEssenceExtraction": "Some substances present in the air of the $(thing)End$(0) void seem to have mutative properties, but are usually too diluted to be harvested directly.$(p)Appropriate tools are required to extract this $(item)Ender Essence$(0) from certain creatures or even the rocks found there.",
+  "botania.page.enderEssenceFromEndermen": "One way to extract ender essence is by literally carving it out of $(thing)Endermen$(0) using a $(l:ender/soulscribe)Soulscribe$(/l). It's a rather crude method and, when used outside the End, will yield the essence only in a diluted form.$(p)However, even such $(item)Diluted Ender Essence$(0) may be sufficient for certain applications.",
```
artemisSystem, line 3581: It might be worth to mention the vinculotus here as an option, since one might have not read that entry, or forgotten the reference to ender essence

artemisSystem: Maybe just a link to the page is enough

artemisSystem:
Nice! This looks very good. See some small comments inline, otherwise good to merge

Xplat/src/main/java/vazkii/botania/common/entity/EnderAirCloudEntity.java
TheRealWormbo: Oh yes, this definitely sounds like a boolean environment attribute, or whatever those things in recent MC version are called. I also thought about making it a biome tag in the meantime, but we'll have to investigate that during the next version port.

Xplat/src/main/java/vazkii/botania/common/block/block_entity/flower/functional/VinculotusBlockEntity.java
TheRealWormbo: We still need to know if this was a redirected teleportation. The `false` marker is intended to only exist during teleportation, but there I didn't see any clean way to remove it immediately. The Vinculotus logic checks all teleportation attempts anyway, so I opted for cleaning up the flag the next time the enderman attempts to teleport and is not affected by the flower.

Xplat/src/main/resources/assets/botania/lang/en_us.json
TheRealWormbo: It's a bit of a tricky sequencing, as the Vinculotus page is only unlocked when you get the Runic Altar, while the Ender Essence page already unlocks as soon as you get a mana pool. (Originally it was unlocked from the start.)
I'll need to look into how linking to locked pages works.

TheRealWormbo: Maybe we can rely on referencing the Vinculotus in the Alfglass Flask entry, and in recipe viewers. That way the entry will definitely have been unlocked (or unlocks don't matter).

TheRealWormbo: I've added a mention of the Vinculotus effect in the Alfglass Flask entry. Also, EMI world interaction recipes now hint at the three possible options for obtaining Ender Essence. (And we continue to neglect JEI and REI when it comes to that kind of information.)

PR #5043, opened 2026-08-24
Update ko_kr.json (apply changes from 20.1 ko and 21.1 en)

2hh8899:
Apply changes between 20.1 and 21.1, and #5042.


TheRealWormbo:
Thank you for contributing translations to Botania.

You may have noticed the build for your PR has failed. That's because for the 1.21.1 branch we enforce a rule mentioned at the top of translation files: If a translation string is the same as in the English version, it should not appear in the translation file.

If you click on the failed checks information, you can access the build information. There, under "Check generated files up to date" it shows a log of our language file check at the end. Look for lines for the ko_kr.json file that state "Dropping copied value for '(key name)'", those are the values that were just copied from the English version.

2hh8899:
@TheRealWormbo Thank you for explaining the issue to me. Those copied values didn't need to be translated (they were things like record descriptions or terra shatterer ranks), so the issue was easily fixed by removing them.

PR #5048, opened 2026-08-30
Upload changelogs and .asc signatures, where appropriate

### LINKED ISSUE #5046: Upload .asc signatures to Modrinth
Apparently this is possible now so try for the next release? https://github.com/modrinth/code/issues/2643

And also for Patchouli

TheRealWormbo:
- changelog text is extracted from web/changelog.md, reading anything between the line containing `version="<version number from tag>"` and the next line containing `---`
- ASCII armored signatures (.asc files) are also uploaded to Modrinth (in addition to GitHub; implements #5046)

### DIFF
--- a/scripts/upload_releases.sh
+++ b/scripts/upload_releases.sh
@@ -9,18 +9,16 @@ VERSION="${TAGNAME/#release-}"
 # also remove '_<anything>' from end (for potential retries)
 VERSION="${VERSION/%_*}"
 MC_VERSION=$(echo "${VERSION}" | cut -d '-' -f 1)
-CHANGELOG_FRAGMENT=$(echo "${VERSION}" | tr . -)
-CHANGELOG_LINK="https://botaniamod.net/changelog.html#${CHANGELOG_FRAGMENT}"
+CHANGELOG_TEXT=$(sed -n "/version=\"${VERSION}\"/,/---/"'{//!p;}' web/changelog.md)
+CHANGELOG_MARKDOWN=$(cat <<EOF
+Changes in this version:
+${CHANGELOG_TEXT}
+EOF
+)
 
 function release_github() {
 	echo >&2 'Creating GitHub Release'
-	gh api \
-	   --method POST \
-	   -H "Accept: application/vnd.github+json" \
-	   -H "X-GitHub-Api-Version: 2022-11-28" \
-	   /repos/VazkiiMods/Botania/releases \
-	   -f tag_name="${TAGNAME}"
-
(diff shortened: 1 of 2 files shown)


scripts/upload_releases.sh
```diff
@@ -10,17 +10,16 @@ VERSION="${TAGNAME/#release-}"
 VERSION="${VERSION/%_*}"
 MC_VERSION=$(echo "${VERSION}" | cut -d '-' -f 1)
 CHANGELOG_FRAGMENT=$(echo "${VERSION}" | tr . -)
```
artemisSystem: This looks unused now

scripts/upload_releases.sh
```diff
@@ -10,17 +10,16 @@ VERSION="${TAGNAME/#release-}"
...
-CHANGELOG_LINK="https://botaniamod.net/changelog.html#${CHANGELOG_FRAGMENT}"
+CHANGELOG_TEXT=$(sed -n "/version=\"${VERSION}\"/,/---/{//!p;}" web/changelog.md)
```
artemisSystem: i can't get this command to work locally with bash, it says `!p: event not found`. Does it work for you? I'm on 3.2.57

TheRealWormbo: That worked fine for me on Linux Mint (which is based on Ubuntu), but maybe it's the flavor of the `sed` command that comes with it. I'll have to check if there's a more generic way to define this command sequence.

scripts/upload_releases.sh
TheRealWormbo: Right, I removed the link as the last thing, that's where that variable was used.

scripts/upload_releases.sh
artemisSystem: i think !p is a bash construct that expands in the string. But yeah, this stuff is hard to test. I'd probably prefer if it was a python script. Or maybe we could use gradle plugins for publishing, because i think there are some plugins that support both cf and mr?

TheRealWormbo: Yes, maybe switching to a plugin makes sense. Not sure if that needs to happen for 1.20.1 still, though.

artemisSystem: We can probably keep it as is on 1.20, and switch to a plugin for 1.21?

TheRealWormbo: > i think !p is a bash construct that expands in the string. But yeah, this stuff is hard to test. I'd probably prefer if it was a python script. Or maybe we could use gradle plugins for publishing, because i think there are some plugins that support both cf and mr?

Just looked it up: If you run the `sed` command directly at a console, you may need to disable history expansion. That's supposedly disabled in shell script execution by default: https://www.gnu.org/software/bash/manual/html_node/History-Interaction.html

